PR #13, opened 2017-10-16
initial commit of propsed spec

joshfix:


cholmes:
Thanks for this Josh. We may want to change the name in case anyone else wants to propose something different. But will get this in so we can jointly edit.

PR #37, opened 2017-11-27
Static catalog merge and major cleanup

cholmes:
Many changes in here - apologies for going so long on one branch.

The main goal was to incorporate the work @jeffnaus and @lossyrob and to try to merge them into a coherent spec that we can put out in the world.

Ended up making a few executive decisions. I'm inclined to just merge things and get to a clear repo and then we can go back and forth.

* renamed the date-time fields to just 'start' and 'end' on a suggestion from Kasey and the Planet team. Originally we had 'start_date', which is confusing if it's actually a time too. And Jeff had 'startDateTime', which seemed a bit too long. Start and end are defined as date-times, and will always obviously have a date and time in them, so it seemed like a good way to go.

* added 'assets' as its own set of links. This was from discussion in person with @lossyrob and @notthatbreezy, as well as support from @kasey. This lets assets clearly be the things you want to download, and also gives us more liberty to add in additional fields. I came close to making it so asset objects are just a link to the download and a link to a product definition. But it seemed aggressive since we don't have an actual product definition schema yet, and we should leave room for more experimentation. So there is little actually defined in assets.

* Played with a sort of prefixing / namespacing, to help designate where fields came from. But rolled it back from any sort of requirement to just something used in a few examples. 

Other work done:

* Took Jeff's spatiotemporal_feature.json and renamed to stac-item.json. I think that at some point
it may make sense to define a spatiotemporal feature in the world, but I think that would just be the start and end times on top of the geometry, and then a stac-item would extend it. Stac items contain more than what I think a spatio-temporal feature would.

* Updated samples, and added examples for Planet, DG (based on Jeff's work), NAIP (from Rob/Seth) and Landsat8. The json-spec ones are pretty solid. The static catalog ones need a bit more work.

* Added readme's throughout - to help explain more.

* Added a 'static catalog recommendations', to discuss emerging best practices that we're not yet ready to include in the main spec.

* Added an example to the swagger api spec that is in line with the stac-item defined elsewhere.

...And probably more that I can't think of. There's still more to be done, but I think with this we can get it on to trunk and call it 0.3 and start telling people to try it out. There's a few more cleanup things I want to do (catalog.json for a schema needs some thought, want to add readme for the API, and update the overall readme for the entire spec, etc), and I'd also like to add a roadmap that will spell out what we are planning to do next.


cholmes:
Ok, merged in. I promise more granular PR's in the future. Thanks to everyone who reviewed.

PR #47, opened 2018-02-12
Removal of HTTP POST for querying.

jskulavi:
…s that are already defined in GeoJSON spec.  General YAML cleanup.  Justifies version bump.

### DIFF
--- a/api-spec/spec.yaml
+++ b/api-spec/spec.yaml
@@ -2,7 +2,7 @@ swagger: '2.0'
 info:
   title: An API for SpatioTemporal Asset Catalogs
   version: 0.0.4
-  description: 
+  description:
     The SpatioTemporal Asset Catalog specification aims to standardize the way
     geospatial assets are exposed online and queried.  A 'spatiotemporal asset'
     is any file that represents information about the earth captured in a
@@ -47,7 +47,7 @@ paths:
         - application/json
       responses:
         '200':
-          description: 
+          description:
             The formal documentation of this API according to the OpenAPI
             specification, version 3.0. I.e., this document.
           schema:
@@ -59,7 +59,7 @@ paths:
   /items:
     get:
       summary: Retrieve STAC compliant Items matching filters
-      description: 
+      description:
         Returns a pageable GeoJSON feature collection of the Items matching the
         query
       operationId: getItems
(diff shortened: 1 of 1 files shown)


joshfix:
I'm a little concerned about just removing POST without proposing some other sort of mechanism for allowing searches with a request body.  We are in a situation where users _will_ be submitting searches with very long filter strings.  I'm also not convinced that HTTP verb usages are as absolutely defined as a lot of the discussion in gitter seems to imply.  There are lots of examples of dominant web companies (Amazon, Google, etc) using various HTTP methods in a large variety of ways, including using POST for searches, PUT for new items, POST for updating items, etc.

For our use cases, it is extremely convenient to provide our users the ability to submit searches via POST.  PUT works fine for both adding and updating (also used by Elasticsearch), and PATCH for partial updates.

The strongest argument for removing POST (or modifying the purpose of POST for adding new items) is to fall in line with WFS 3.0, however WFS 3.0 is not even finalized/released yet.  I believe this could (should) be a good conversation to have between the STAC and WFS authors/devs before we remove it.

jskulavi:
This PR is the manifestation of [this conversation](https://gitter.im/SpatioTemporal-Asset-Catalog/Lobby?at=5a81aa78f283b8e546500ee6).  There is no limit to the length of a GET request's query string according to the W3C specification.  The long query string limit that you are concerned about is a constraint imposed by a browser.  

Vendor-specific convenience functionality should be implemented by the vendors themselves as "extensions" while conforming to governed standards set forth by the W3C.  This PR is to realign the STAC specification with said standards.

cholmes:
@jskulavi  - thanks for the contribution! 

I don't think the conversation on this on github has been fully resolved. But the clean-up of unused imports and tweaking definitions is great, would you  mind putting those in to their own PR? And then we can continue to discuss the POST here as well, keeping this PR open. I feel it also overlaps with the generalization question and getting a full strategy. 

And Josh makes a good point about aligning with WFS. Which I think does not do the POST.

cholmes:
Going to close this, as in person discussions re-oriented POST to just be used on the /search/stac endpoint, and for POST to be the default, and GET optional. cc @gsf-trutherford who will be making a proposal for the new Query proposal.

PR #48, opened 2018-02-16
Type cleanup, removal of unused types

jskulavi:


hgs-trutherford:
There is additional cleanup work being done across a couple pull requests and issues to take care of some of these issues. I think #65 and #71 are good places to continue discussion. This is targeting an old version of the spec, so I'm closing in favor of the other ongoing work.

PR #49, opened 2018-02-19
Bbox remodel

jskulavi:
Parsing the query string for array `[xmin,ymin,xmax,ymax]` or WKT-like `bbox(xmin,ymin,xmax,ymax)` notation proved to be cumbersome and unreliable to implement.

Migrating the bbox parameter model to a canonical use of query string key/value parameters is proposed with this pull request.  

The proposed structure is modeled after OSM query implementation:  `https://www.openstreetmap.org/query?`

### DIFF
--- a/api-spec/spec.yaml
+++ b/api-spec/spec.yaml
@@ -66,7 +66,10 @@ paths:
       tags:
         - Items
       parameters:
-        - $ref: '#/parameters/bbox'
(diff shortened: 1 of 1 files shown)


hgs-trutherford:
#74 is attempting to align the filters with the WFS approach. I'm closing as I think that is a better place to discuss this issue.

PR #53, opened 2018-03-06
Fix link name

rouault:


cholmes:
Thanks @rouault! Just merged (in to a file that changed a lot, but the fix was still relevant)

PR #55, opened 2018-03-08
moved thumbnail from links to assets; updated asset format

joshfix:


cholmes:
Could you break these in to two pull requests? Partially because I think both need to a modify a lot more files. Definitely the main json definition at https://github.com/radiantearth/stac-spec/blob/dev/json-spec/json-spec.md Ideally we'd update the  json-schema folder and all the examples in one PR. Though I think just two main 'sample' examples would be ok. Plus update the narrative texts. 

I'm happy to merge in a pull request that's a subset of everything to be done, but we should at least open issues on what needs to be updated.

joshfix:
Chris, I thought when you originally recommended two PRs, I thought you were referring to one PR for the change to make thumbnails mandatory for assets, and one PR to make the second "something else" required asset.  I'm not exactly sure how to split this into two PRs... would you like one PR that keeps assets as an array but simply removes thumbnail as a mandatory link?  And a second that changes assets to a dictionary and includes updates to all the samples?

cholmes:
Hey @joshfix - I was thinking we'd try to break things up in to lots of granular commits. But I'm thinking of just making master/ a 'stable' branch and just merge lots of stuff in to dev. So these PR's should be fine to just do in one big one. Thanks for all the work!

cholmes:
Merged this in, as dev branch is now an open for any changes that get us closer to the 0.4 milestone. Tackles parts of #61, #62, #59 and #65

Issue #56, opened 2018-03-08
`self` links should always be absolute URLs


alkamin:
This change had unanimous agreement. This would allow all other relative links to be clearly resolved. While it may make moving catalogs slightly more involved, we feel it is a reasonable burden that results in a more easily navigable catalog.

cholmes:
We should probably clarify what a 'root' link means in STAC API, and what it should point at.

cholmes:
So this is all up to date for `self` links. I think still need to double check for `root`, but I think I'm just going to change this issue to self only, and then we need to do a bit more work in general on defining how root works, especially in context of stac api stuff.

PR #57, opened 2018-03-08
updated stac-item json schema for hackathon 2

jeffnaus:
Changed assets to a object (dictionary)
added thumbnails as an optional item in assets, set the type to #asset
moved requirements from the root properties into the referenced core properties

### DIFF
--- a/json-spec/json-schema/stac-item.json
+++ b/json-spec/json-schema/stac-item.json
@@ -8,23 +8,6 @@
   "allOf": [
     {
       "$ref": "#/definitions/core"
-    },
-    {
-      "properties": {
-        "properties": {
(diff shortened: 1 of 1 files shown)


cholmes:
Awesome, looks great. Makes good progress on #62, #61, #59 

I think there will likely be a few more needed updates for milestone 4, in particular the new time stuff in #64. It'd be good to also do #56 - I think it's just change the self 'string' to a 'url'.

The other major todo will be updating all the examples to validate against this. 

I'll merge this in to dev, since that's just an open workspace for 0.4.0, so feel free to make a new PR for any more JSON schema updates.

Thanks @jeffnaus !

PR #69, opened 2018-03-16
Fixing broken links

fredliporace:
Seems that leaving the 'http://' prefix out assumes that the link is relative to the repo


cholmes:
thanks!

PR #71, opened 2018-03-16
Features/post search

hgs-trutherford:
These are the base filters for STAC. I intend to investigate extension fragments next.


api-spec/spec.yaml
```diff
@@ -127,6 +119,29 @@ paths:
           description: An error occured.
           schema:
             $ref: '#/definitions/exception'
+  /search:
+    post:
+      summary: >-
+        Retrieve items matching filters. The request is a string that matches
+        the GET query exactly. POST is primarily to be used when the request
+        length exceeds GET maximums (which is not uncommon with geospatial
+        queries)
```
cholmes: I think I took a note that the group discussed flipping this around so POST should be prioritized over GET? If so we should flip this language around.

gsf-trutherford: You are right. I missed that comment in my updates.

jskulavi:
/[VERB] is an antipattern.  You all need to correct direction, else this effort is destined to fail

jskulavi:
👎

cholmes:
That is a good point that /verb isn't the most restful. I've done a quick survey of how others do search, and it does seem like word 'search' is often in there. In Planet's data api we have 'searches' which is more of a noun - https://developers.planet.com/docs/api/reference/#operation/CreateSearch and you POST your 'search' there. But then also offer a 'quick search' end point.

Github's API seems to use a verb - https://developer.github.com/v3/search/ Facebook seems to use search as a verb to https://developers.facebook.com/docs/places/search 

Definitely am open to feedback on 'the correct direction' - not sure if you mean on the /verb anti-pattern or something larger, but it'd be great if you could articulate that, perhaps on the google group: https://groups.google.com/forum/#!forum/stac-spec

cholmes:
Generally this looks good to me. @joshfix - would be great if you could review and merge.

joshfix:
I don't have an issue with /search, because "search" is in fact both a verb and a noun.  Outside of that, the bbox definition looks a little suspicious to me.  It's typed as a string, but I believe it should be an array of numbers.  Something like this:

```
  bbox:
    description: 'West, South, East, North'
    type: array
    items:
      minItems: 4
      maxItems: 4
      type: number
      minimum: -180
      maximum: 180
    example:
      - -180
      - -90
      - 180
      - 90
```

hgs-trutherford:
@cholmes 

I'm actually considering closing this pull request all together and starting fresh using #76 and #74 as the base. I think this pull request is getting out dated as we speak. What do you think?

cholmes:
Yeah, that sounds good @hgs-trutherford

PR #73, opened 2018-03-21
changing start/end to datetime, as per #64

cholmes:
Updates for the main spec changes to go to `datetime` from `start` and `end`. Fuller discussion at https://github.com/radiantearth/community-sprints/blob/master/03072018-ft-collins-co/notes/group-discussion.md#name-of-time-fields

@joshfix / @gsf-trutherford - would be great to have your review. Also I just updated wfs-stac.yaml - is that going to be the go forward? Will we delete spec.yaml? Or rename wfs-stac.yaml to that when it's done? Happy to update there too.

Would be great to have other's eyes on this too. I'm not doing the examples, as I want to update all the examples at once, so just doing these updates makes the changes clear.

### DIFF
--- a/api-spec/wfs-stac.yaml
+++ b/api-spec/wfs-stac.yaml
@@ -490,8 +490,7 @@ definitions:
             - - -122.308150179
               - 37.488035566
       properties:
-        start: '2016-05-03T13:21:30.040Z'
-        end: '2016-05-03T13:21:30.410Z'
+        datetime: '2016-05-03T13:21:30.040Z'
         provider: 'http://www.cool-sat.com'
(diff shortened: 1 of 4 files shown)


gsf-trutherford:
I think wfs-stac.yaml is not as obvious for a new user, but it does make it clear that STAC is associated with wfs. I think I'm leaning toward renaming this one back to spec.yaml and leaving the wfs connection to be defined in our readme.

cholmes:
+1 to renaming back.

matthewhanson:
I started to push a PR with a rename, but someone beat me to it. It's renamed on dev.

PR #74, opened 2018-03-21
Features/filter extensions

hgs-trutherford:
This is a document describing the core filters and a recommended approach to add extension filters.

### DIFF
--- a/api-spec/filters.md
+++ b/api-spec/filters.md
@@ -2,7 +2,7 @@ STAC Core is intending to provide a minimal set of features so that there is a l
 implementations. The core concepts in STAC are Spatial and Temporal data, so those are the only filters we want
 to support in Core.
 
-Filters should be a simple as possible for a client to construct. Since we are returning json, we want to also
+Filters should be as simple as possible for a client to construct. Since we are returning json, we want to also
 consume json. This allows clients to support filtering without additional tools. So the core filters will be 
 JSON based. The search enpoint will accept application/json format queries, and GET on the collection will 
 support URL encoded JSON. POST search the is recommended way to filter results to avoid the URL encoding 
@@ -39,7 +39,53 @@ __Filter Extensions__
 
 Spatial and Temporal filtering is not enough for many use cases. STAC is designed to be extensible, so we encourage additional filters to be provided as extensions.
 
-Filter extensions are intended to provide a place to specify additional filter types and behavior. Some filters may need to target vendor specific fields or use a more complex structure for defining the filter.
+Filter extensions are intended to provide a place to specify additional filter types and behavior. Some filters may need to target vendor specific fields or use a more complex structure when defining the filter. For example, a geospatial intersection filter that supports a polygon as input.
 
+Filter extensions should be specified as an openapi fragment. This is a section of YAML that can be incorporated into the larger api definition. These fragments can then be combined into an implementation specific api definition that describes the full api.
+
+Example fragment:
+```
+  IntersectsFilter:
+    type: object
+    description: >-
+          Only returns items that intersect with the provided polygon.
+    properties:
+      intersects:
+        $ref: '#/definitions/Polygon'
+  Polygon:
+    type: object
+    properties:
+      type:
+        type: string
+        enum:
+          - polygon
+      coordinates:
+        $ref: '#/definitions/Polygon2D'
+    required:
+      - type
+      - coordinates
+  Polygon2D:
+    type: array
+    minItems: 1
+    items:
+      $ref: '#/definitions/LinearRing2D'
+  LinearRing2D:
+    type: array
+    minItems: 4
+    items:
+      $ref: '#/definitions/Point2D'
+    example: [[-104,35],[-103,35],[-103,34],[-104,34],[-104,35]]
+  Point2D:
+    description: A 2d geojson point
+    type: array
+    minimum: 2
+    maximum: 2
+    items:
+      type: number
+      format: double
+    example:
+      - -104
+      - 35
+```
 
 
--- a/api-spec/filters.md
+++ b/api-spec/filters.md
@@ -29,11 +29,11 @@ example
 The temporal query will be based on ISO 8601 and should support time ranges as well as equality. To support range
 queries, we are using a simple JSON based language. Ranges will be specified as an object with keys indicating the 
 
-Equality is specified as `{"time": "2018-03-20T16:11:44.353Z"}`
-Before is `{"time":{"lt":"2018-03-20T16:11:44.353Z"}}`
-After is `{"time":{"gt":"2018-03-20T16:11:44.353Z"}}`
-Before with Equality is `{"time":{"lte":"2018-03-20T16:11:44.353Z"}}` 
-After with Equality is `{"time":{"gte":"2018-03-20T16:11:44.353Z"}}`
+Equality is specified as `{"time": "2018-03-20T16:11:44.353Z"}`  
+Before is `{"time":{"lt":"2018-03-20T16:11:44.353Z"}}`  
+After is `{"time":{"gt":"2018-03-20T16:11:44.353Z"}}`  
+Before with Equality is `{"time":{"lte":"2018-03-20T16:11:44.353Z"}}`  
+After with Equality is `{"time":{"gte":"2018-03-20T16:11:44.353Z"}}`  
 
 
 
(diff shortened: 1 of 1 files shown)


api-spec/filters.md
```diff
@@ -0,0 +1,133 @@
+STAC Core is intending to provide a minimal set of features so that there is a low bar to entry for catalog
+implementations. The core concepts in STAC are Spatial and Temporal data, so those are the only filters we want
+to support in Core.
+
+Filters should be as simple as possible for a client to construct. Since we are returning json, we want to also
+consume json. This allows clients to support filtering without additional tools. So the core filters will be 
+JSON based. The search enpoint will accept application/json format queries, and GET on the collection will 
+support URL encoded JSON. POST search the is recommended way to filter results to avoid the URL encoding 
+issues that can happen with GET.
```
cholmes: Maybe say that POST is required, and we recommend GET as well. 

We also might call out how a POST makes sense in REST - that it's against a 'search' endpoint, not the items themselves. Perhaps not at the top, but somewhere (could also make the recommendation on POST lower).

api-spec/filters.md
```diff
@@ -0,0 +1,133 @@
...
+it is an optional field in GeoJSON.
+
+example
+```
+{
+  "bbox": [-180,-90,180,90]
+}
+```
+
+The temporal query will be based on ISO 8601 and should support time ranges as well as equality. To support range
```
cholmes: Do we want to switch this to RFC 3339 for consistency? I forget the subtleties of the difference between the two, like if RFC 3339 is sufficient? (should also link to whichevery one we use)

api-spec/filters.md
```diff
@@ -0,0 +1,133 @@
...
+queries, we are using a simple JSON based language. Ranges will be specified as an object with keys indicating the comparison to use.
+
+Equality is specified as `{"time": "2018-03-20T16:11:44.353Z"}`  
+Before is `{"time":{"lt":"2018-03-20T16:11:44.353Z"}}`  
+After is `{"time":{"gt":"2018-03-20T16:11:44.353Z"}}`  
+Before with Equality is `{"time":{"lte":"2018-03-20T16:11:44.353Z"}}`  
+After with Equality is `{"time":{"gte":"2018-03-20T16:11:44.353Z"}}`
```
cholmes: Would be good to add a between example here - like time is before june 1st and after may 15.

api-spec/filters.md
```diff
@@ -0,0 +1,133 @@
...
+Equality is specified as `{"time": "2018-03-20T16:11:44.353Z"}`  
+Before is `{"time":{"lt":"2018-03-20T16:11:44.353Z"}}`  
+After is `{"time":{"gt":"2018-03-20T16:11:44.353Z"}}`  
+Before with Equality is `{"time":{"lte":"2018-03-20T16:11:44.353Z"}}`  
+After with Equality is `{"time":{"gte":"2018-03-20T16:11:44.353Z"}}`  
+
+Filter Extensions
+-----------------
+
+Spatial and Temporal filtering is not enough for many use cases. STAC is designed to be extensible, so we encourage additional filters to be provided as extensions.
```
cholmes: I worry a bit about being so flexible that a core implementation is not that useful. Or that an extension has to be very explicitly defined. 

Like could we just say that simple fields (numbers + text) should offer comparisons? I'm just thinking of like the use case of searching for satellite imagery - date and time and cloud cover. Can we make it a bit easier on content profiles if we don't make them explicitly define the querying for cloud cover? Like I was sorta imagining that a content profile like 'earth observation' wouldn't have to provide a modified openapi document, unless they really wanted some crazy new filter. 

I'm more than happy for those who do mongo / elasticsearch to push back due to implementation pain. I can see arguments that it's better to explicitly define the search terms, instead of doing a blanket statement. But I at least want to hear that's the reason, and we're not just going for extensibility for the sake of it.

I'd like us to present an opinionated core, that actually makes decisions about what should be done. And I think I'm ok to start just saying they're 'required' WFS can be the super flexible thing, and we can point people at that if STAC API doesn't meet their purposes. And I believe we also can require a bit more from the STAC API, because we can fall back on static stac's for people who want to keep things really simple.

api-spec/filters.md
```diff
@@ -0,0 +1,133 @@
...
+    example:
+      - -104
+      - 35
+```
+
+It is likely that there are schemas that should be used in common for types of filters that target different fields. We should define a common set of filter types that can be used in defining filters for different fields.
+- NumberRange
+- TimeRange
+- Text
+- ArrayIncludes
```
cholmes: Will this cover presence of assets? I think that'd be the other good one to think about pulling up explicitly, perhaps even put in core. Enable search by presence of 'cloud optimized geotiffs' or 'json metadata'. Or since we made it optional, by thumbnail.

api-spec/filters.md
```diff
@@ -0,0 +1,133 @@
...
+    type: array
+    minimum: 2
+    maximum: 2
+    items:
+      type: number
+      format: double
+    example:
+      - -104
+      - 35
+```
```
cholmes: I think it'd be good to publish this as a full 'stac official extension' or something - that we say this is the way to do things. We can debate on if it should be fully core or not, but I think we should have a set of filters that we say are the core extensions, that are recommended for most API's (but we tell clients they need to use the openapi doc to determine what they can query).

api-spec/filters.md
```diff
@@ -0,0 +1,133 @@
...
+Time intervals:  
+`{"time": "P1Y"}`  Assume Duration/now if no time specified?  
+`{"time": "2018/P1M"}` Any time in january of 2018  
+
+
+Adding filters to search
+------------------------
+
+I am hoping that we can agree to support openapi3 syntax going forward. It would allow a simple addition to the searchBody portion using allOf:
+```
```
cholmes: I think this would be ok, but I do think it is important to provide some form of swagger / openapi approximation. What would that look like? Like what's the 'alternative' with swagger? I'm ok if it's leaving off a decent chunk of functionality, as long as we just explain that.

api-spec/filters.md
hgs-trutherford: I took a stab at adding that to the document. I'm thinking I'll make a few small changes here, then merge so others can contribute going forward. It's hard for others to make changes to my feature branch!

api-spec/filters.md
hgs-trutherford: Updated, RFC 3339 is more explicit and is probably the better one to use.

api-spec/filters.md
hgs-trutherford: I added an example

cholmes:
Feel free to merge whenever you're ready @hgs-trutherford - changes look good and it'd be good to have this in.

api-spec/filters.md
hgs-trutherford: I think querying on every property by default is a little too open ended. It doesn't put a clear set of boundaries to be "core" compliant with a strongly defined set of search parameters. And I also worry about the scalability issues of allowing users to mix and match any number of search criteria. Some implementations might support that easily, but do we make it a requirement that all implementations should support indexes on all properties? I know our first implementation is going to be mongo based and it requires indexes to be defined ahead of time.

I do want to be opinionated on the schema used for specific data types though. I think number fields should always use the NumberRange filter schema. I'm not yet sure how to document that other than making it a strongly worded recommendation.

I could see a "search all primitives" extension that says something like "use the propery name as the key, and the value should match the type: <table of type to filter schema>"

api-spec/filters.md
hgs-trutherford: I hesitate a little on querying on the assets array. Are the types of assets strongly defined?

I was hoping that there would be enough information in the properties section so the assets don't need to be queryable. But if needed, I think a text query of ` {"assets.type": "cloud-geotiff"}` might be able to do it. It would only return records that can match that criteria. I'm not sure about all databases, but I think using the dot syntax on an array assumes an exists query on elements in the array.

api-spec/filters.md
hgs-trutherford: I agree, I think there has been enough discussion around geo intersects with a polygon that we should at least make it a first tier extension that is "official". Or even possibly part of the core.

api-spec/filters.md
hgs-trutherford: I didn't provide a swagger 2.0 version, but it would be something like:

definitions:
  searchBody:
    description: The search criteria
    type: object
    properties:
      bbox:
        $ref: '#/definitions/bbox'
  bbox:
    type: array
    items:
      type: number

PR #75, opened 2018-03-23
Sample updates

cholmes:
Various updates to the samples to align with the new ways of doing things.

In doing this work it became clear that we also still need to put something forward for products / asset definitions. I don't think we're ready for anything definitive in the main spec yet, since implementations are still right. But having a point of view around a recommended way would be good. Hopefully @matthewhanson will propose something for EO profile, as most of the examples are EO. We can specify more strongly there, and then figure out what to pull out to core.

Perhaps we could at least define some asset types and non-required fields. 

I think one lens to think through this stuff would be what we need to drive decent html pages. Certainly some of the naming / links would be nice, and as well as a 'type' (to know that something is a COG and can be a slippy map) and perhaps a 'group' for how to display (metadata vs basic products, etc). 

I also removed the DG and NAIP static catalogs, as they seemed like a lot to maintain. Once we have real production catalogs we can put samples in. cc @fredliporace - would be great to have that CBERS example in the static catalog examples when it's all ready.


matthewhanson:
Overall this looks good, but I'm not sure what the "product" json for each band is. That's a little different than what we talked about for the Asset Definition which would describe multiple assets.  I think the entire record really is a "product", rather than each of the assets being a product.

What would this product JSON contain?

@cholmes @jeffnaus

cholmes:
Oh, that is just a relic of what was there before. Should be replaced with
what we talked about. In the previous version I just put a few different
potential ways to do it.

Feel free to update the samples.

matthewhanson:
@cholmes ok, I'm fine with merging this then, and will work on updating them.

fredliporace:
@cholmes I'm working on a script to generate the stac files directly from CBERS metadata files, the script repo is https://github.com/fredliporace/cbers-2-stac. I'll submit the files as a PR into this branch, please let me know if there is a better way to review the files.

One quick question about properties/provider, should this be the agency/company that operates the satellite or the satellite identification? I was inclined to use 'INPE' (agency) for CBERS but the Landsat example uses a satellite id ('LANDSAT_8' instead of 'USGS'), while the Planet example uses the company name.

fredliporace:
Hmm,  switched to the 'dev' branch but then my PR also include  your changes since I started to work on the 'sample_update' branch. I switched back to merge into 'sample_update' branch.

cholmes:
Shoot, forgot to actually hit 'merge' on this one @fredliporace Should be in now, and you should be able to switch to dev.

PR #76, opened 2018-03-23
WFS3 core yaml doc

hgs-msmith:
I've spent some time this week working toward an updated OpenAPI spec for the union of WFS3 core, STAC core, and STAC extensions. Of course, the starting point has to be a usable WFS3 core yaml doc. Such a doc doesn't yet exist, and there are at least three sources of information:
- **Sample OpenAPI doc** (https://github.com/opengeospatial/WFS_FES/blob/master/openapi.yaml)
- **OpenAPI fragments** (https://github.com/opengeospatial/WFS_FES/tree/master/core/openapi)
- **Latest draft of the core standard** (https://rawgit.com/opengeospatial/WFS_FES/master/docs/17-069.html)

Unfortunately, these are not synchronized, resulting in contradictions. The wfs3-core.yaml doc is my attempt to reconcile these sources. In general, I have given deference to the standard doc, as it seems most up-to-date, and probably should be considered definitive.

This yaml doc surely has plenty of errors - even though the Swagger editor doesn't flag any problems, until you try to generate and run some code from it, you won't really know if it is correct. Nevertheless, if this looks like a decent starting point, I propose to use this as a basis for:
- Josh's extensions (both STAC and "transactions")
- Tim's filter proposal
- An OpenAPI 2 backport to use with the tools.

If it looks correct enough, I would also suggest that we give it to the WFS team. It really belongs on their repo.

### DIFF
--- a/api-spec/WFS3-core.yaml
+++ b/api-spec/WFS3-core.yaml
@@ -107,15 +107,17 @@ paths:
             text/html:
               schema:
                 type: string
-  '/collections/{name}':
+  '/collections/{collectionId}':
     get:
-      summary: describe the {name} feature collection
+      summary: 'describe the {name} feature collection'
       operationId: describeCollection
       tags:
         - Capabilities
+      parameters:
+      - $ref: '#/components/parameters/collectionId'
       responses:
         '200':
-          description: Metadata about the {name} collection shared by this API.
+          description: 'Metadata about the {name} collection shared by this API.'
           content:
             application/json:
               schema:
@@ -132,9 +134,9 @@ paths:
             text/html:
               schema:
                 type: string
-  '/collections/{name}/items':
+  '/collections/{collectionId}/items':
     get:
-      summary: retrieve {name} features
+      summary: 'retrieve {collectionId} features'
       description: >-
         Every feature in a dataset belongs to a collection. A dataset may
         consist of multiple feature collections. A feature collection is often a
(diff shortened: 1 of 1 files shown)


cholmes:
Hey @hgs-msmith - sorry for the delay on reviewing. But this looks really good. And I definitely agree it belongs in the WFS3 repo. It's definitely a cleaner example than the 'buildings' one.

The one thing is I'm seeing a few little errors when I load it in to http://editor.swagger.io: 

swagger_editor

I just did 'load url' on https://raw.githubusercontent.com/radiantearth/stac-spec/fdba87b998dd37b1d012933fec61bf0b7539bbc5/api-spec/WFS3-core.yaml

If you resolve that I'm happy to merge here, and once we get it in to WFS3 we can reference it.

hgs-msmith:
Corrected semantic errors:

- Cleaned up $ref for collectionId and featureId
- Added /components/schema/collectionId definition
- Added /components/schema/featureId definition
- Added /components/schema/link definition

hgs-msmith:
Chris, it looks like I'm not authorized to merge.

api-spec/WFS3-core.yaml
```diff
@@ -0,0 +1,478 @@
...
+            type: number
+            minimum: -180
+            maximum: 180
+          example:
+            - -180
+            - -90
+            - 180
+            - 90
+    featureCollectionGeoJSON:
+      type: object
```
hgs-trutherford: I think a feature collection also needs a type property set to FeatureCollection. I think that is usually done with an enum in openapi?

cholmes:
Hrm, it looks like you're a collaborator, not sure if there's something else I'm supposed to do, maybe have to explicitly turn on self merge or something.

hgs-msmith:
Chris, what's the most effective way to promote this doc to the WFS team?

cholmes:
I'd say just make a pull request on https://gitter.im/opengeospatial/WFS_FES

That should spur discussion.

PR #77, opened 2018-03-26
CBERS MUX example

fredliporace:
Not validated yet, will be updated as soon as new fields are included.
Includes a CBERS metadata example and its STAC item representation.


static-catalog/examples/CBERS/CBERS_4_MUX_20170528_090_084_L2.json
```diff
@@ -0,0 +1,44 @@
...
+            23.463987
+          ]
+        ]
+      ]
+    ]
+  }, 
+  "properties": {
+    "datetime": "2017-05-28T09:01:17Z", 
+    "provider": "INPE", 
+    "license": "CC-BY-SA-3.0"
```
cholmes: Would be great to add some additional fields here from the metadata. The recommendation is to prefix them, see https://github.com/radiantearth/stac-spec/blob/master/static-catalog/static-recommendations.md#prefixes

@matthewhanson is working on the 'EO profile', which should give some common fields / definitions for things like sun angle, view angle, etc. But in the meantime you could just use a provider prefix, like inpe:satellite_name, inpe:orientatationAngle, etc.

I'd say add the fields that could make sense for users to search. The rest can live in extended metadata.

static-catalog/examples/CBERS/CBERS_4_MUX_20170528_090_084_L2.json
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
```
cholmes: Looks like you don't have the 'assets' yet, but putting the full metadata file in the asset list would be a good idea. Then people can get the ephemeris and other more obscure metadata.

fredliporace:
OK, will add the README.md as soon as I have a first version with links and assets sections.

Since my main reference are the examples I prefer to work for now with the 'sample_update' branch, but I may change the PR to 'dev' if that's better.

static-catalog/examples/CBERS/CBERS_4_MUX_20170528_090_084_L2.json
fredliporace: Just included the 'eo' section based on the current Landsat-8 and Planet examples. 

Is it possible to assume that all angles (sun_elevation, sun_azimuth, off_nadir_angle) are in degrees and distances (resolution) are in meters?

For eo:off_nadir_angle, is there a convention about the angle sign or are we interested here only in the absolute value?

fredliporace:
Removed the WIP prefix to allow merge with partial example

cholmes:
Just merged in, as we can all just work on dev.

PR #83, opened 2018-04-03
Api spec

hgs-msmith:
Defining STAC as as an extension of WFS3 core, now defined at: https://github.com/opengeospatial/WFS_FES/blob/master/openapi.yaml
Includes two files:
* fragment that includes only the STAC elements to be added to WFS3 core
* complete API, union of WFS3 core and STAC

### DIFF
--- a/api-spec/WFS3core+STAC.yaml
+++ b/api-spec/WFS3core+STAC.yaml
@@ -1,0 +1,631 @@
+openapi: 3.0.1
+info:
+  title: STAC API conforming to the OGC Web Feature Service standard
+  version: 0.0.1
+  description: >-
+    This is a sample OpenAPI definition that conforms to the OGC Web Feature
+    Service specification (conformance classes: "Core", "GeoJSON", "HTML" and
+    "OpenAPI 3.0").
+  contact:
+    name: Acme Corporation
+    email: info@example.org
+    url: 'http://example.org/'
+  license:
+    name: CC-BY 4.0 license
+    url: 'https://creativecommons.org/licenses/by/4.0/'
+servers:
+  - url: 'https://dev.example.org/'
+    description: Development server
+  - url: 'https://data.example.org/'
+    description: Production server
+paths:
+  /:
+    get:
+      summary: landing page of this API
+      description: >-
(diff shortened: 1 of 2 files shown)


api-spec/WFS3core+STAC.yaml
```diff
@@ -0,0 +1,631 @@
+openapi: 3.0.1
+info:
+  title: STAC API conforming to the OGC Web Feature Service standard
+  version: 0.0.1
+  description: >-
+    This is a sample OpenAPI definition that conforms to the OGC Web Feature
+    Service specification (conformance classes: "Core", "GeoJSON", "HTML" and
+    "OpenAPI 3.0").
```
cholmes: Would be good to add in the fact that this also implements the STAC specification. Indeed the description should be more like 'This is a sample OpenAPI definition of a SpatioTemporal Asset Catalog API that also conforms to the OGC ...'

api-spec/WFS3core+STAC.yaml
```diff
@@ -0,0 +1,631 @@
...
+      tags:
+        - STAC
+      requestBody:
+        content:
+          'application/json':      
+            schema:
+              $ref: '#/components/schemas/searchBody'
+      responses:
+        '200':
+          description: A feature.
```
cholmes: Should add the STAC Item response here, the geojson as an example. The ones in the spec.yaml and wfs-stac yaml are up to date with the latest.

api-spec/WFS3core+STAC.yaml
```diff
@@ -0,0 +1,631 @@
...
+                type: string
+# START STAC-specific operations                
+  /search/stac:
+    post:
+      summary: >-
+        retrieve items matching filters.
+      operationId: searchSTAC
+      tags:
+        - STAC
+      requestBody:
```
cholmes: Can we specify the filters here in some way? Even if it's just an example that includes the two main filters.

hgs-msmith:
Added STAC-specific comments
Added additional itemCollection types
Added /search/stac example

hgs-msmith:
I will work up a STAC-only example next.

PR #84, opened 2018-04-03
Added example of STAC API standalone

hgs-msmith:


api-spec/STAC-standalone.yaml
```diff
@@ -0,0 +1,395 @@
+openapi: 3.0.1
+info:
+  title: STAC API without WFS3
+  version: 0.0.1
```
cholmes: Should we change this (and the others) to 0.4.0?

api-spec/STAC-standalone.yaml
```diff
@@ -0,0 +1,395 @@
+openapi: 3.0.1
+info:
+  title: STAC API without WFS3
```
cholmes: I'd prefer more just 'standalone stac api', so a dev totally new to this doesn't have to figure out what wfs3 is to get their head around it.

hgs-msmith:
Chris, you can go ahead with those changes.

PR #89, opened 2018-04-04
Stac api updates [WIP]

cholmes:
Pushing ahead with some narrative updates. I'll likely continue to add to this branch - just want to be sure there are some eyes on it. Feel free to commit whenever makes sense @hgs-msmith and then I can just start a new PR.


cholmes:
Yeah, a caveat would be good. I'll write one quick and add to this PR

Issue #86, opened 2018-04-03
Specify GET search request for STAC API


cholmes:
I think we want to provide guidance on what the GET endpoint for the STAC API looks like. I think POST in the search endpoint should be the required one. But I think it'd be good to have a recommendation on a GET endpoint, that is the same filter values, etc. 

We could do this as an extension, to emphasize that it's optional. Though I'd be fine if we had it in core and just said it was optional.

cholmes:
Marking this 0.4.0, but I'm more than happy to let it slip if we've got too much to do by tomorrow. Just don't want to lose track of it.

PR #90, opened 2018-04-04
Added GET version of /search/stac

hgs-msmith:
This a proposed simple alternative to the POST method.


hgs-msmith:
BTW, if we like this, I'll add it to WFS3core+STAC.yaml also.

api-spec/STAC-standalone.yaml
```diff
@@ -19,8 +19,39 @@ servers:
     description: Production server
 paths:
   /search/stac:
+    get:
+      summary: >-
+        retrieve items matching filters. Intended as a shorthand API for simple queries.
```
cholmes: I'd throw a little '(optional)' bit in here so implementors know it's not required.

hgs-msmith: Looks good - I'll add this to WFS3core+STAC.yaml also.

hgs-msmith:
Fixes #86 
Fixes #90

PR #91, opened 2018-04-04
add Collection extension spec

matthewhanson:
a draft spec for a Collection extension. A Collection is simply a way to combine common metadata into a single record that applies to a group of `Item` records.

### DIFF
--- a/json-spec/extensions/stac-collection-spec.md
+++ b/json-spec/extensions/stac-collection-spec.md
@@ -1,0 +1,154 @@
+# STAC Collection Extension Spec
+
+A group of STAC `Item` objects from a single source usually shared common metadata. This is especially true with satellite imagery that uses the STAC EO Extension. Rather than including these common metadata fields on every `Item`, they can be grouped into a `Collection`.
+
+## Collection Extension Description
+
+| element             | type info                 | name                    | description                                                                                 | 
+|----------------------|---------------------------|-------------------------|---------------------------------------------------------------------------------------------| 
+| collection_name | string | Collection Name | A name given to the Collection
+| collection_description | string (optional) | Collection Description | A human readable description of the collection
+
+A `Collection` does not have many specific fields, as it may contain any fields that are in the core spec as well as any other extension. If a field is specified in `Collection`, it will be ignored in any `Item` that links to that `Collection`. This is important because a `Collection` is the metadta that is common across all `Item` objects. If a field is variable at all, it should be part of the `Item` record.
+
+#### Linking to a `Collection`
+An `Item` links to a collection by providing a "link" with "rel" set to "collection". For example
+```
+{ "rel": "collection", "href": "link/to/collection/record" }
+```
+
+#### Using a `Collection`
+The fields from the `Collection` record can be merged with an `Item` record to get the complete metadata for the `Item`.
+```
+# a collection
+{
+    "properties": {
+        "collection_name": "my_collection",
+        "provider": "me",
+        "license": "MIT"
+    }
+}
+# an item (incomplete)
+{
+    "id": "SCENE_001",
+    "properties": {
+        "datetime": "2017-01-01T00:00:00Z",
+        "geometry": {...}
+    },
+    "links": {
+        { "rel": "collection", "href": "link/to/my_collection" }
+        ...
+    },
+    "assets": {...}
+}
+```
+
+The merged `Item` then looks like this:
+```
+{
+    "id": "SCENE_001",
+    "properties": {
+        "collection_name": "my_collection",
+        "provider": "me",
+        "license": "MIT"    
+        "datetime": "2017-01-01T00:00:00Z",
+        "geometry": {...}
+    },
+    "links": {
+        { "rel": "collection", "href": "link/to/my_collection" }
+        ...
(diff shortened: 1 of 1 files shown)


json-spec/extensions/stac-collection-spec.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Collection Extension Description
+
+| element             | type info                 | name                    | description                                                                                 | 
+|----------------------|---------------------------|-------------------------|---------------------------------------------------------------------------------------------| 
+| collection_name | string | Collection Name | A name given to the Collection
+| collection_description | string (optional) | Collection Description | A human readable description of the collection
+
+A `Collection` does not have many specific fields, as it may contain any fields that are in the core spec as well as any other extension. If a field is specified in `Collection`, it will be ignored in any `Item` that links to that `Collection`. This is important because a `Collection` is the metadta that is common across all `Item` objects. If a field is variable at all, it should be part of the `Item` record.
```
cholmes: What's the advantage of the collection field overriding the Item?

I general it seems to me better that if someone only looked at a single piece of data and didn't follow the collection link that they could count on a field being correct, as opposed to having to check the collection to see if it got overridden. 

 I was also thinking it could be a valuable feature to have Item's be able to override collections. Like if there's 'field_z' that 98% of items have a value of 'true' and 2% have 'false' then an implementer could just stick field_z='true' at the collection level.

But I'm probably just missing something as to why that route would be bad.

json-spec/extensions/stac-collection-spec.md
```diff
@@ -0,0 +1,154 @@
...
+
+#### Linking to a `Collection`
```
cholmes: I'm sorta wondering if we should call it 'collection metadata' or 'collection info' or something. Especially once you move from earth observation world to WFS world - the 'collection' is the set of data, it's not the common metadata. And the logical endpoint in WFS is something like /collections/<collection_name>/(?) Like in that path something like /collections/buildings/info.json would make more sense to me than like /collections/buildings/collection.json

I'm also still contemplating how this could cover the 'root catalog' stuff as well - https://github.com/radiantearth/stac-spec/tree/dev/static-catalog#catalog Like the fields in it - contact, formats, keywords, homepage, etc could make sense in the same file? Maybe it's that the static catalog link would be required to implement both the collection schema and the 'catalog' schema, and the catalog schema is just that it has links?

json-spec/extensions/stac-collection-spec.md
```diff
@@ -0,0 +1,154 @@
...
+An `Item` links to a collection by providing a "link" with "rel" set to "collection". For example
+```
+{ "rel": "collection", "href": "link/to/collection/record" }
+```
+
+#### Using a `Collection`
+The fields from the `Collection` record can be merged with an `Item` record to get the complete metadata for the `Item`.
```
cholmes: Should make it clear if it's just properties that get merged, or if every field in the collection gets merged. Like it's not clear if the 'bands' below would be in the 'merged' full item. I think you intend it to be, but should make it clear in the doc.

json-spec/extensions/stac-collection-spec.md
matthewhanson: I was originally thinking that it could be default values, but that could be overridden.  
The problem I see with that is the main reason Collection level metadata is useful to the developer is because you *know* they are all the same.  If my bands are described in the collection level metadata than I can use that for all my items. If that data isn't common across all items then I'm back to having to check both, and then somehow reconcile them if they are different.  IMHO it ends up defeating the whole purpose for having collection level metadata to begin with. It's most important with the band information, I really need to know that the all of my items contain the same band metadata, but I could see it as needed for other fields as well.

OTOH I do see overriding as valuable too, but it's really important to know whether or not the field really is in 100% of the items or it is not.  I could see this as a feature of collections, being able to specify if a field is overridable or not.

json-spec/extensions/stac-collection-spec.md
matthewhanson: I went with Collection for two reasons. 

- It seemed compatible with WFS, although I haven't thought in detail about how it would work in practice, so your points are well taken.
- For NASA data, this is what a "Collection" is - a group of similar data, some might refer to it as a product.

I haven't thought about the root catalog stuff at all.

json-spec/extensions/stac-collection-spec.md
matthewhanson: Ah yes good point. Yes the idea was that everything gets merged. 

I think it's a good approach overall, because it means we don't have to have confusing specs for what goes in a Collection record, it's up to the provider to decide + it's easy to make a complete Item record from both of them.

json-spec/extensions/stac-collection-spec.md
cholmes: Cool. I'll see if I can come up with something tomorrow (thursday).

json-spec/extensions/stac-collection-spec.md
cholmes: Yeah, I like it. Just need to make it clear. If you can add it in to the PR great, if not I'll merge in tomorrow and add it in as I understand it.

PR #92, opened 2018-04-04
Features/api review

hgs-trutherford:
Updating to use the object syntax for time range queries instead of the period syntax. The period syntax is not well supported by libraries, so I don't think it should be part of the core. I think it should be an extension to add period syntax support.

I also changed the organization of timeRange to be in line with the bbox schema being defined as a parameter and removed the timeRange schema.

Also some minor formatting fixes for the bbox description.

### DIFF
--- a/api-spec/STAC-fragment.yaml
+++ b/api-spec/STAC-fragment.yaml
@@ -62,6 +62,81 @@ paths:
               schema:
                 type: string
 components:
+  parameters:
+    limit:
+      name: limit
+      in: query
+      description: |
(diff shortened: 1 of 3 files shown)


hgs-msmith:
Agreed - let's go with the time range approach and get some practical experience using it.

hgs-msmith:
I should have thought of this earlier. I don't think the object syntax for query parameters is supported for OpenAPI 2.0. So, this will prevent us from backporting this to a Swagger example.

PR #93, opened 2018-04-05
Stac api updates for latest json spec

cholmes:
@hgs-msmith and @hgs-trutherford - I wanted you guys to take a look at these. I updated for the latest json spec. Was hoping to just make narrative changes today, but things also need to be in line with the specs.

I'd like to do more to make things consistent, like update all the examples to be consistent how an api might actually be designed. And maybe its a separate doc, but I'd like an example openapi thing that is not templated, that shows 'real' STAC collections in the WFS.

But not sure if I'll get to that today, so submitting this for review. If you don't manage to review by the end of the day that's cool, I'll merge in, just wanted to be sure this was seen.

### DIFF
--- a/api-spec/STAC-fragment.yaml
+++ b/api-spec/STAC-fragment.yaml
@@ -304,16 +304,19 @@ components:
               - - -122.308150179
                 - 37.488035566
         properties:
-          start: '2016-05-03T13:21:30.040Z'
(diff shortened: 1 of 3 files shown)


cholmes:
Cool. 

The page token might be useful, but it could also just be a recommendation. I don't think it's really getting used, and indeed if it is then everything should still work fine.

PR #102, opened 2018-04-17
Bug fixes for swagger2 yaml examples. Rev to 0.4.1

hgs-msmith:
Added extensions for creating and editing STAC items
Fixed bugs discovered in existing yaml docs:
1. POST /search/stac : changed return type to itemCollection from featureCollectionGeoJSON
2. definitions/item: 
  - changed bbox to #/definitions/bbox
  - updated assets property to correctly model the intended associative array
3. definitions/itemProperties: updated start/end to datetime


hgs-msmith:
Chris, I agree with all of your comments. Let's get it step up the way we want it - we can bail on these two PRs and resubmit the same material organized as you propose. What do you think?

cholmes:
sure, sounds great.

hgs-msmith:
Removed the extensions examples for this PR - will submit a separate PR for this.

PR #103, opened 2018-04-18
Bug fixes for OpenAPI3 yaml. Rev to 0.4.1

hgs-msmith:
Added extensions for creating and editing STAC items
Fixed bugs discovered in existing yaml docs:
* updated assets property to correctly model the intended associative array
* cleaned up searchBody example


hgs-msmith:
Yes, that bothered me also, but I haven't figured out how to reference the core definition of bbox and time in both the query parameter and the composition of the searchBody object. Any ideas?

hgs-msmith:
Removed the extensions examples for this PR - will submit a separate PR for this.

PR #104, opened 2018-04-20
Proposed transaction extensions

hgs-msmith:
This is a formalization of the transaction extensions originally proposed by Josh Fix.
I will add a standalone OpenAPI fragment to this PR soon.


extensions/transaction/WFS3core+STAC+extensions.yaml
```diff
@@ -0,0 +1,1008 @@
...
+              schema:
+                $ref: '#/components/schemas/exception'
+            text/html:
+              schema:
+                type: string
+    patch:
+      summary: update an existing feature by Id with a partial item definition
+      description: >-
+        Use this method to update an existing feature. Requires a GeoJSON 
+        fragement (containing the fields to be updated) be submitted.
```
hgs-trutherford: Should we call out https://tools.ietf.org/html/rfc7396 here? I think that RFC closely matches what we were discussing at the last STAC meeting.

extensions/transaction/WFS3core+STAC+extensions.yaml
```diff
@@ -0,0 +1,1008 @@
...
+        assets:
+          analytic:
+            name: 4-Band Analytic
+            href: >-
+              http://cool-sat.com/catalog/collections/cs/items/CS3-20160503_132130_04/analytic.tif
+          thumbnail:
+            name: thumbnail
+            href: >-
+              http://cool-sat.com/catalog/collections/cs/items/CS3-20160503_132130_04/thumb.png
+    partialItem:
```
hgs-trutherford: How does partialItem differ from Item? Is it just removing the required portion? I think that would be covered by calling out to the rfc I suggested above and we can just document Item once?

hgs-msmith:
No, I did not. I don't know why that got included in the PR. Only the 3 files in extensions/transaction should be included in this PR. I think I can just delete that extra file from this branch.

hgs-msmith:
Yes, I just put the MD file up as a placeholder. Do you have time to fill out some content?

hgs-msmith:
OK, WFS3-core.yaml is removed - I think Git saw it as removed from DEV, but not MASTER, therefore it looked like a new file.

extensions/transaction/transaction-fragment.yaml
```diff
@@ -0,0 +1,206 @@
+paths:
+  '/collections/{collectionId}/items':
+    post:
+      summary: add a new feature to a collection
```
hgs-trutherford: Should this say "feature(s)" to indicate the feature collection?

extensions/transaction/transaction-fragment.yaml
```diff
@@ -0,0 +1,206 @@
...
+      operationId: postFeature
+      tags:
+        - Extensions
+      parameters:
+        - $ref: '#/components/parameters/collectionId'
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/itemCollection'
```
hgs-trutherford: Perhaps this could be anyOf:
-item
-itemCollection 

to support single features and collections?

Issue #107, opened 2018-04-23
Planet example doesn't use property namespaces


hgs-trutherford:
The planet example is currently inconsistent with the other examples. It shows all the properties without the trait set namespaces. I think we need to agree on the names with or without namespacing. It will make searching and parsing difficult if the property keys are not consistent.

https://github.com/radiantearth/stac-spec/blob/master/json-spec/examples/planet-sample.json

Issue #105, opened 2018-04-23
Band information in Landsat8 example is not as specified


m-mohr:
I think the Landsat example at https://github.com/radiantearth/stac-spec/blob/master/json-spec/examples/landsat8-sample.json has some errors, which I struggled with during my work on #81.

- Why is "eo:bands" part of the assets? I assume it's an error and should be moved either one level up or into properties.
- The landsat example uses `fwhm` instead of `full_width_half_max`. ;)

cholmes:
Apologies for that. Someone else pointed out something similar a couple days ago. I'll try to cut a 0.4.1 release with this asap.

cholmes:
Ok, I think this is fixed. Moved bands out, I just did an example link to collections, and updated the readme to point at the example in extensions - https://github.com/radiantearth/stac-spec/tree/master/extensions/examples So that also moves out the fwhm stuff.

Thanks @m-mohr for the feedback! We're excited to collaborate on the 'collection' level stuff. Sorry we didn't get a ton fleshed out for this release, but I'm sure it will continue to evolve, and we're also happy to take queues from you.

m-mohr:
@cholmes Thanks!

Additionally, I think you should have a look at https://github.com/radiantearth/stac-spec/blob/master/extensions/examples/landsat8-merged.json . There is also the eo:bands in the wrong place. Sorry, I forget to mention that in the initial issue text. The files are now basically the same, maybe the merged example can be removed at all?

Take your time with the collections, I am happy to contribute as much as I can. Have some initial things posted in #81.

cholmes:
Well the idea of the 'merged' was that it showed the combination of the landsat8-sample.json and the L1T-collection.json in one. Demonstrating that what 'complete' Item record could look like, pulling in the 'collection' fields in to the smaller Item file. So it's like the full representation of a record, if a client expanded things out.

Though I do now realize that it doesn't use full_width_half_max so that should be updated.

m-mohr:
@cholmes Okay, thanks for the clarification, I understand that now. Still, there some to be some inconsistencies (or I don't understand it correctly):

- eo:bands is also still part of assets in the merged example.
- Why is eo:bands not part of properties in the collection?
- Why does landsat8-sample still duplicate provider, license, eo:gsd and other fields that already exist in the collection?
- Why is collection_name and collection_description missing in the merged example?

[...]

PR #109, opened 2018-04-24
Example fixes

cholmes:
Updates on examples:

 #105 - updated the landsat example to the one that is in the extensions folder, as I'd always meant to make that the core example

#107 - fixes to try to do this. Would like to get CRS in, but will need more time on that.


json-spec/examples/planet-sample.json
```diff
@@ -55,30 +55,31 @@
...
+    "pl:anomalous_pixels": 0.14,
+    "pl:columns": 9111,
+    "pl:epsg_code": 32724,
+    "pl:ground_control": true,
+    "pl:instrument": "PS2",
+    "pl:item_type": "PSScene4Band",
+    "pl:origin_x": 508263,
+    "pl:origin_y": 8557524,
+    "pl:pixel_resolution": 3,
+    "pl:provider": "planetscope",
```
hgs-trutherford: Is pl:provider supposed to be different from provider? Wouldn't planetscope be more of a product or sensor name than a provider?

Also, I wouldn't hold anything up for it. But I think we are going to see tons of different date strings per provider. We might want to figure out the published, updated, acquired, start, end, etc. convention at some point.

json-spec/examples/landsat8-sample.json
```diff
@@ -63,6 +61,7 @@
     { "rel":"self", "href": "http://landsat-pds.s3.amazonaws.com/L8/153/025/LC81530252014153LGN00/LC81530252014153LGN00.json"},
     { "rel":"alternate", "href": "https://landsatonaws.com/L8/153/025/LC81530252014153LGN00", "type": "html"},
     { "rel":"catalog", "href":  "http://landsat-pds.s3.amazonaws.com/L8/catalog.json"}
+    { "rel":"collection", "href": "http://landsat-pds.s3.amazonaws.com/L8/L1T-collection.json"}
```
hgs-trutherford: Should this link to the actual file in our repo? Or since this is an example we want to make it look real?

json-spec/examples/planet-sample.json
cholmes: I paused on that one, but I kinda wanted to show that one could just take one's internal fields and map what makes sense and just include the rest. Definitely a redesigned Planet api with STAC and eo profiles in mind would change that. And indeed it'd probably use the eo sensor field. 

I had the same thought about published, updated, acquired as well. Published and updated feel like they should almost be at a WFS level, as any feature might have those fields, and it'd be good if clients could expect a common name. And then some optional time fields in the EO profile seem like they could be useful.

I do think we should pretty soon take a survey of the fields people add in their provider section to the EO profile and just get to common definitions - more optional fields in EO profile. Like rows and columns, origin_x, origin_y, epsg code, etc.

json-spec/examples/landsat8-sample.json
cholmes: I went back and forth on that myself. I ended up on 'make it look real', and then added a note in the readme that the extensions folder has a full example...

cholmes:
Going to merge this, to not let great be the enemy of good, so I can push out a 0.4.1. Will be easy to do a 0.4.2 if we want to continue to make tweaks.

PR #113, opened 2018-05-29
updated EO spec to match new collection spec

matthewhanson:
related to this PR: https://github.com/radiantearth/stac-spec/pull/112 the current EO spec referred to the existing collection spec.

This PR updates the EO spec to match the collection spec in the PR above.

In addition this PR includes one other important change. eo:crs has been removed (the CRS in WKT format) and replaced with eo:epsg to refer to the EPSG code of the data.


cholmes:
Would it be possible to break out the eo:crs -> eo:epsg in to its own PR? If it's a pain to do that's ok - can just merge here. But I'd like to have the commits be a record of changes, and this one feels important enough to warrant its own PR.

extensions/stac-eo-spec.md
```diff
@@ -132,6 +81,14 @@ The band's common_name is the name that is commonly used to refer to that band's
 |lwir11    | 10.5 - 11.5 |       |           |     10     |          |      31   
 |lwir12    | 11.5 - 12.5 |        |          |     11     |          |      32
 
-### Sample data
 
-See the [examples/](examples/) folder for a full landsat definition.
\ No newline at end of file
+## Using Collections
```
cholmes: I like that this section exists a lot, but I think it'd be good to move it further up in the document, and maybe even group the field additions in to collection vs item level. These would be 'recommendations' - it is good to give the flexibility for data that may want to change some of the values. But I think we should promote the collection concept more in the EO spec, while also getting across that you don't _have_ to use it.

extensions/stac-eo-spec.md
```diff
@@ -1,40 +1,35 @@
...
-GSD values in the bands those should be specified in the `eo:bands` section of the Collection, specified below.
+**eo:bands** This is a dictionary of band information where each key in the dictionary is an identifier for the band (e.g., "B01", "B02", "B1", "B5", "QA"). See below for more information on band metadata.
 
-**eo:crs**: The Coordinate Reference System (CRS) is the native reference system (sometimes called a 'projection') used by the data, provided in [Well-Known Text (WKT) format](https://en.wikipedia.org/wiki/Well-known_text). This field is required. If the data does not have a CRS, such as in the case of non-rectified imagery with Ground Control Points, eo:crs should be set to null.
+**eo:epsg**: The [EPSG code](http://epsg.io) is a code that refers to the spatial reference system (sometimes called a 'projection') used by the data. If the data does not have a CRS, such as in the case of non-rectified imagery with Ground Control Points, eo:crs should be set to null.
```
cholmes: I mentioned this in a comment on the PR, but it'd be great if we could break this out to its own PR, as it's a fairly substantial change, and I want to be sure it shows up in release notes, etc.

PR #114, opened 2018-06-04
Update links

matthewhanson:
Update links to be a dictionary, like assets.

The JSON schemas need to be updated, but they are out of date on a few other things so it would be good if someone could open up a PR for an updated schema.

Francesco Bartoli suggested that each links item should still contain the 'rel' attribute. I'm just concerned that will end up being the same as the key in the dictionary in which case it would be redundant. Thoughts?

### DIFF
--- a/extensions/examples/L1T-collection.json
+++ b/extensions/examples/L1T-collection.json
@@ -3,8 +3,8 @@
 
 
     "properties": {
-        "collection_name": "L1T",
-        "collection_description": "Landat 8 imagery that is radiometrically calibrated and orthorectified using gound points and Digital Elevation Model (DEM) data to correct relief displacement.",
+        "collection": "L1T",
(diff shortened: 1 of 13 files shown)


matthewhanson:
Also note that links is updated for the collections and EO extensions in the PRs #112 and #113

cholmes:
Updated to merge to dev instead of master, so we can get it in with a set of changes, clean up all the schemas and release.

cholmes:
I think we should keep the 'rel' field. Though they will often be the same as the 'rel' link I think they may not always be. Rel makes it very clear that if an html link is created from the json then it should be set as the 'rel' attribute. If it's just the key then people may lose that thread more easily.

I could also see cases where the rel is different than the link key. Like we might have a number of 'rel'='alternate', where the link key / 'type' is item-rss, item-html, etc. In the link you'd need 'rel' and 'type', but in the key you'd want to uniquely identify the 'link type'. At some point we should make some recommended 'link types' with their 'rel' values.

PR #116, opened 2018-06-21
Update collection ext

matthewhanson:
Updated collection extension to use a prefix in the field names it adds. Also adds a field so now there is:

cx:id
cx:name
cx:description

cc @cholmes @mojodna


cholmes:
Oh, the other thing that would be really nice to have, though I'm not exactly sure how to make it, is a json schema validator. Like ideally the main validator would have a way to 'add on' an extension's json, so you could have it check your cx: or eo: fields too. But definitely not required to get this PR in.

mojodna:
I realized somewhat belatedly that there's no reason it needs to be a 2-letter prefix. `c:` would work just as well and be slightly less baroque.

extensions/examples/landsat8-sample.json
```diff
@@ -63,7 +63,7 @@
     "self": { "rel":"self", "href": "http://landsat-pds.s3.amazonaws.com/L8/153/025/LC81530252014153LGN00/LC81530252014153LGN00.json"},
     "alternate": { "rel":"alternate", "href": "https://landsatonaws.com/L8/153/025/LC81530252014153LGN00", "type": "html"},
     "catalog": { "rel":"catalog", "href":  "http://landsat-pds.s3.amazonaws.com/L8/catalog.json"},
-    "collection": { "rel":"collection", "href": "http://landsat-pds.s3.amazonaws.com/L8/L1T-collection.json"}
+    "cx:collection": { "rel":"cx:ollection", "href": "http://landsat-pds.s3.amazonaws.com/L8/L1T-collection.json"}
```
mojodna: Typo: `cx:ollection`.

cholmes: Fixed in https://github.com/radiantearth/stac-spec/pull/122 (among other things)

PR #119, opened 2018-06-22
Add validation for object-style links and assets

mojodna:


cholmes:
Looks great, thanks @mojodna! I seemed to have rights to your branch, so I just removed the thumbnail checking as it was breaking the validator included in the stac repo, plus we removed it as a required field.

cholmes:
D'oh - did this against master. Trying to do everything against dev first. Usually I catch these first, but missed it.

Issue #54, opened 2018-03-06
Adding (content-)type to links


rouault:
For consistency with WFS 3 links, it could be interesting to add a "type" property to links to describe the content-type.

So the liinks from https://github.com/radiantearth/stac-spec/blob/dev/static-catalog/examples/Planet/20170129_181030_0e1f_analytic.json would become:

```
"links": [
    { "rel": "self", "type": "application/json", "href": "http://cool-sat.com/catalog/CS3-20160503_132130_04/CS3-20160503_132130_04.json"},
    { "rel": "thumbnail", "type": "image/png", "href":"thumbnail.png"},
    { "rel": "catalog", "type": "application/json",  "href": "http://cool-sat.com/catalog/"},
    { "rel": "acquisition",  "type": "application/json", "href": "http://cool-sat.com/catalog/acquisitions/20160503_56"}
]
```

Advantages: possibiliity to add alternate representations for the same resource
Drawbacks: may make client life harder to determine "unique" resources. In particular for list of items.

e.g
```
    { "rel": "catalog", "type": "application/json",  "href": "http://itemA.json"},
    { "rel": "catalog", "type": "text/html",  "href": "http://itemA.html"},
    { "rel": "catalog", "type": "application/json",  "href": "http://itemB.json"},
    { "rel": "catalog", "type": "text/html",  "href": "http://itemB.html"},
```

So perhaps not such a good idea to have multiple representations. Or perhaps provide them as rel:alternate in the itemXXXX.json file ?

cholmes:
Where's the section on WFS links? I'm definitely interested in consistency with them. And picked this style of rel links so that type could be possible. Does WFS require that all links have a type?

I'm not sure if there will be many cases with the 'rel' + type being the same. Hrm, except for if we do a 'source' and like a mosaic could have multiple 'source' files it refers to. Though what's the use case for the client to figure out the unique resources?

joshfix:
FYI we do have a type property for links, it's just not required.  See the swagger spec:

https://github.com/radiantearth/stac-spec/blob/dev/api-spec/spec.yaml#L200-L215

cholmes:
Setting this for 0.6.0 - as I think there's growing need for this. I'm looking more at 'assets', but I think it'd be good for links too. 

I think we also need to figure out types and 'sub-types', like tiff / geotiff, and then 'cloud optimized geotiff'. 

[...]

PR #123, opened 2018-07-01
Updating CBERS example description

fredliporace:
Some minor broken links also fixed.
CBERS static STAC catalog will be reset to use 0.5.0 for the next scene ingestions, starting 07/03.
Maybe not in time for 0.5.0, but this could be part of a 0.5.1 minor.

### DIFF
--- a/json-spec/examples/CBERS_4_MUX_20170528_090_084_L2.json
+++ b/json-spec/examples/CBERS_4_MUX_20170528_090_084_L2.json
@@ -1,11 +1,11 @@
 {
-  "id": "CBERS_4_MUX_20180622_090_084_L2",
+  "id": "CBERS_4_MUX_20170528_090_084_L2",
   "type": "Feature",
   "bbox": [
-    23.501111,
-    13.70101,
-    24.848708,
-    14.987979
+    23.465111,
+    13.700498,
+    24.812825,
+    14.98818
   ],
   "geometry": {
     "type": "MultiPolygon",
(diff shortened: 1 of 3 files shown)


json-spec/examples/CBERS_4_MUX_20170528_090_084_L2.json
```diff
@@ -51,55 +51,59 @@
...
-      "href": "https://cbers-stac.s3.amazonaws.com/CBERS4/MUX/090/084/CBERS_4_MUX_20180622_090_084_L2.json"
+      "href": "https://cbers-stac.s3.amazonaws.com/CBERS4/MUX/090/084/CBERS_4_MUX_20170528_090_084_L2.json"
     },
     "catalog": {
       "rel": "catalog",
       "href": "https://cbers-stac.s3.amazonaws.com/CBERS4/MUX/090/catalog.json"
     },
-    "cx:collection": {
+    "collection": {
       "rel": "collection",
```
fredliporace: @cholmes I'm not sure here, should that be c:collection?

json-spec/examples/CBERS_4_MUX_20170528_090_084_L2.json
```diff
@@ -51,55 +51,59 @@
...
+      "href": "s3://cbers-pds/CBERS4/MUX/090/084/CBERS_4_MUX_20170528_090_084_L2/CBERS_4_MUX_20170528_090_084_L2_BAND6.xml",
       "required": true,
       "type": "xml"
     },
     "B5": {
-      "href": "s3://cbers-pds/CBERS4/MUX/090/084/CBERS_4_MUX_20180622_090_084_L2/CBERS_4_MUX_20180622_090_084_L2_BAND5.tif",
+      "href": "s3://cbers-pds/CBERS4/MUX/090/084/CBERS_4_MUX_20170528_090_084_L2/CBERS_4_MUX_20170528_090_084_L2_BAND5.tif",
       "type": "GeoTIFF",
+      "format": "COG",
       "required": true,
```
fredliporace: This is an example of what I mentioned in the gitter forum, it is not clear to me the difference between 'type' and 'format'. Also, if we should use 'COG' for cloud optimized and 'GetTIFF' for not optimized.

cholmes: Yeah, we need to just set down some recommendations for this. Discussion in https://github.com/radiantearth/stac-spec/issues/54

PR #125, opened 2018-07-07
Minor fixes for json schema and openapi specs

francbartoli:


json-spec/json-schema/stac-item.json
```diff
@@ -49,41 +49,48 @@
...
-                      }
-                    }
-                  ]
-                }
+              "items": {
+                "$ref": "#/definitions/link"
               },
-              "patternProperties": {
-                ".+": {
-                  "$ref": "#/definitions/link"
```
cholmes: @francbartoli - did you mean to delete these? and the "pattern" stuff above? I'm pretty sure it's me messing this up - I accidentally merged it to master when I wanted to on dev, and then tried to revert, and it messed everything up. If you can revert this commit in your branch that'd be great. And I'll try to fix it on master, where it seems to still be messed up.

cholmes:
Thanks for the fixes!

francbartoli:
@cholmes thanks for merging even I'm not convinced at all that the current json schema is defined properly. I'd like to get in depth into it in my spare time

PR #132, opened 2018-07-30
Spell out FWHM to match eo:bands field descriptions

jbants:
This PR changes the extensions examples to match the field descriptions in https://github.com/radiantearth/stac-spec/blob/master/extensions/stac-eo-spec.md


cholmes:
Thanks for the PR! Much appreciated.

Issue #140, opened 2018-08-01
Inconsistencies in catalog specification


m-mohr:
I went through the catalog specification, i.e. the readme (README.md), recommendations (static-recommendations.md) and JSON Schema (json-schema/catalog.json). There are several inconsistencies and problems with it:
* Fields are not clearly defined (see also #78, #133, #137, #138)
* The recommendations mention a "Product". What is this about? It is not mentioned in the Readme or JSON Schema.
* I think a "properties" property should be mentioned for the catalogs as @matthewhanson mentioned in the chat yesterday that the EO extension (and others) could be used in the catalog, but it's not clear where to put them.
* The JSON schema needs to be updated, it is not up-to-date with recent developments of the rest of the spec. Examples: Style of links (#134 / #127), License (#133), datetime vs. startDate / endDate (partly covered by #130), ...
* There is a property "TODO" in the JSON schema of the catalog.

m-mohr:
Should be fixed with #239 .

m-mohr:
JSON schema is getting updated in #250.

PR #144, opened 2018-08-02
[WIP] Fixed and cleaned up JSON schema for STAC catalogs

m-mohr:
Cleaned up JSON schema for STAC catalogs. Should still be compliant with 0.5.2 as previous scheme was an invalid JSON schema anyway.

Notable changes, partly fixing #140:
- Removed proprietary license object
- Removed unused definitions (assets, geometry)
- Disallowed nulled links (one link is required as of spec)
- Added proposed properties for providers
- Removed the breaking "TODO" property.

Would still like to change
```
        "requesterPays": {
          "type": "string"
        }    
```
to
```
        "requesterPays": {
          "type": "boolean"
        }    
```
But it wouldn't validate any of the current catalogs so kept it as a string for now.

It would also be nice to add more descriptions and add a property "properties" as mentioned in #140.


m-mohr:
Would still like to change
```
        "requesterPays": {
          "type": "string"
        }    
```
to
```
        "requesterPays": {
          "type": "boolean"
        }    
```
But it wouldn't validate any of the current catalogs so kept it as a string for now.

m-mohr:
I saw that some catalogs use the proprietary license object, so we may want to add it back.

m-mohr:
Closing for now. JSON schema for catalogs will require a major rewrite anyway once datasets are available.

PR #147, opened 2018-08-13
Fixing #146. Changing dateTime to date-time and url to uri to fix JSON Schema

### LINKED ISSUE #146: STAC JSON Schema format tags use some invalid values
In https://github.com/radiantearth/stac-spec/blob/master/api-spec/WFS3core%2BSTAC.yaml there is a reference to "dateTime" and "url" string formats

```
type: string
format: dateTime
```

Those should be `date-time` and `uri` to be valid JSON schema.

jasongilman:
Fixes #146

### DIFF
--- a/api-spec/STAC-fragment.yaml
+++ b/api-spec/STAC-fragment.yaml
@@ -345,7 +345,7 @@ components:
       properties:
         next:
           type: string
-          format: url
+          format: uri
           description: >-
             A URL to obtain the next paginated data set. If not present, the
             client should utilize the `nextPageToken`.
--- a/api-spec/STAC-standalone-swagger2.yaml
+++ b/api-spec/STAC-standalone-swagger2.yaml
@@ -430,7 +430,7 @@ definitions:
     properties:
       next:
         type: string
-        format: url
+        format: uri
         description: >-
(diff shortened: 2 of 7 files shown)


m-mohr:
I agree with the date-time change, but url was a valid format as of the OpenApi 3 spec (anything is allowed, uri is just an example in the spec). URI is actually more permissive than URL and if URI is intended, the respective descriptions should be changed, too.

jasongilman:
I didn’t realize that about the URL. This pull request should probably be rejected anyway since it was for master instead of dev.

m-mohr:
Oh, that's true. You can close the PR yourself and create a new one... :)

cholmes:
It's an easy drop down to change the PR to be against dev instead. But it looks like it lead to conflicts, though it's just that we remove the swagger2 stuff. So @jasongilman - feel free to close yourself and make a new one, or just clean up your branch for a clean merge.

m-mohr:
I think this issue gets obsolete with #167 anyway. At least I couldn't find any references to dateTime.

m-mohr:
Let's close this, it's fixed by #167.

Issue #142, opened 2018-08-01
Band names should be properties instead of object keys


hgs-truthe01:
The band definitions use a string as the key for the bands object. The names of those keys don't seem well defined. The examples show "B1" and "B01".

In order to search for all items with a nir band, it would be helpful to be able to index the band definitions. As it is now, the band names would need to be consistent across all items to be easily searchable. For example "eo:bands.common_name=nir" is easier than "eo:bands.**B1**.common_name=nir" where **B1** is unknown ahead of time or mixed with **B01** across items.

I would suggest bands be changed to an array and band_name be added to the band definition instead of as a key

@hgs-msmith

m-mohr:
This sounds related to the discussion of https://github.com/radiantearth/stac-spec/issues/127#issuecomment-405300991. With the band names as keys it is easier to access them directly by band name and to make sure they are unique. Of course that is a problem in the searching scenario and maybe just a application-specific optimization as mentioned in the referenced comment.

hgs-trutherford:
Are band names well known? Most of the time band names seem to be hinting at band order, which I think is well defined. Arrays are a better fit for representing order as object keys are not ordered in JSON. I guess I'm unsure when I would use a string to reference "band 3" instead of band[2].

Also, if you know the band name ahead of time, don't you already know the band information? Band definitions generally don't change very often.

jeffnaus:
The EO spec includes band name in its extension writeup, here: https://github.com/radiantearth/stac-spec/blob/master/extensions/stac-eo-spec.md

The comon_name element of the band object is the value.

We probably should make a more complete list of common band names.

hgs-truthe01:
I wanted to add an additional data point here.

My team was just looking as adding a dataset with a red edge band and it ended up looking like:
```
eo:bands: {
...
   'red edge': {
      common_name: 'red edge',
      center_wavelength: 0.710,
      full_width_half_max: 0.040
    }
}
```
I usually try pretty hard to avoid spaces in object keys. We could certainly provide a rule like underscore replacement, but that is adding unneeded complexity.

Also, the common_name already provides that information.

For the HSI case, the band order is what matters and that is lost in the dictionary lookup. Are we also going to add a band index field that has the order? In my opinion, that is what an array is for.

m-mohr:
[...]

Issue #127, opened 2018-07-11
Update catalog.json to new 'links' style


cholmes:
Update catalog.json in the examples and the schema validation to do link json in the same way we do in 0.5.0. Like do
```
{
  "name": "LC08/01/107/061",
  "description": "Landsat 8 Collection 1 Path 107 Row 61",
  "links": {
    "self": {
      "href": "https://storage.cloud.google.com/gcp-public-data-landsat/LC08/01/107/061/catalog.json",
      "rel": "self"
  },
  "parent": {
    "href": "https://storage.cloud.google.com/gcp-public-data-landsat/LC08/01/107/061/catalog.json",
    "rel": "self"
  },
  "root": {
    "href": "https://storage.cloud.google.com/gcp-public-data-landsat//catalog.json",
    "rel": "self"
  },
  "collection": {
    "href": "https://storage.cloud.google.com/gcp-public-data-landsat/LC08/01/catalog.json",
    "rel": "self"
  },
  "items": [
    {
      "rel": "item",
      "href": "https://storage.cloud.google.com/gcp-public-data-landsat/LC08/01/107/061/LC08_L1GT_107061_20130816_20170503_01_T2.json"
    },
  {
    "rel": "item",
    "href": "https://storage.cloud.google.com/gcp-public-data-landsat/LC08/01/107/061/LC08_L1GT_107061_20150721_20170406_01_T2.json"
 },
...
```

@mojodna - I think this will need a change to STAC browser. Maybe we should make this a 0.6.0 release? Though it feels like a relatively minor change, just for the static browser, and one we should have made in 0.5.0.

I suppose we could also just keep it the same, and have the catalog links not be keyed. cc @matthewhanson

mojodna:
+1 for consistency across all "lists" of things.

However, @metasim's question about how to model STAC as Postgres DDL (I was interpreting it as "static STAC") got me thinking about this a bit more... In order to model it, a join table would need to be introduced that contains the "key", which complicates things a bit.

Remind me again why lists are no longer modeled as arrays; I missed that discussion and still don't fully buy the change.

cholmes:
I believe @matthewhanson advocated for switching from arrays. I think the idea was to make it easier to reference a particular asset (originally was just assets, then links for consistency). So instead of making implementations search through the whole array to find the 'geotiff' asset, or the 'metadata' asset, they can just use the geotiff key to get exactly what they want. 

[...]

PR #151, opened 2018-08-14
Return links and assets to lists

mojodna:
I'd like to drop the use of lists-as-objects in favor of using vanilla JSON arrays for Item `links` and `assets` as well as catalog-level `links`. My feeling is that the former introduces application-specific "optimizations" into the core spec (in the form of facilitating lookups) while dirtying the semantics associated with those lists and actually _adding_ complexity:

* JSON objects have no inherent ordering (in practice, they do in most implementations, but that seems like a bad assumption to make); this means that the resulting lists won't necessarily be consistent across viewers
* a non-semantic element is introduced (the object key) that introduces ambiguity; some implementations will apply meaning to this (a `cog` key in an asset collection) in ways that potentially conflict with "better" resolution rules (`ref` / `rel` / `type`)
* a collection of things is inherently a list; to illustrate this, actions like "give me the first asset" (`jq .assets[0]`) become more difficult (`jq ".assets | to_entries | first | .value"`; I had a hard time figuring this out)
* since object keys are now required, STAC generators will need to produce them; most keys will be opaque strings (UUIDs, etc) with duplication needing to become a consideration (merging `links` collections from multiple catalogs becomes a non-trivial operation (in case keys overlap), especially if it's deemed important to try to retain object keys)

(This is an incomplete PR; additional references to lists-as-objects should be found in prose, JSON schema, and examples.)

Refs #127 
Refs #142

### DIFF
--- a/json-spec/json-spec.md
+++ b/json-spec/json-spec.md
@@ -29,8 +29,8 @@ your feedback will be directly incorporated.
 | id              | string          | Provider ID                | Provider ID for the item                       													| 
 | geometry        | geojson         | Geometry                   | Bounding Box + Footprint of the item in lat/long (EPSG 4326)										|
 | datetime 		  | date and time   | Date and Time 	         | The searchable date/time of the assets, in UTC (Formatted in RFC 3339)    						| 
-| links           | dictionary           | Resource Links             | Dictionary of link objects to resources and related URLs (self required), each with a unique key								|
-| assets          | dictionary           | Assets            	   	 | Dict of asset objects that can be be download (at least one required, thumbnail strongly recommended), each with a unique key |
+| links           | list            | Resource Links             | List of link objects to resources and related URLs (self required), each with an optional, unique `id` attribute                                 |
+| assets          | list            | Assets            	   	 | List of asset objects that can be be download (at least one required, thumbnail strongly recommended), with an optional, unique `id` attribute |
 | provider        | string          | Provider     (optional)    | Provider name  																					|
 | license         | string          | Data License (optional)    | Item's license name based on SPDX License List or following guidelines for non-SPDX licenses 	|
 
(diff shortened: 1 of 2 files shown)


matthewhanson:
@mojodna I've been the main one pushing for these assets and links to be objects rather than lists, and since I won't be there for this sprint I suspect I'll lose this one.
However @mojodna makes a good case here, especially for auto-generated items. I've been working on some code that generates derived STAC items and have run into this problem of what to name the keys.

The original motivation here was for users to be able to easily specify which assets they are interested in. Being lists doesn't prevent this, just changes the logic in the client for determining it. *My main interest* is that users must absolutely be able to specify assets by a consistent key across collections, or by a band common_name in the case of the EO extension. If this is still possible switching to lists then I'm on board.

cc @cholmes @scisco

mojodna:
In the PR, I included an `id` attribute. I think that's better as `key` and can be used to achieve what @matthewhanson wants to achieve, I think. 

Also, a `role` attribute (e.g. thumbnail or primary) may serve some of the purpose of the object keys.

matthewhanson:
Since the group meeting Wed morning, I've thought about this more and have an additional argument for why assets should remain as dictionaries. First, as I said before dictionary keys really only become important when using Collections (or whatever they are to be called now) which define a set of common metadata across Items.

When you do have a Collection of Items in this manner though it is important that the assets have keys, and that they are unique so that an Item can be merged with the Collection-level metadata to get a complete record.

Example:

For example, a landsat 8 collection can define the assets:
```
"assets": {
   "B1": {
      "type": "image/geotiff-cog" 
      "eo:bands": ["B1"],
      "description": "This is an asset for Band 1 (the coastal band)"
}
```

An individual item can then define just the href to the asset for that item:
```
"assets": {
   "B1": {
      "href": "path/to/datafile" 
}
```

The Item can then be merged with the Collection record to get a complete record. This is how [sat-search](https://github.com/sat-utils/sat-search/) works. The merging may occur on the server side instead, but either way the use of dictionaries provides an unambiguous way to merge these, whereas using lists requires a lot of logic of iterating and finding matching records. Worse, if a key is duplicated the behavior is undefined. Maybe you do want two thumbnails, but different formats for them. 

For lists to work we need to ensure that an id field is unique across all items.

Related to https://github.com/radiantearth/stac-spec/issues/153

cc @mojodna @cholmes @jeffnaus

cholmes:
Wanted to summarize where we got during the sprint:

Everyone agreed that links (both in catalogs and in items) should be lists. It was done later, and done to 'align' with assets, but didn't have any real good reasoning and had some drawbacks.

In general people also felt that it was 'ok' to have links and assets not share the exact structure. Though obviously there's a slight preference to keep things aligned.

So perhaps make this PR about just the links. We can keep discussing assets here, or move it to its own issue.

cholmes:
Closing this - decision was to change links to an array (which has happened), and to leave assets as dictionary. We are not opposed to changing it at some point (and indeed those in favor of assets as dictionary mostly just want to be sure there is an id that is required + referencable), but aren't going to do it for the 0.6.0 release.

PR #159, opened 2018-08-14
use travis ci to check schema and api spec is valid

justinwp:
a little rusty with jsonschema and not sure if ajv should be able to get the remote ref


justinwp:
passing here: https://travis-ci.org/justinwp/stac-spec

cholmes:
Going to close this, as we have CircleCI on. But super psyched that you worked on this @justinwp - thanks a lot! Glad we got ahead on CI.

I do think we need to enable the json schema validation on the circleCI, should be sure we have an issue for that.

Issue #150, opened 2018-08-14
Mosaic / composite item type


mojodna:
Grouped images should be representable within a STAC Catalog. These may range from multiple parts of a DigitalGlobe strip (multiple assets, but logically and effectively treated as a single asset alongside additional (metadata, thumbnail, etc.) assets) to a curated collection of items that an entity wishes to share.

Properties of an element within such a collection of assets:

* layer index
* range of valid resolutions (i.e. zooms) for an individual asset
* asset validity footprint (potentially represented using quad keys corresponding to Web Mercator tiles at a resolution greater than that of the image)
* resampling method to use when necessary
* whether to stretch values to shift into a visible range
* min/max values for each band when stretching
* custom NODATA value
* band selections

In the case of components of a DG strip, this would be included in the list of assets (ideally with some indication that it should be used in preference to individual components, perhaps using [`q` values](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html): "Quality factors allow the user or user agent to indicate the relative degree of preference for that media-range, using the qvalue scale from 0 to 1 (section 3.9). The default value is q=1.")

A curated collection of items potentially equates to a STAC Item in terms of usage, so this is something to reconcile.

Many of these things provide hints for display purposes and aren't necessarily descriptors for the item itself, so that also merits consideration.

The combination of resolution ranges and quad keys allows a tiler to know when a given asset should be included (and in what order) in a composite image.

We envision this as a small-ish JSON file that is HTTP-accessible and can be used in place of a COG URL (or `POST`ed) with something like tiles.rdnt.io.

/cc @sharkinsspatial

matthewhanson:
@mojodna
Most of these fields look to me like they could apply to any EO data - for example NODATA and gain/offset values.  This is actually a problem right now when dealing with landsat data since each band has different gain/offsets to get to TOA' (note not true TOA but TOA without sun angle correction), and that info is only available through the MTL metadatata file, not in the STAC item.

simonff:
Yes, none of these are mosaic-specific. We will add them to the raster extension(s) of the Dataset spec.

@mojodna : What's a sample use case for quality factors?

@matthewhanson : For reference, in EE almost all properties from the MTL file are stored in each Landsat asset's metadata, and thus TOA can be computed on the fly. In Collection 1 average sun angle can be taken into account: https://landsat.usgs.gov/using-usgs-landsat-8-product

[...]

Issue #148, opened 2018-08-14
Storage profiles


mojodna:
[Static STAC Catalog entries](https://github.com/radiantearth/stac-spec/blob/master/static-catalog/README.md#catalog) include a `provider` element which isn't particularly well-specified. The current example is S3-specific:

```json
    "provider": {
        "scheme": "s3",
        "region": "us-east-1",
        "requesterPays": "true"
    }
```

I propose that storage profiles be defined for each type of object store (or just plain HTTP) so that keys have more meaningful values. The above would then be:

```json
"provider": {
  "s3:region": "us-east-1",
  "s3:requesterPays": true
}
```

However, this presumes that all Items falling beneath such a parent catalog assume the same attributes, which may not be correct. Instead, perhaps these should be `asset` properties and the `provider` block omitted:

```javascript
"assets": [
  {
    "href": "s3://.../...",
    "type": "...",
    "name": "...",
    "s3:region": "us-east-1", // can be inferred from the bucket name, but explicit is better than implicit
    "s3:requesterPays": true
  }
]
```

Proposed keys:

* `s3:region` - name of the AWS region that a bucket is in, e.g. `us-west-2` (enum)
* `s3:requesterPays` - whether the S3 Requester Pays feature is enabled for this item; implies that AWS credentials must be used to access it (boolean)
* `s3:public` - whether the item can be retrieved without S3 credentials; this allows clients to create HTTP(S) URLs and make items downloadable directly (boolean)

matthewhanson:
@mojodna a few things

- We discussed having object fields but decided to keep things simple originally and currently we don't have any fields that are objects under properties, which simplifies searching. Object fields, such as links and assets, we moved out of properties to the top level for this region. I think there are multiple fields that make sense to make object types, but we need to define how a user searches these fields.

- This potentially duplicates a lot of info across items. In sat-api collections don't include assets, and there isn't anything specifically that excludes parts of assets being defined, but we need to specify how these would end up getting merged. I will start a new ticket for this.

- Looks like we need to specify an s3 extension that adds these additional fields, and likewise create one for google storage.

m-mohr:
I like the idea of profiles for storage. We were thinking about storage for the datasets, too, but we had defined a fixed set, which I am not so happy with. What do you think @simonff? Would that be a more flexible way to go for datasets, too?

simonff:
[...]

PR #164, opened 2018-08-15
Dataset metadata spec

justinwp:
Just getting it down somewhere...


m-mohr:
@justinwp We agreed with Chris earlier this morning that we keep the dataset spec out of stac-spec repo and put it into the sprint repo for now. For example, I started here: https://github.com/radiantearth/community-sprints/tree/datasets-spec/08132018-menlo-park-ca/datasets-spec

dataset-spec/README.md
```diff
@@ -0,0 +1,92 @@
...
+| license_url   | string                                | Dataset License URL             | Dataset's license URL SHOULD be specified if `license_name`  does not contain a SPDX License Identifier. |
+| provider      | [Provider Object]                     | Data Provider                   | The organization that creates the content of the dataset.    |
+| host          | Host Object                           | Storage Provider                | The organization that hosts the dataset.                     |
+| geometry      | [GeoJSON Object](http://geojson.org/) | Spatial extent (required)       | The spatial extent covered by the dataset as [GeoJSON](http://geojson.org/) object. |
+| datetime      | string                                | Temporal extent (required)      | Temporal extent covered by the dataset. Date/time intervals MUST be formatted according to ISO 8601. Open date ranges are not supported by ISO 8601 and MUST be encoded as proposed by [Dublin Core Collection Description: Open Date Range Format](http://www.ukoln.ac.uk/metadata/dcmi/date-dccd-odrf/2005-08-13/). |
```
cholmes: I think it could be good here to not reuse datetime and geometry, as those are item specific, and instead really talk about the extent. 

Ideal to me would be to align with WFS3, so we can add the dataset fields directly to their 'collection' metadata. They have it like:

"extent": {
        "spatial": [ 7.01, 50.63, 7.22, 50.78 ],
        "temporal": [ "2010-02-15T12:34:56Z", "2018-03-18T12:11:00Z" ]
      },

See 'example 4' on https://rawgit.com/opengeospatial/WFS_FES/master/docs/17-069.html

I could be up for a different form, like flatten to spatial_extent and temporal_extent - but I don't think we should reuse the item fields, and I don't believe a dataset should need to conform to the Item schema. If we go with something different I'd want to advocate to WFS3 group to align with us. And I'd see a WFS 3 compliant STAC implementation using these dataset fields in their /collections/{collectionId}/ endpoint.

dataset-spec/README.md
```diff
@@ -0,0 +1,92 @@
...
+| process_graph | Process Graph Object                  | Processing chain                | ...                                                          |
```
cholmes: What is this? And what's the definition of it? May make sense as an extension?

dataset-spec/README.md
m-mohr: I'm fine with renaming it to spatial extent and temporal extent, we just used those names for consistency within STAC. I'll change that in the next minutes. I don't like the the way temporal extents are specified in WFS though as it doesn't allow to leave the end time out (i.e. open date ranges).

dataset-spec/README.md
m-mohr: Process graph was added for provenance, but will definitely be an extension. The same applies for dimensions. Will change that, too.

dataset-spec/README.md
```diff
@@ -0,0 +1,85 @@
...
+| url     | string | Organization homepage | Homepage of the provider.                       |
+
+###  Host Object
+
+| Element        | Type    | Name                  | Description                                                  |
+| -------------- | ------- | --------------------- | ------------------------------------------------------------ |
+| description    | string  | Description           | Detailed description to explain the hosting details. [CommonMark 0.28](http://commonmark.org/) syntax MAY be used for rich text representation. |
+| scheme         | string  | Scheme (required)     | Values: S3, GCS, URL, OTHER                                  |
+| id             | string  | Identifier (required) | Host-specific identifier such as an URL or asset id.         |
+| region         | string  | Region                | Provider specific region where the data is stored.           |
```
matthewhanson: Is region primarily an AWS thing or is it general to all cloud providers?

dataset-spec/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+##  EO extension (eo) - Items and Datasets
+
+TODO
+
+We follow the STAC EO extension, but propose additional fields:
+
+| Element      | Type     | Name               | Description           |
+| ------------ | -------- | ------------------ | --------------------- |
+| unit         | ?        |                    |                       |
```
matthewhanson: If some sort of unit field is provided it should probably be per-band. But I'm not convinced it's needed, what are the possible values here?

dataset-spec/README.md
```diff
@@ -0,0 +1,85 @@
...
+#### Bands
+
+TODO
+
+We follow the STAC EO extension for bands, but propose additional fields:
+
+| Element   | Type     | Name          | Description                                                  |
+| --------- | -------- | ------------- | ------------------------------------------------------------ |
+| nodata    | [number] | Nodata values | The no data value(s).                                        |
+| data_type | string   | Data Type     | Data type for band values including its bit size. Values: uint8, uint16, uint32, uint64, int8, int16, int32, int64, float16, float32, float64 |
```
matthewhanson: I do think nodata, gain, and offset are useful to have per band, but not sure about data_type. The other fields may not be provided in the actual datafile (although providers should be encouraged to set them), but data_type will always be available.

dataset-spec/README.md
```diff
@@ -0,0 +1,85 @@
+# Dataset Spec for STAC
+
+## Introduction
+
+One topic of interest has been the search of datasets*, instead of within a dataset, i.e. in (sub-)catalogs, items and assets. [STAC](https://github.com/radiantearth/stac-spec) is focused on search within a dataset, but it includes some simple constructs to catalog datasets. This could be an independent spec that STAC uses, and others can also independently use, to describe datasets in a lightweight way.
+
+*\* There is no standardized name for the concept we are describing here. Others called it: dataset series (ISO 19115), collection (CNES, NASA), dataset (JAXA), dataset series (ESA), product (JAXA).*
+
+## Core
+
```
matthewhanson: In the EO spec we have gsd (Ground Sample Distance) at the top level, and it's also provided per band because resolution may vary by band.  At the top level it represents the best resolution to enable searching. I think it makes sense at the Dataset level rather than the Item level.

dataset-spec/README.md
m-mohr: The schema was written before we knew about the idea of storage profiles in #148. I would really like to have profiles instead of having the storage details directly baked into the dataset spec.

region would probably be an AWS specific thing, which should be in a separate profile as proposed in #148. If others have regions, too, then they should have it separately in their profiles aswell.

dataset-spec/README.md
m-mohr: Yes, makes sense to have it per band (and per dimension). Moved it to the bands. An example would be °C for temperatures, but could be any unit of measurement. Preferably SI, but could also link to [UDUNITS](https://www.unidata.ucar.edu/software/udunits/) or [the dictionary of UoM](https://www.unc.edu/~rowlett/units/).

dataset-spec/README.md
m-mohr: data_type was a field I ported from openEO and got spec'ed by my predecessor. I agree, I think it doesn't really make sense. Removed it for now.

dataset-spec/README.md
matthewhanson: I'm not sure I entirely get what you are saying.  
I think datasets are a concept which will be used in general with STAC - although I'm still not sure if the intention is that datasets are always present and if they are themselves part of core. I personally think they should part of core, the include core fields: temporal and spatial extent are the unions of core fields, license, provider...

So the EO extension, or any extension, should define additions to both the Dataset and the Item.

m-mohr: I hope they are part of the core and always present, but not sure what others think.

I agree with what you are saying about extensions and that answers basically my question. So I would expect that the additional fields we are proposing for datasets will be incorporated into the EO extension. If you are okay with that, I'd already move them to the eo extension in the branch we are currently working on. They are currently a bit badly located in the dataset-spec.

dataset-spec/README.md
```diff
@@ -0,0 +1,86 @@
...
+#### Bands
+
+We follow the STAC EO extension for bands, but propose additional fields:
+
+| Element | Type     | Name          | Description                                                  |
+| ------- | -------- | ------------- | ------------------------------------------------------------ |
+| nodata  | [number] | Nodata values | The no data value(s).                                        |
+| offset  | number   | Offset        | Offset to convert band values to the actual measurement scale. |
+| scale   | number   | Scale         | Scale to convert band values to the actual measurement scale. |
```
matthewhanson: Been thinking about this more and scale and offset definitely need to be specified per asset, not per band. Individual items can have different scale and offsets. While Landsat collection 1 uses a consistent scale/offset for all scenes, this was not always the case and a provider may certainly provide them per Item, it is not really a band characteristic, it is an asset characteristic.
I had thought of it because there are other reasons too - for instance I might take the RGB bands and scale them to  a Byte and provide that as a visible True Color image, or other bands to make a False Color image, but still want to reference the bands that make up that 3-band image.

matthewhanson: In the same vain, units really is per asset as well. I might provide a TOA datafile as well as a Surface reflectance file.

And in those cases I might want to use different nodata values as well, so they really should all be per-asset.  Units and nodata can be specified per asset in the Dataset, but scale/offset needs to be specified for each asset in each Item.

dataset-spec/README.md
```diff
@@ -0,0 +1,86 @@
...
+| version         | string                                | Dataset Version                 | Version of the dataset. [Semantic Versioning (SemVer)](https://semver.org/) SHOULD be followed. |
+| license         | string                                | Dataset License Name (required) | Dataset's license(s) as a [SPDX License identifier or expression](https://spdx.org/licenses/) or `proprietary` if the license is not on the SPDX license list. See `license_url` for more information. |
+| license_url     | string                                | Dataset License URL             | Dataset's license URL SHOULD be specified if `license` is set to `proprietary`. |
+| provider        | [Provider Object]                     | Data Provider                   | The organizations that created the content of the dataset.   |
+| host            | Host Object                           | Storage Provider                | The organization that hosts the dataset.                     |
+| spatial_extent  | [GeoJSON Object](http://geojson.org/) | Spatial extent (required)       | The spatial extent covered by the dataset as [GeoJSON](http://geojson.org/) object. |
```
fredliporace: @m-mohr Is this to be interpreted as 'possible extent' or as 'current extent'? My concern here is that some missions have the capacity to image most of the Earth but do not make a systematic acquisitions - CBERS for instance.
The footprint for CBERS-4 MUX scenes may be obtained [here](https://viewer.remotepixel.ca/#3/40/-70.5), just select CBERS on the upper right panel. You'll see that most of northern Canada is not covered yet by the dataset, but a new scene may be acquired anytime in the future. In that case should the spatial_extent be changed? If is is changed could the dataset version be the same?

dataset-spec/README.md
fredliporace: @m-mohr @matthewhanson gain (scale) and offsets may have distinct values depending on the date, see the case for [LS4 and LS5](https://landsat.usgs.gov/april-25-2017-changes-l4-and-l5-radiometric-calibration-collection-1-processing). How should that cases be represented? I also prefer gain instead of scale, but this may be a Landsat thing.

simonff: @matthewhanson : I don't quite follow. What is an asset in this case? My understanding that by assets STAC people usually mean files on disk. Do you mean a specific file for a specific item? But a gdal file can have two bands that have different gain/offset.

For reference, datasets in EE almost always have the same gain/offset for each band across all items. The two exceptions are Landsat and SeaWiFS OceanColor dataset that has different processing for different items in the same dataset. We did not dare to change Landsat, but for SeaWiFS we applied gain/offset before ingestion to make the data easier to use in ARD-like manner. 

So in general, yes, each item can have its own gain/offset. I've never seen nodata values differ across items, though. If they do, it might make more sense to split the dataset in two or more.

dataset-spec/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+##  EO extension (eo) - Items and Datasets
+
+We follow the STAC EO extension where meaningful (see below), but propose additional fields:
+
+| Element      | Type     | Name               | Description           |
+| ------------ | -------- | ------------------ | --------------------- |
+| asset_schema | object   | Asset Schema       | TODO                  |
+| nodata       | [number] | Nodata values      | The no data value(s). |
+| pyramid      | object   | Pyramid parameters | TODO                  |
```
simonff: So far I only used this as en enum describing the way averaging is down during pyramiding: MEAN for normal continuous bands like EO observations or temperature/DEM/etc, MODE for categorical values (eg, landcover classes), SAMPLE for the rest (typically this means bitmask bands). SAMPLE means just take one possibly random value, as nothing else makes sense. (In reality in this case we always take UL value for each 2x2 grid).

dataset-spec/README.md
matthewhanson: Ok, that makes sense though - each cloud provider should define it's own storage profile (extension?)

dataset-spec/README.md
matthewhanson: Yes, I think those additional fields belong in the EO extension...although I think that I'd still like to see some of the non-varying asset info in Datasets, but I'll make the case and provide some examples in the EO extension.

dataset-spec/README.md
m-mohr: I'd go that route, yes. Is that dataset specific or could that be also catalog or item specific?

dataset-spec/README.md
m-mohr: @simonff, could you come up with a proposal for this field? And maybe also asset_schema? These initially were your ideas and I think you have a much better idea to define them.

dataset-spec/README.md
```diff
@@ -0,0 +1,86 @@
...
+| keywords        | [string]                              | Keywords                        | List of keywords describing the dataset.                     |
+| version         | string                                | Dataset Version                 | Version of the dataset. [Semantic Versioning (SemVer)](https://semver.org/) SHOULD be followed. |
+| license         | string                                | Dataset License Name (required) | Dataset's license(s) as a [SPDX License identifier or expression](https://spdx.org/licenses/) or `proprietary` if the license is not on the SPDX license list. See `license_url` for more information. |
+| license_url     | string                                | Dataset License URL             | Dataset's license URL SHOULD be specified if `license` is set to `proprietary`. |
+| provider        | [Provider Object]                     | Data Provider                   | The organizations that created the content of the dataset.   |
```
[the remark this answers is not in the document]
m-mohr:
```
{
  "id":"sentinel2-processed",
  "description":"Sentinel 2 NDVI max composite processed by Final Processor and Another Processing Comp, data originates from ESA.",
  "spatial_extent":{

  },
  "temporal_extent":"2015/2018",
  "license":"",
  "provider":{
    "name":"Final processor, Inc.",
    "url":"http://www.final-corporation.com"
  },
  "pc:process_chain":{
    "process":"max_time_composite"
  },
  "history":[
    {
      "provider":{
        "organization":"Another Processing Comp, Inc.",
        "url":"http://processing.inc"
      },
      "dataset_url":"http://processing.inc/datasets/sentinel2-processed/catalog.json",
      "pc:process_chain":{
        "process":"ndvi"
      }
    },
    {
      "provider":{
        "organization":"ESA",
        "url":"http://esa.eu"
      },
      "dataset_url":"http://esa.eu/data/sentinel-2"
    }
  ]
}
```

dataset-spec/README.md
m-mohr: For openEO we had it defined as current extent and so I had that in mind, but I am open to both. Whoever has the best arguments wins. ;-) 

@fredliporace For whatever reason, I can't comment on the comment regarding gain/scale below, so I'll add the comment here: 
- Honestly, it's just a name. Don't care whether its gain or scale. It has the same meaning, correct?
- Not sure whether we can cover all of that, currently it can't represented. Maybe just leave it out if it depends on another property? I don't have a proper solution for it yet - but open for proposals.

dataset-spec/README.md
m-mohr: Which of these non-varying info would that be?

dataset-spec/README.md
m-mohr: @matthewhanson In the normal "STAC way" with Items and Assets that might be feasible, but for openEO (and GEE?), we still need those information - whenever meaningful - per dataset or band, as we don't have any items/assets where this information is available! 

@fredliporace 
- Honestly, it's just a name, but I am more used to scale. Don't really care whether its gain or scale. It has the same meaning, correct?
- Not sure whether we can cover all of that, currently it can't represented. Maybe just leave it out if it depends on another property? I don't have a proper solution for it yet - but open for proposals.

m-mohr:
Added a first draft for the dimension extension, is open for review.
Was partly proposed by a partner of openEO on basis of the [OGC Coverage Implementation Schema](http://docs.opengeospatial.org/is/09-146r6/09-146r6.html#18).

dataset-spec/README.md
m-mohr: Great, then I'd like to have your support here (regarding global extensions): #186 ;-)

Sure, I think more issues would make sense! I'd like a proposal on this. Would you mind doing that? I don't think it's simply copy and paste (minus url) from the assets sepc in the items?

dataset-spec/README.md
cholmes: I think I prefer possible extent. That's a lot easier to implement for static providers. And someone who wants it to be 'current extent' can do that if they want - the current extent is certainly within the possible extent.

I don't want catalogs to feel like they have to always be updating this field.

cholmes: I lean towards 'possible extent'. An implementor could choose to make theirs 'current extent', since the current should be a subset of the possible. But I think it's better to not require static catalogs to keep updating their extent every single time there's new data outside their current.

m-mohr:
Another issue we should have in mind: #150

dataset-spec/README.md
m-mohr: "possible" vs "current" also applies for the temporal_extent as open date ranges would always be "possible". I would like to have them and for consistency and the reasons mentioned above, I now slightly prefer "possible", too. But with open date ranges we are not compatible with WFS, see also https://github.com/opengeospatial/WFS_FES/issues/155.

dataset-spec/README.md
```diff
@@ -0,0 +1,62 @@
...
+| host            | Host Object                           | Storage Provider                | The organization that hosts the dataset.                     |
+| spatial_extent  | [GeoJSON Object](http://geojson.org/) | Spatial extent (required)       | The spatial extent covered by the dataset as [GeoJSON](http://geojson.org/) object. |
+| temporal_extent | string                                | Temporal extent (required)      | Temporal extent covered by the dataset. Date/time intervals MUST be formatted according to ISO 8601. ToDo: Support open date ranges |
```
cholmes: It could be really good to try to get compatibility with WFS on the extent fields. They use:
```
      "extent": {
        "spatial": [ 7.01, 50.63, 7.22, 50.78 ],
        "temporal": [ "2010-02-15T12:34:56Z", "2018-03-18T12:11:00Z" ]
      },
```
If we want we can try to influence them to adopt our convention, but we should have good reasoning.

m-mohr: * I'm perfectly fine with the additional level of hierarchy (i.e. putting them both in an object with the key extent).
* Spatial is fine, too.
* Temporal on the other hand is a problem. Open date ranges are missing. I already opened an issue on that: https://github.com/opengeospatial/WFS_FES/issues/155
* What is missing from your example is trs and crs. Do we need to support them?

cholmes: I just used 'example 4' from the WFS spec. It notes: 'Coordinate reference system information is not provided as the service provides geometries only in the default system (WGS84 longitude/latitude)'. So seems like we could just say for dataset we require default.

m-mohr: Okay, than that's fine. Curious what the WFS crew is coming up with for the other issues mentioned. Will change that in the dataset spec. Temporal extent is still a string for now and need to find out how they define 3D bboxes (= incl. the z-axis).

dataset-spec/README.md
m-mohr: Added "potential" to both extents.

dataset-spec/README.md
m-mohr: Should be mostly WFS compatible now.

dataset-spec/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+One topic of interest has been the search of datasets*, instead of within a dataset, i.e. in (sub-)catalogs, items and assets. [STAC](https://github.com/radiantearth/stac-spec) is focused on search within a dataset, but it includes some simple constructs to catalog datasets. This could be an independent spec that STAC uses, and others can also independently use, to describe datasets in a lightweight way.
+
+*\* There is no standardized name for the concept we are describing here. Others called it: dataset series (ISO 19115), collection (CNES, NASA), dataset (JAXA), dataset series (ESA), product (JAXA).*
+
+## Core
+
+| Element     | Type              | Name                            | Description                                                  |
+| ----------- | ----------------- | ------------------------------- | ------------------------------------------------------------ |
+| name        | string            | Identifier (required)           | Identifier for the dataset that is unique across the provider. The identi |
```
fredliporace: @m-mohr possible accidental text 'cut'

m-mohr:
I fleshed out the spec a bit more. 
Additionally, I cleaned up this PR to just contain the dimension extension and the dataset spec for now. 
I think it is time for a review to get this in so we can built and discuss better based on this first proposal.

hgs-trutherford:
This looks pretty good to me. I think reading through this got me a little closer helped me understand how the concepts of "catalog", "collection" and "dataset" are being handled.

I'm not sure it's clear where the extensions fit in? Are they just added as additional properties in the root dataset object? Is dimensions a dataset specific extension or does it also apply to Items?

It sounds like any extension (eo in particular was discussed) can be applied at the dataset level? If so, do we need to indicate that item metadata takes precedence over dataset metadata?

dataset-spec/README.md
```diff
@@ -0,0 +1,103 @@
+# STAC Dataset Spec
+
+[STAC Items](https://github.com/radiantearth/stac-spec/json-spec/) are focused on search within a dataset*. Another topic of interest is the search of datasets, instead of within a dataset.  The Dataset Spec is an independent spec that STAC Items are *strongly recommended* to use. Other parties can also independently use this spec to describe datasets in a lightweight way.
```
cholmes: Is it that STAC Items 'use' the spec? I guess 'use' is less clear to me, it seems to imply that they should maybe implement the fields? Maybe describe what 'use' means? Like 'STAC Items are recommended to provide a link to a dataset definition'?

dataset-spec/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+The Datasets Spec is a superset of the [Catalog Spec](../static-catalog/). I shares the same fields and therefore every Dataset is also a valid Catalog. Datasets can have both parent Catalogs and Datasets and child Items, Catalogs and Datasets.
```
cholmes: I -> It

Also not sure if 'is a superset' is the best way to explain it. I'd say more like 'extends the catalog spec'. Probably would add a bit of color, like 'extends the catalog spec with additional fields to describe the set of items in the catalog'.

dataset-spec/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+| Element  | Type     | Description                                                  |
+| -------- | -------- | ------------------------------------------------------------ |
+| spatial  | [number] | **REQUIRED.** Potential *spatial extent* covered by the dataset. West, north, east, south edges of the spatial extent. Only WGS84 longitude/latitude is supported. The list of four numbers can be extended to six numbers to support a 3D spatial extent. |
+| temporal | string   | **REQUIRED.** Potential *temporal extent* covered by the dataset. Date/time intervals MUST be formatted according to ISO 8601. Open date ranges are supported by omitting either the start or the end time. Example for data from the beginning of 2019 until now: `2009-01-01T00:00:00Z/`. |
+
+### Provider Object
+
+The object provides information about a provider. A provider is any of the organizations that created or processed the content of the dataset and therefore influenced the data offered by this dataset.
```
cholmes: I think you said somewhere that multiple providers are allowed? If so I'd make it a bit more explicit here. If not then we should provide guidance on which one of the 'any...that created or processed'.

dataset-spec/README.md
```diff
@@ -0,0 +1,103 @@
...
+| Field Name | Type   | Description                                                  |
+| ---------- | ------ | ------------------------------------------------------------ |
+| name       | string | **REQUIRED.** The name of the organization or the individual. |
+| url        | string | Homepage of the provider.                                    |
+
+###  Host Object
+
+The objects provides information about the storage provider hosting the data. 
+
+**Note:** The idea of storage profiles is currently [discussed](https://github.com/radiantearth/stac-spec/issues/148). Therefore, scheme, id and region may be removed from the final spec once this concept id introduced to STAC.
```
cholmes: typo: id -> is

dataset-spec/example-s2.json
```diff
@@ -0,0 +1,47 @@
...
+  "license": "proprietary",
```
cholmes: Is the S2 license 'proprietary'? If it is in our current definition maybe we should expand the definition a bit. Or ideally find the ideal spdx license. For landsat I think we just used https://spdx.org/licenses/PDDL-1.0.html as getting across the intent of US public domain stuff...

dataset-spec/README.md
m-mohr: Improved that as suggested.

dataset-spec/README.md
m-mohr: The Provider object provides information about a single provider, but the Dataset object can hold multiple providers in an array. I made that more clear in the provider field.

dataset-spec/example-s2.json
m-mohr: Honestly, I don't really know and don't feel in the position to decide that. That example is taken from GEE and I handled it the way GEE does.

dataset-spec/README.md
m-mohr: Fixed.

m-mohr:
Mostly fixed the issues mentioned by Chris. One more review required.

dataset-spec/example-s2.json
cholmes: Ok, cool. Super minor issue, so let's just leave it, and following GEE sounds good.

dataset-spec/README.md
cholmes: Makes sent, sounds great.

cholmes:
Looks good to me, and looks like my green review checkmark is still there. Just need one more review - @matthewhanson or @hgs-truthe01 ?

m-mohr:
I just added another change, which improves the WFS 3 compatibility as we are now using their specification for temporal extents with one exception. We were using ISO8601 date ranges before and WFS3 uses a two element array. We changed to two element arrays, but here it is allowed to change one of the elements to null to support open date ranges. Proposed that also in the WFS issue tracker, maybe we get that into WFS (but doesn't seem anybody is currently very interested in temporal extents ;-) ).

m-mohr:
@hgs-truthe01 We are using RFC3339 now. That's what I wanted to say with my latest comment. ;)
GeoJSON is not applicable here for temporal extents as it covers only a single timestamp. For spatial extents we are following WFS3, too, which doesn't allow GeoJSON. So I think we got the most suitable solutions from both.

dataset-spec/README.md
```diff
@@ -0,0 +1,103 @@
...
+| title       | string            | A short descriptive one-line title for the dataset.          |
+| description | string            | **REQUIRED.** Detailed multi-line description to fully explain the entity. [CommonMark 0.28](http://commonmark.org/) syntax MAY be used for rich text representation. |
+| keywords    | [string]          | List of keywords describing the dataset.                     |
+| version     | string            | Version of the dataset. [Semantic Versioning (SemVer)](https://semver.org/) SHOULD be followed. |
+| license     | string            | **REQUIRED.** Dataset's license(s) as a SPDX [License identifier](https://spdx.org/licenses/) or [expression](https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60) or `proprietary` if the license is not on the SPDX license list. Proprietary licensed data SHOULD add a link to the license text, see the `license` relation type. |
+| provider    | [Provider Object] | A list of data providers, the organizations which influenced the content of the dataset. Providers should be listed in chronological order with the most recent provider being the last element of the list. |
```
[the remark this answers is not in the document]
cholmes: Cool. Are you saying we should just mention that this will usually just be one provider? Or that we should shift from array to a single object?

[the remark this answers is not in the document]
simonff: I suggest shifting to single object to reduce compelxity. (In the internal EE catalog, I started with providers a list and found that this is not necessary in 99.9% of cases, and the remaining 0.1% is better handled by provenance.) I'm open to revisiting this, but I'd like to see more examples first.

dataset-spec/README.md
```diff
@@ -0,0 +1,103 @@
...
+The following types are commonly used as `rel` types in the Link Object of a Dataset:
+
+| Type    | Description                                                  |
+| ------- | ------------------------------------------------------------ |
+| self    | **REQUIRED.** *Absolute* URL to the dataset file itself. This is required, to represent the location that the file can be found online. This is particularly useful when in a download package that includes metadata, so that the downstream user can know where the data has come from. |
+| root    | URL to the root [STAC Catalog](../static-catalog/) or Dataset. |
+| parent  | URL to the parent [STAC Catalog](../static-catalog/) or Dataset. |
+| child   | URL to a child [STAC Catalog](../static-catalog/) or Dataset. |
+| item    | URL to a [STAC Item](../json-spec/).                         |
+| license | The license URL for the dataset SHOULD be specified if the `license` field is set to `proprietary`. If there is no public license URL available, it is RECOMMENDED to supplement the STAC catalog with the license text in separate file and link to this file. |
```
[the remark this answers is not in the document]
m-mohr: Why should we distinguish them? Don't see a good reason yet and we don't do it for items, too.

dataset-spec/README.md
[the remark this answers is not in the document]
m-mohr: I think this need more discussion, but I think an issue would be the better place as this PR is already very messy.

dataset-spec/README.md
```diff
@@ -0,0 +1,103 @@
...
+| description | string            | **REQUIRED.** Detailed multi-line description to fully explain the entity. [CommonMark 0.28](http://commonmark.org/) syntax MAY be used for rich text representation. |
+| keywords    | [string]          | List of keywords describing the dataset.                     |
+| version     | string            | Version of the dataset. [Semantic Versioning (SemVer)](https://semver.org/) SHOULD be followed. |
+| license     | string            | **REQUIRED.** Dataset's license(s) as a SPDX [License identifier](https://spdx.org/licenses/) or [expression](https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60) or `proprietary` if the license is not on the SPDX license list. Proprietary licensed data SHOULD add a link to the license text, see the `license` relation type. |
+| provider    | [Provider Object] | A list of data providers, the organizations which influenced the content of the dataset. Providers should be listed in chronological order with the most recent provider being the last element of the list. |
+| host        | Host Object       | Storage provider, the organization that hosts the dataset.   |
```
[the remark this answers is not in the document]
m-mohr: Open to add that, but I think we can add that with another PR shortly after this PR.

dataset-spec/README.md
```diff
@@ -0,0 +1,103 @@
...
+| item    | URL to a [STAC Item](../json-spec/).                         |
+| license | The license URL for the dataset SHOULD be specified if the `license` field is set to `proprietary`. If there is no public license URL available, it is RECOMMENDED to supplement the STAC catalog with the license text in separate file and link to this file. |
+
+## Extensions
+
+Important related extensions for the dataset spec:
+
+* [EO extension](../extensions/stac-eo-spec.md)
+  Please note that some fields such as `eo:sun_elevation ` or `eo:sun_azimuth` are only meaningful on the item level and MUST not be used in datasets.
+* [Dimensions extension](../extensions/dimensions)
```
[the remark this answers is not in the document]
m-mohr: We have at least a draft for this extension in contrast to provenance. And I have data for the dimension extension. I'll ask my colleague to publish that. But we can move that to a separate PR, if that's preferred by you or others.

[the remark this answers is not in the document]
simonff: Right, I'd rather reduce this PR and then calmly discuss the change separately.

dataset-spec/README.md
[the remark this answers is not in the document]
simonff: Having to create a local license file means that the data provider has not organized their licenses well and may change them later, making local copies obsolete. But this is speculative, so I'm okay with not changing this now.

dataset-spec/README.md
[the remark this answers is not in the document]
m-mohr: Okay, will separate when I'm back in the office in 9 hours ;)

dataset-spec/README.md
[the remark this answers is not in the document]
m-mohr: Okay, great. Please, make sure to open an issue on this one so we don't forget.

dataset-spec/README.md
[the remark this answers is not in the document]
m-mohr: Dimensions extension has been moved to PR #227.

m-mohr:
Moved Dimensions extension as requested. Can be merged now, I think.

PR #166, opened 2018-08-15
Add fields query parameter fragment

pwideman:
Adding the yaml openapi fragment for the new fields query parameter.


api-spec/Fields-fragment.yaml
```diff
@@ -0,0 +1,63 @@
...
+        The optional fields parameter defines the shape of the output features. 
+
+        The geometry member determines whether the geometry is populated or is null. The
+        include and exclude members specify an array of property names that are either 
+        included or excluded from the result, respectively. If both include and exclude 
+        are specified, include takes precedence.
+      required: false
+      schema:
+        type: object
+      properties:
```
jasongilman: This should be indented one level. Also I think you can replace the schema with a ref potentially.

Issue #131, opened 2018-07-18
Consider having API spec use GeoJSON spec for geometry


metasim:
Currently the `api-spec` schema defines it's own geometry types, while the static spec depends/defers on the `geojson.json` schema.

https://github.com/radiantearth/stac-spec/blob/b7fe13e5d9aab371ead0e80a9c9cb7c97c49ec5c/api-spec/STAC-fragment.yaml#L174-L217

Furthermore, the structure of the `api-spec` types appear to be different from the GeoJson types. 

Consider having a single set of geometry types across all schemas.

cholmes:
cc @hgs-msmith and @hgs-trutherford 

I think the intent was to have the geometries be the same as GeoJSON - we just couldn't refer to the geojson.json directly as its a yaml document. 

@metasim - could you lay out what where the types between api-spec and geojson diverge? A PR to change it would be great, unless Michael and Tim have a reason it's the way it is.

hgs-trutherford:
@cholmes I think one of the early versions of the STAC spec had that geojson schema and we just leveraged it. If there is a better way to indicate a goejson object, I am all for it.

metasim:
@cholmes I'll try to take a look at it soon, what a api-spec mod would look like... I'm not all that adept with JSON Schema (more of an XSD background), so I'm not as confident on what I'd come up with. Excuses aside, here's where it seems the `stack-item.json` definition "defers" to GeoJSON.


https://github.com/radiantearth/stac-spec/blob/b7fe13e5d9aab371ead0e80a9c9cb7c97c49ec5c/json-spec/json-schema/stac-item.json#L19

hgs-trutherford:
The new fragment model we added should help with this. We can pull the GeoJSON definition out into a separate file. I think the yaml library is intended to support JSON schema as well? @scisco

hgs-truthe01:
#253 Should help fix this. I chose to use the schema definitions hosted by geojson.io, but if there is a preference for schemastore, I can probably switch over.

m-mohr:
Seems to be fixed.

PR #167, opened 2018-08-15
Add query support

hgs-trutherford:
Removing swagger2 to standardize on openapi3

Adding a simple query language

Adding a catalog definition endpoint at /stac

search updated to /stac/search to match new structure


api-spec/WFS3core+STAC.yaml
```diff
@@ -1,842 +1,909 @@
-openapi: 3.0.1
-info:
-  title: STAC API integrated with OGC Web Feature Service standard
-  version: 0.5.1
-  description: >-
-    This is an example OpenAPI specification of a SpatioTemporal Asset Catalog
```
hgs-trutherford: I think we should put the title and description overrides back into the WFS fragment.

hgs-trutherford:
The order that things appear in the docs is different since they are now generated from multiple fragments. 

Also, the README still needs to be updated with information about the merge process and tooling.
@scisco Did you say you were up for the README update?

api-spec/extensions/query.fragment.yaml
```diff
@@ -0,0 +1,74 @@
...
+    queryProp:
+      description: Apply query operations to a specific property
+      anyOf:
+        - type: object
+          description: if the object doesn't contain any of the operators, it is equivalent to using the equals operator
+        - type: object
+          description: Match using an operator
+          properties:
+            eq:
+              type: object
```
m-mohr: I think we should omit the type here. No type means any type (see https://github.com/OAI/OpenAPI-Specification/issues/1388) and I think that's what you meant by type object.

api-spec/extensions/query.fragment.yaml
```diff
@@ -0,0 +1,74 @@
...
+              description: Find items with a property that is equal to the specified value
+            neq:
+              type: object
```
m-mohr: Same as above, omit the type.

api-spec/extensions/query.fragment.yaml
```diff
@@ -0,0 +1,74 @@
...
+            lt:
+              type: number
+              description: Find items with a property value less than the specified value
+            gte:
+              type: number
+              description: Find items with a property value greater than or equal the specified value
+            lte:
+              type: number
+              description: Find items with a property value greater than or equal the specified value
+            startsWith:
```
m-mohr: Are startsWith, endsWith and contains case-sensitive or insensitive?

api-spec/extensions/query.fragment.yaml
```diff
@@ -0,0 +1,74 @@
...
+  /stac/search:
+    get:
+      parameters:
+        - $ref: '#/components/parameters/query'
+components:
+  parameters:
+    query:
+      name: query
+      in: query
+      description: query for properties in items. Use the JSON for of the queryFilter used in POST.
```
m-mohr: "for of" ?

api-spec/definitions/WFS3core.fragment.yaml
```diff
@@ -0,0 +1,385 @@
+paths:
```
m-mohr: I did NOT review this file as I hope that it is just taken from the WFS spec and doesn't need further review.

api-spec/definitions/STAC-standalone.yaml
```diff
@@ -1,355 +1,543 @@
...
-            client should utilize the `nextPageToken`.
-          example: http://api.cool-sat.com/query/gasd312fsaeg/ANsXtp9mrqN0yrKWhf-y2PUpHRLQb1GT-mtxNcXou8TwkXhi1Jbk
-tags:
-  - name: STAC
-    description: 'Extension to WFS3 Core to support STAC metadata model and search API'
\ No newline at end of file
+openapi: 3.0.1
+info:
+  title: The SpatioTemporal Asset Catalog API (standalone)
+  version: 0.5.1
```
m-mohr: Outdated.

@cholmes How do we make sure we always update all the version numbers hidden in lots of files?

api-spec/definitions/STAC-standalone.yaml
```diff
@@ -1,355 +1,543 @@
...
+    specification. Any service that implements this endpoint to allow search of
+    spatiotemporal assets can be considered a STAC API. The endpoint is also
+    available as an OpenAPI fragment that can be integrated with other OpenAPI
+    definitions, and is designed to slot seamless into a WFS 3 API definition.
+  contact:
+    name: Cool Sat Corp
+    email: info@cool-sat.com
+    url: 'http://cool-sat.com/contact/'
+  license:
+    name: CC-BY 4.0 license
```
m-mohr: Should probably be Apache-2.0 by default, which is the STAC license.

api-spec/definitions/STAC-standalone.yaml
```diff
@@ -1,355 +1,543 @@
...
+info:
+  title: The SpatioTemporal Asset Catalog API (standalone)
+  version: 0.5.1
+  description: >-
+    This is an OpenAPI definition of the core SpatioTemporal Asset Catalog API
+    specification. Any service that implements this endpoint to allow search of
+    spatiotemporal assets can be considered a STAC API. The endpoint is also
+    available as an OpenAPI fragment that can be integrated with other OpenAPI
+    definitions, and is designed to slot seamless into a WFS 3 API definition.
+  contact:
```
m-mohr: I'd remove the placeholder data or put in something useful.

api-spec/definitions/STAC-standalone.yaml
```diff
@@ -1,355 +1,543 @@
...
+    name: Cool Sat Corp
+    email: info@cool-sat.com
+    url: 'http://cool-sat.com/contact/'
+  license:
+    name: CC-BY 4.0 license
+    url: 'https://creativecommons.org/licenses/by/4.0/'
+servers:
```
m-mohr: I'd remove the placeholder data or put in something useful.

api-spec/definitions/STAC-standalone.yaml
```diff
@@ -1,355 +1,543 @@
...
+      explode: false
+    time:
+      name: time
+      in: query
+      description: >
+        Either a date-time or a period string that adheres to RFC3339. Examples:
+
+        * A date-time: "2018-02-12T23:20:50Z" 
+
+        * A period: "2018-02-12T00:00:00Z/2018-03-18T12:31:12Z" or
```
m-mohr: I know that this was taken from the WFS spec, but periods like this is not defined in RFC3339 and therefore are invalid. Those things originate from ISO 8601. See also https://github.com/opengeospatial/WFS_FES/issues/155

api-spec/definitions/STAC-standalone.yaml
```diff
@@ -1,355 +1,543 @@
...
+      properties:
+        code:
+          type: string
+        description:
+          type: string
+    link:
+      type: object
+      required:
+        - href
+      properties:
```
m-mohr: As below, I'd add `additionalProperties: true` to allow other fields such as type and title in the link, too.

api-spec/definitions/STAC-standalone.yaml
```diff
@@ -1,355 +1,543 @@
...
+    searchBody:
+      description: The search criteria
+      type: object
+      allOf:
+        - $ref: '#/components/schemas/bboxFilter'
+        - $ref: '#/components/schemas/timeFilter'
+        - $ref: '#/components/schemas/intersectsFilter'
+        - type: object
+          properties:
+            limit:
```
m-mohr: What would be the default limit?

api-spec/definitions/STAC-standalone.yaml
```diff
@@ -1,355 +1,543 @@
...
+      properties:
+        time:
+          $ref: '#/components/schemas/time'
+    intersectsFilter:
+      type: object
+      description: Only returns items that intersect with the provided polygon.
+      properties:
+        intersects:
+          $ref: '#/components/schemas/polygon'
+    polygon:
```
m-mohr: Can we just reference GeoJSON geometries here? See also #131

api-spec/definitions/STAC-standalone.yaml
```diff
@@ -1,355 +1,543 @@
...
+      description: >
+        Either a date-time or a period string that adheres to RFC 3339.
+        Examples:
+
+
+        * A date-time: "2018-02-12T23:20:50Z"  
+
+        * A period:
+
+        "2018-02-12T00:00:00Z/2018-03-18T12:31:12Z" or
```
m-mohr: See above, same issue with the periods.

api-spec/definitions/STAC-standalone.yaml
```diff
@@ -1,355 +1,543 @@
...
+        `time` are selected.
+
+
+        If a feature has multiple temporal properties, it is the decision of the
+
+        server whether only a single temporal property is used to determine the
+
+        extent or all relevant temporal properties.
+      example: '2018-02-12T00:00:00Z/2018-03-18T12:31:12Z'
+    catalogDefinition:
```
m-mohr: See below, same issues as in the other catalog definition. Can we share schemas? Maybe even just reference the (hopefully updated and valid) JSON schema of the catalog?

api-spec/STAC-query.yaml
```diff
@@ -0,0 +1,692 @@
+openapi: 3.0.1
```
m-mohr: This is generated from the fragments, right? So no review required?

api-spec/STAC-standalone.yaml
```diff
@@ -1,485 +1,542 @@
...
-          description: >-
-            A URL to obtain the next paginated data set. If not present, the
-            client should utilize the `nextPageToken`.
-          example: >-
-            http://api.cool-sat.com/query/gasd312fsaeg/ANsXtp9mrqN0yrKWhf-y2PUpHRLQb1GT-mtxNcXou8TwkXhi1Jbk
-tags:
-  - name: STAC
-    description: Extension to WFS3 Core to support STAC metadata model and search API
\ No newline at end of file
+openapi: 3.0.1
```
m-mohr: This is generated from the fragments, right? So no review required?

api-spec/definitions/WFS3core.fragment.yaml
```diff
@@ -0,0 +1,385 @@
...
+            - 'http://www.opengis.net/def/uom/ISO-8601/0/Gregorian'
+          default: 'http://www.opengis.net/def/uom/ISO-8601/0/Gregorian'
+        temporal:
+          description: Begin and end times of the temporal extent.
+          type: array
+          minItems: 2
+          maxItems: 2
+          items:
+            type: string
+            format: dateTime
```
m-mohr: We inherited this problem from WFS (see https://github.com/opengeospatial/WFS_FES/issues/155 - fourth issue), but we may change this anyway from `dateTime` to `date-time`? According to specs the latter is correct. See also https://github.com/radiantearth/stac-spec/issues/146

api-spec/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "description": "Generate STAC openapi defintions from fragments.",
+  "main": "index.js",
+  "scripts": {
+    "generate-standalone": "yaml-files definitions/STAC-standalone.yaml STAC-standalone.yaml",
+    "generate-query-ext": "yaml-files definitions/STAC-query.yaml STAC-query.yaml",
+    "generate-wfs": "yaml-files definitions/WFS3core+STAC.yaml WFS3core+STAC.yaml",
+    "generate-all": "npm run generate-standalone && npm run generate-wfs && npm run generate-query-ext"
+  },
+  "author": "",
+  "license": "ISC",
```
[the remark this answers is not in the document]
hgs-trutherford: Oops, I forgot to update that from the default that npm generates. It looks like the json-spec package file has the same issue. I'm inclined to just remove it from the package json all together and inherit the LICENSE file at the top. This is not intended to be published as an npm module, so I think that is OK?

api-spec/package-lock.json
```diff
@@ -0,0 +1,73 @@
+{
```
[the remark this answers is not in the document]
hgs-trutherford: I tend to agree with you, but I try to follow the npm recommendations to keep things consistent. I would default to leaving it in unless there is a technical issue that comes up.

api-spec/extensions/query.fragment.yaml
hgs-trutherford: I'm not sure how I missed that as a feature! I was struggling with an "any" type and this makes it much easier! Thanks

api-spec/definitions/STAC-standalone.yaml
hgs-trutherford: I didn't change that in this PR. I'm happy to update it, but I'm not sure a license change should be hidden in this PR?

api-spec/definitions/STAC-standalone.yaml
hgs-trutherford: Another copy paste. I'm hesitant to overload this PR with changes since it is already so big.

api-spec/definitions/STAC-standalone.yaml
hgs-trutherford: It is defined in the RFC though: take a look at the periods section:
Periods:

   period-explicit   = iso-date-time "/" iso-date-time
   period-start      = iso-date-time "/" duration
   period-end        = duration "/" iso-date-time

   period            = period-explicit / period-start / period-end

https://tools.ietf.org/html/rfc3339

api-spec/STAC-query.yaml
hgs-trutherford: correct

api-spec/STAC-standalone.yaml
hgs-trutherford: correct

api-spec/extensions/query.fragment.yaml
hgs-trutherford: typo, meant to be "form of"

api-spec/definitions/STAC-standalone.yaml
hgs-trutherford: default is 10, but I think we can do better on consolidating the schema. Probably a separate PR.

api-spec/definitions/STAC-standalone.yaml
hgs-trutherford: I would like to. I didn't really want to carry the catalogDefinition around. I'll take a stab at referencing it instead.

hgs-trutherford:
pushing up a few small changes. I want to do another pass with @m-mohr suggestions in mind.

api-spec/WFS3core+STAC.yaml
m-mohr: Do you still want to do that or is that a todo for later?

api-spec/package-lock.json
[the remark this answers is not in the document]
m-mohr: Fair enough, we'll see how it works out.

api-spec/extensions/query.fragment.yaml
m-mohr: I think case insensitive search gives better results. You never really know whether people will write "landsat" or "Landsat" or "LANDSAT". Saw all of them in API implementations for openEO and we would only get one result with case-sensitive. Sure, case insensitive is more work to be implemented. Let's make an issue out of it, then people can discuss.

api-spec/definitions/WFS3core.fragment.yaml
m-mohr: I assume we'll update this anyway every now and then to be compatible with WFS.

api-spec/package.json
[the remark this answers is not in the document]
m-mohr: Open to all solutions that don't claim any other specific license than Apache-2.0.

api-spec/definitions/STAC-standalone.yaml
m-mohr: Agree, made issue #201 to track this.

api-spec/definitions/STAC-standalone.yaml
m-mohr: Okay, fine for now.

m-mohr:
Thanks, the most important issues were solved by @hgs-trutherford. I opened some related issues (#200, #201) and after merging Tim wants to do another pass with the minor issues in mind.

I think I'd still squash the commits. All these "Update Menlo_Notes.md" commits would clutter the history with "private" commits...

api-spec/extensions/query.fragment.yaml
cholmes: I sounded in on the issue - I think case insensitive makes a bit more sense, but don't feel strongly.

m-mohr:
Tried to fix the merge conflicts. Could you re-check, @hgs-trutherford?

cholmes:
Agreed that squashing the commits would be good. Could you do that @hgs-trutherford and then I can merge?

api-spec/extensions/query.fragment.yaml
hgs-trutherford: I think I'm OK with stating case insensitive. This is intended to be a simplified search, so I'm not sure it makes sense to separate items by case

hgs-trutherford:
@cholmes I went ahead and merged. I hope that is OK. I saw you both had it approved. I'm happy to make additional changes as additional PRs.

PR #172, opened 2018-08-15
adding /stac catalog entry

hgs-trutherford:
Adding a catalog json endpoint to allow stac crawlers to interact with a dynamic catalog.


hgs-trutherford:
I took an initial stab at creating the catalog schema from the example file, but I think it should be updated based on the feedback from the sprint.

api-spec/STAC-standalone.yaml
```diff
@@ -349,6 +362,51 @@ components:
 
         extent or all relevant temporal properties.
       example: '2018-02-12T00:00:00Z/2018-03-18T12:31:12Z'
+    catalogDefinition:
+      type: object
+      properties:
```
m-mohr: We should add the required fields:
```
required:
- name
- description 
- links
```

api-spec/STAC-standalone.yaml
```diff
@@ -349,6 +362,51 @@ components:
...
+        name:
+          type: string
+          example: NAIP
+        description:
+          type: string
+          example: Catalog of NAIP Imagery
+        license: 
+          type: string
+          example: PDDL-1.0
+        links:
```
m-mohr: We should add `additionalProperties: true` to allow additional properties like type or title.

api-spec/STAC-standalone.yaml
```diff
@@ -349,6 +362,51 @@ components:
...
+          items:
+            type: string
+        keywords:
+          type: array
+          items:
+            type: string
+        homepage:
+          type: string
+        provider:
+          type: object
```
m-mohr: `additionalProperties: true` should be added, too.

api-spec/STAC-standalone.yaml
```diff
@@ -21,7 +21,20 @@ servers:
   - url: 'http://www.cool-sat.com'
     description: Production server
 paths:
-  /search/stac:
+  /stac:
```
m-mohr: The geometry in line 442 could also refer to the GEOJSON geometry, i.e.
```
        geometry:
          $ref: "http://json.schemastore.org/geojson.json#/definitions/geometry"
```
But that's a nice to have and by no way blocking. May also just be solved with #131.

api-spec/STAC-standalone.yaml
hgs-trutherford: I'm a little confused on where you wan't me to add additionalProperties here. license is a string and links is an array. I am adding it at the catalogDefinition level which I think is what you wanted on your comment below as well?

m-mohr: Was meant for the links object, seems I marked the wrong line, sorry.

hgs-trutherford:
adding suggested changes
 @m-mohr is that what you meant with additional link properties?

m-mohr:
It actually also makes sense in the catalog object where you put it, but one of them I was expecting in the provider object. I just added that. If there is some reason against it, just let me know. Other than that, great changes. Can be merged.

cholmes:
Awesome, merging in.

PR #173, opened 2018-08-15
removing references to swagger2

hgs-trutherford:
standardize on openapi 3


cholmes:
Crossing my fingers that tooling for auto-generation from openapi 3 catches up soon, and if we get to a 1.0 and it hasn't then we should make a swagger version. But for now it'll be much easier to iterate in one.

Issue #139, opened 2018-08-01
Define more asset/link attributes & define a core set of types


cholmes:
There hasn't been a clear standard of what additional attributes / field names to include in STAC Items. We have enough real world examples that we should be able to set down more than just the href and rel link.

- [x] add type attribute for assets #175 
- [x] add type attribute for links (assets do mime_type, wfs links do just 'type' but use the mime type - need to decide which to do, and revisit name for assets too)
- [x] add 'title' attribute for links and assets. #263 
- [x] Define common types #180 (perhaps expand that PR to cover links too)
- [x] update examples for these.

We also should define a set of core types that are commonly used. Is it geotiff / geotif / cog, etc. See #54 as well, as it has some discussion of this. The list should start as just a best practice recommendation, and keep it open to evolve. But align initial tooling around those types.

m-mohr:
Having a (required?) title would help humans to explore the catalogs more easily as you can only guess from rel and href what the referenced entity is about. I'd like to reference two related issues here for the discussion: https://github.com/mojodna/stac-browser/issues/2 and https://github.com/mojodna/stac-browser/issues/4

Having a title (and a type) would also comply to RFC5988.

cholmes:
I don't think title should be required, but I do think we should have it speced.

m-mohr:
Yes, I fully agree.

Should we change that in the assets, too? There it is already specified as 'name'.

Issue #138, opened 2018-08-01
Format field in the catalog


m-mohr:
There's a `format` field in the catalog. It lacks documentation and it is not clear what exactly should be included there. Is it the union of all formats that are available throughout the items/assets? The example names "cog" and "geotiff", but who standardizes these names? Users may also use MIME-types or file extensions.

For example GeoTiffs could be referred to as:
* tif or tiff (file extension)
* gtiff (GDAL short name)
* geotiff (somehow self describing name from the example)
* image/tiff, image/tiff;subtype=geotiff or similar (MIME type)
* cog (arbitrary names of special GeoTiff flavors)
* probably more

m-mohr:
Seems to be related to some discussions in #54.

m-mohr:
Removed with #175.

PR #175, opened 2018-08-15
add type to asset

jeffnaus:
[issue #139](https://github.com/radiantearth/stac-spec/issues/139) - specifically requests a required mime type to be added to assets.  This is what this PR does

[issue #138](https://github.com/radiantearth/stac-spec/issues/138) - Asks about a format field in the catalog.  I've removed this from the catalog as it doesn't make sense to repeat this information from the item assets

### DIFF
--- a/json-spec/json-schema/stac-item.json
+++ b/json-spec/json-schema/stac-item.json
@@ -137,13 +137,17 @@
     "asset": {
       "type": "object",
       "required": [
-        "href"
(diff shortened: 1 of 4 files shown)


cholmes:
Still needs to update all the examples. 

I'd also like to see a table of common mime-types, for recommendations on the types people will actually use. 

But will merge and we can continue these on dev.

Issue #44, opened 2017-12-07
Non Polygon geometry types


mikeskaug:
Great work on this! It came along at just the right time.

We just started working towards implementing a static catalogue using the STAC spec. I should start off by saying that we have non-imagery data, which I realize is not the first priority for the STAC spec. The only part of the core STAC spec which causes us problems is the requirement that the geometry be of either Polygon or MultiPolygon type.

We have data collected as point measurements, so it would make the most sense to use a GeoJSON Point as the geometry type for each Item. We could define a polygonal envelope around each point to satisfy the requirement that the geometry be a Polygon or MultiPolygon, but it would be somewhat ambiguous how we define the polygon.

joshfix:
Definitely would be good to include more geometries.  Do you have any bandwidth to add them to the swagger doc and submit a PR?

cholmes:
Thanks Mike! 

Can you describe the data a bit more? We were definitely imagining that all the 'assets' in a catalog would represent more than a single point. That they really point to some other file that people download, that is the thing they want. So I'm curious what that 'thing' is in your case - the asset that is represented by the point. And like is there a thumbnail? Something to browse? That was another requirement.

If everything of interest at the point can be represented as fields, instead of another file download, then I'm not sure STAC is appropriate. Like if it's just an observation that can be described in a normal GeoJSON / Shapefile. But I'd be super into exploring a 'spatiotemporal feature' static thing with you - represent your data online as linked geojson to be crawled. And then STAC would just be an extension of the spatiotemporal feature that includes assets & a thumbnail.

mikeskaug:
The data exists as several different formats at different points in the processing, but each one is a multidimensional array associated with a single geographic point. So the 'asset' in our case is some big binary file storing the n-dimensional array. To put it in the imaging context, imagine what you would get with a hyperspectral camera with one pixel. Eventually we reach a gridded data product that can be saved as geotiff and would fit the STAC spec perfectly, but we have several levels of data before that to organize, search and manage. 

I guess the thumbnail is also an issue. For the specific data I'm working on now, it's natural to represent it in an image format, although the width/height dimensions are not spatial, but in general it might be difficult to think of a thumbnail representation.

cholmes:
Very cool. I _think_ it makes sense to fit that into STAC, and I really like the idea of representing the full derivation from the gridded data product back to the multidimensional array. Though I think that could happen even if the 'point thing' with assets wasn't technically stac, but used the same constructs. 

[...]

PR #177, opened 2018-08-15
remove restriction to only polygon multi-polygon for a STAC item feature

jeffnaus:
This keeps us from having a custom (restricted) type of geojson.  This makes the documentation simpler and just removes the need to check for specific geometry types in the json schema.

### DIFF
--- a/json-spec/json-schema/stac-item.json
+++ b/json-spec/json-schema/stac-item.json
@@ -25,21 +25,10 @@
           "required": [
             "id",
             "type",
-            "geometry",
             "links",
             "assets"
           ],
           "properties": {
-            "geometry": {
-              "properties": {
-                "type": {
-                  "enum": [
-                    "Polygon",
-                    "MultiPolygon"
-                  ]
-                }
-              }
-            },
             "id": {
               "title": "Provider ID",
               "description": "Provider item ID",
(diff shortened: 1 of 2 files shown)


json-spec/json-schema/stac-item.json
```diff
@@ -25,21 +25,10 @@
           "required": [
             "id",
             "type",
-            "geometry",
             "links",
             "assets"
           ],
           "properties": {
-            "geometry": {
```
m-mohr: I think removing the geometry property altogether is wrong. It should be still required, but the enum should be extended to all types supported by GeoJSON.

jeffnaus: This only removes the geometry from the STAC Item schema, it still exists in the GeoJSON schema and it is still required that a STAC item be a GeoJSON object.

jeffnaus: We only had the geometry defined in the STAC item schema to hold the polygon/multipolygon only restriction.  Now that that restriction is removed, we can remove the redundant geometry definition.

m-mohr: Oh, thanks for the clarification. Didn't know that we inherit from the GeoJSON spec.

cholmes:
Just saw #44  and it did remind me of my fear of people using STAC to represent vector data. I think I can get over it, and am compelled by the reasoning that this makes things cleaner - we can just use the geojson spec directly. Making 1 asset required as discussed in #187 is one thing that does help in preventing people from just putting in vector data.

cholmes:
Merging in. Still one little thing to do on #193 to get this completely in, but happy to put this on dev.

PR #183, opened 2018-08-16
Add ci checks for api

scisco:


m-mohr:
@scisco Is the CI broken? #186 and #185 for example don't get a result, which would block merging them.

scisco:
@m-mohr The CI in this PR is broken because my changes have an error that I have to fix. 

In the other two PRs that you mentioned #186 and #185 the build has not started because the user who has committed does not have an account on CircleCI. You should login to circleci using your github account, then add a new commit to the PR (branch) which will start the build

m-mohr:
Oh, that sounds overly complicated and was not mentioned as requirement on sprint 3 AFAIK. TravisCI doesn't require this. Can this be simplified or do we require all contributors to go through this steps? Also, I don't like to clutter the commit history...

scisco:
We could certainly switch to Travis. What I explained only applies to PRs from forked repos btw. I think Travis has the same issue though. The problem is that the CI environment needs to checkout the code, for which it needs the necessary permission. When you login, you give the CI access to the forked repo on your account.

The new commit is only needed for the first time, because it's not possible to trigger a build for commits in the past.

matthewhanson:
Yeah I'm pretty sure the same problem exists in Travis because it needs to build an individuals repo when forked and isn't granted permission by default.

PR #188, opened 2018-08-21
Allow 0 for cloud cover.

m-mohr:
During STAC sprint 3 we discussed that it seems wrong that cloud cover can't be set to 0, i.e. no clouds.
@matthewhanson


matthewhanson:
ha, yeah that's an oversight, definitely need to be able to set it to 0.

Issue #203, opened 2018-08-27
README for the api generation tools.


hgs-truthe01:
We should have a readme with instructions and a description of how the yaml fragments work.

cholmes:
@hgs-truthe01 - wondering if you could take this? Or do we need more from @scisco ? Perhaps @matthewhanson can help.

cholmes:
@mojodna - if you can take this on too that'd be great. May take a bit of detective work. But it'd be good to describe how to properly edit OpenAPI files to work for STAC and how our system builds them.

m-mohr:
Done with #432.

PR #208, opened 2018-08-28
Adding an overview of the fragments system

hgs-truthe01:
This is to address #203 

This is a first pass at documenting our thinking on the fragment system. I wanted to get something in the readme to provide some guidance. It could definitely use more content on how to author an extension/fragment.


m-mohr:
Definately good to have this description.

I think it could be a good idea to separate "Developer" things and the API specification into separtate documents. Not sure whether the average user is interested in the more internal technical stuff for development of the spec itself.

m-mohr:
Oh and we should somehow add the generation process to the CI. I just ran the commands and one of the generated files doesn't seem to be up-to-date. We should ensure they are always up-to-date.

Issue #197, opened 2018-08-24
Change 'links' back to a list from a dictionary.


cholmes:
There was lots of discussion at the sprint on lists vs dictionary. At sprint 2 we changed links to a dict in #114 But in #127 we realized we had gone a bit too far. At the sprint we discussed going back to lists for everything, see #151  The conclusion from everyone is that we should definitely shift links back to lists. 

I believe 'assets' is still up for debate, so will just limit this issue to the links. But we may want to update assets too.

PR #219, opened 2018-08-29
Change 'links' back to a list from a dictionary (#197)

m-mohr:


m-mohr:
Sorry, you need to review again. Seems there were some dependencies between #207 and #202, which screwed things up. Changes are the same as in #207.
Lessons learnt: Never make a PR based on another PR.

m-mohr:
Oh, I think that was just my fault yesterday night. Shouldn't work that late from home. ;-) Thanks for approving again.

m-mohr:
Was reviewed in #207.

Issue #205, opened 2018-08-28
Restructuring the repository


m-mohr:
As the spec and repository grows, things get structured based on historical decisions and don't reflect the current state of the spec well enough. As discussed on Gitter, there are several things that - IMHO - need to be restructured / changed:

**Root:**
- [x] Rename folder json-spec to item-spec
- [x] Rename folder static-catalog to catalog-spec
- [x] README: Update to reflect current changes (maybe shorten things a bit or put things in separate files)
- [x] Separate Contributing guidelines
- [x] how-to-help: Update to reflect current changes
- [ ] roadmap: Update to reflect current changes (#296)
- [x] Remove package-lock.json (no corresponding package.json available?)

**API spec:**
- [x] README: Update to reflect current changes
- [ ] Review naming (see below)
- [ ] Separate into api-spec.md and README.md (see Item spec) (#297)

**Extensions:**
- [x] Move EO extension + examples to a separate folder
- [x] Move Collection extension + examples to a separate folder (wait for #174)
- [x] Move start-end-datetime extension to a separate folder (wait for #221)

I think all extensions should reside in a separate folder. They should consist of at least three files anyway, which is enough to get a separate folder to not clutter the extensions folder.

**JSON/Item spec:**
- [x] Move sample and sample-full to examples folder
- [x] README: Rework completely to reflect that it is the item spec folder, moving some stuff to the top level (as it applies to all specs).
- [x] Move package.json / package-lock.json to .cirecleci or wherever central validation info should be stored
- [x] Check whether we still need the json-schema/geojson.json
- [x] Kill the NAIP examples (#204)

**Static catalog / catalog spec:**
- [x] Needs a complete rewrite with the introduction of datasets.
- [x] Make a separate file for static catalog related information and check whether the static-recommendations can be added there.
- [x] Separate into catalog-spec.md and README.md (see Item spec)
- [ ] Review naming (see below)

**Dataset Spec:**
- [x] Separate into dataset-spec.md and README.md (see Item spec)

**General:**
- [ ] Format all examples using the same style, e.g. by using https://jsonformatter.org/ (#298)

There are probably things to be added, feel free to add them below.

matthewhanson:
+1, all looks good to me. I don't have any additional things off-hand, but I think we should probably do such a reorg sooner rather than later.

cholmes:
Just put this in a review comment on the relevant PR, but worth pulling up here:

[...]

PR #230, opened 2018-08-31
Restructured and improved validation

m-mohr:
Restructured and improved validation

- Moved everything in one folder to be usable for all STAC things (#205)
- Added a testable example for a basic catalog
- Restructured datasets to have an examples folder
- Adapted README files
- Made validation calls easier by using npm scripts
- Turned on CI for the most basic item, catalog, dataset and scientific extension examples
- Fixed other validation related bugs
- Removed the requirement for type in assets as it is not marked required in the written spec (or is this wrong in the written spec?).


m-mohr:
Yes, it is. On this branch it is in fact already checking the basic examples from each spec in the CI.

PR #231, opened 2018-08-31
Add Sentinel-2 example, make trival cleanup changes to several other example specs

philvarner:
title says it all.


cholmes:
Thanks @philvarner ! 

Curious if you're building a fuller sentinel2 catalog that this is a part of? Would love to include it if so.

Would you mind making the PR against dev? Our process is to do all development against dev, and then on 'release' merge to master. There has been some restructuring on dev, so everything may not come in super cleanly.

PR #233, opened 2018-08-31
Remove logos (moved to stac-site)

cholmes:
I originally put the logos here since this was the only STAC repo. Though it's still in progress we do have https://github.com/radiantearth/stac-site and that's probably a more appropriate place for logos. Plus it can make this repo a bit less busy.


cholmes:
Oh yeah, putting one on the readme is a good idea. I think I just uploaded them really quickly to use for blog posts, and hadn't looked at the readme for a bit. I'll do a PR now.

Issue #158, opened 2018-08-14
Update implementation list


metasim:
A number of implementations are popping up (e.g. DevelopmentSeed's sat-util, Element 84's cmr). It would be helpful to have a catalog of these listed in the `sat-api` section.

cholmes:
We have https://github.com/radiantearth/stac-spec/blob/master/implementations.md which is an obvious place, but is a bit out of date. It should likely include what version of STAC the implementations are at.

A bit nicer is https://github.com/radiantearth/stac-spec/blob/master/static-catalog/examples.md which we did since we removed the static catalog 'examples'. But the implementations should be merged in, and examples should maybe just refer to the github repos? Not sure, it is useful to see the real world examples - but maybe it just says to check the implementations. Could also make sense to do one small catalog / dataset example.

metasim:
Both are awesome for us.... the static more pertinent to what we want to do right now. Just having a destination is a great start. Close this?

cholmes:
I'd like to leave it open as a reminder to update before the next release...

m-mohr:
I made a PR (#238) to merge and update the implementations/examples.

m-mohr:
@cholmes , do you still want to update this, right? Or was my merge already all to do? Then this could be closed.

cholmes:
Ideally we have this updated with links to some actual 0.6.0 implementations that are made after the release candidate. And we should be sure to specify that the other implementations are still 0.5.0 (but likely updating soon)

cholmes:
Going to close this, as I'd like to move implementations to http://stacspec.org, so they're not like tied to releases, etc.

PR #238, opened 2018-09-05
#158: Merged, updated and restructured implementations

m-mohr:


implementations.md
```diff
@@ -2,75 +2,83 @@
...
+* [py-stac](https://github.com/raster-foundry/pystac) was used to create the ISERV catalog. It is still pretty tied to that data type.
+* [go-stac](https://github.com/planetlabs/go-stac) does validation of static catalogs.
+* [Open Imagery Network](https://openimagerynetwork.github.io/) was the first attempt at a static catalog. The OpenAerialMap 
+  bucket is in active use. It is planned to evolve OIN and OAM architectures to use static catalogs.
+* [AWS Public Datasets](http://aws.amazon.com/public-datasets/), particularly [Landsat](http://aws.amazon.com/public-datasets/landsat/), 
+  [NAIP](https://aws.amazon.com/public-datasets/naip/) and [Sentinel](http://sentinel-pds.s3-website.eu-central-1.amazonaws.com/) all
+  get close to the ideas of STAC static catalogs, and hopefully will evolve to implement the standard.
+* Josh Fix's [open-catalog](https://github.com/joshfix/open-catalog) demonstrates generating code from the core swagger spec.
+* Element84 made a [clojure implementation](https://github.com/Element84/catalog-api-spec/tree/dev/implementations/e84) of the spec that serves as a lightweight proxy to NASA's CMR, constrained to just data in USGS EROS.
```
fredliporace: @m-mohr the software that is used to update the CBERS STAC catalog as new scenes are ingested is available from https://github.com/fredliporace/cbers-2-stac. There is also a static STAC crawler in https://github.com/fredliporace/stac-crawler, which I'm using to populate sat-api from a static catalog. I may prepare a text using the same structure above covering both tools.

m-mohr: I just merged two files here and did not really update links or checked for up-to-dateness. But feel free to suggest updates.

m-mohr: Added to the list (see chapter Ecosystem).

Issue #94, opened 2018-04-05
Create JSON-schema for EO profile validation


cholmes:
It'd be good to have a JSON-Schema file that can be used to validate the Earth Observation profile, so people can check that they're compliant.

cc @jeffnaus & @matthewhanson

m-mohr:
I could definitely support here as I am familiar to both EO extension and JSON schema. Not sure how familiar @matthewhanson is with JSON schema?

matthewhanson:
Haven't dealt much with JSON schema, I'd take a crack at it because I did want to figure it out but won't have time until week after next. So if you get some spare time and want to do it have at it @m-mohr

m-mohr:
I'll put it on my list, but not sure either about time. We should first concentrate on getting the EO extension renamed and the other fields for the datasets added etc. (see https://github.com/justinwp/stac-spec/blob/feat/dataset/extensions/stac-eo-spec.md)

cholmes:
Yup, definitely lower priority on this one - let's focus on the 'must-haves' first. Plus all those things shifting might lead to tweaks to EO profile.

m-mohr:
New PR: #266

m-mohr:
Got merged, thanks @jeffnaus for the initial work and PR!

PR #240, opened 2018-09-05
#94: add EO extension JSON schema

jeffnaus:
Wrote JSON schema based off of https://github.com/radiantearth/stac-spec/blob/dev/extensions/stac-eo-spec.md.  I'm doing the same "inheritance" thing where an EO item is still a STAC item which is still a geoJSON feature.


extensions/json-schema/stac-eo-extension.json
```diff
@@ -0,0 +1,124 @@
...
+      "$ref": "#/definitions/eo"
+    }
+  ],
+  "definitions": {
+    "eo": {
+      "allOf": [
+        {
+          "oneOf": [
+            {
+              "$ref": "https://raw.githubusercontent.com/radiantearth/stac-spec/master/json-spec/json-schema/stac-item.json#/definitions/core"
```
m-mohr: The validator (ajv) we are using in this repository doesn't allow referencing schemas by URL, so we couldn't use this in the CI. Can this be changed to a relative URL? Otherwise I'd just remove the inheritance. Don't see a major advantage here. You could just validate against Item an EO extension separately.

extensions/json-schema/stac-eo-extension.json
```diff
@@ -0,0 +1,124 @@
...
+                  },
+                  "eo:sun_azimuth": {
+                    "title": "Sun Azimuth",
+                    "description": "Sun azimuth angle. From the scene center point on the ground, this is the angle between truth north and the sun. Measured clockwise in degrees (0-360).",
+                    "type": "number"
+                  },
+                  "eo:sun_elevation": {
+                    "title": "Sun Elevation",
+                    "description": "Sun elevation angle. The angle from the tangent of the scene center point to the sun. Measured from the horizon in degrees (0-90).",
+                    "type": "number"
```
m-mohr:
```
"minimum": 0
"maximum": 90
```
should be added.

extensions/json-schema/stac-eo-extension.json
```diff
@@ -0,0 +1,124 @@
...
+                  },
+                  "eo:azimuth": {
+                    "title": "Azimuth",
+                    "description": "Viewing azimuth angle. The angle measured from the sub-satellite point (point on the ground below the platform) between the scene center and true north. Measured clockwise from north in degrees (0-360).",
+                    "type": "number"
+                  },
+                  "eo:sun_azimuth": {
+                    "title": "Sun Azimuth",
+                    "description": "Sun azimuth angle. From the scene center point on the ground, this is the angle between truth north and the sun. Measured clockwise in degrees (0-360).",
+                    "type": "number"
```
m-mohr:
```
"minimum": 0
"maximum": 360
```
should be added.

extensions/json-schema/stac-eo-extension.json
```diff
@@ -0,0 +1,124 @@
...
+                  },
+                  "eo:off_nadir": {
+                    "title": "Off Nadir",
+                    "description": "Viewing angle. The angle from the sensor between nadir (straight down) and the scene center. Measured in degrees (0-90).",
+                    "type": "number"
+                  },
+                  "eo:azimuth": {
+                    "title": "Azimuth",
+                    "description": "Viewing azimuth angle. The angle measured from the sub-satellite point (point on the ground below the platform) between the scene center and true north. Measured clockwise from north in degrees (0-360).",
+                    "type": "number"
```
m-mohr:
```
"minimum": 0
"maximum": 360
```
should be added.

extensions/json-schema/stac-eo-extension.json
```diff
@@ -0,0 +1,124 @@
...
+                  },
+                  "eo:cloud_cover": {
+                    "title": "Cloud Cover",
+                    "description": "Estimate of cloud cover as a percentage (0-100) of the entire scene. If not available the field should not be provided.",
+                    "type": "integer"
+                  },
+                  "eo:off_nadir": {
+                    "title": "Off Nadir",
+                    "description": "Viewing angle. The angle from the sensor between nadir (straight down) and the scene center. Measured in degrees (0-90).",
+                    "type": "number"
```
m-mohr:
```
"minimum": 0
"maximum": 90
```
should be added.

extensions/json-schema/stac-eo-extension.json
```diff
@@ -0,0 +1,124 @@
...
+                  },
+                  "eo:epsg": {
+                    "title": "EPSG",
+                    "description": "EPSG code of the datasource's projection system.",
+                    "type": "integer"
+                  },
+                  "eo:cloud_cover": {
+                    "title": "Cloud Cover",
+                    "description": "Estimate of cloud cover as a percentage (0-100) of the entire scene. If not available the field should not be provided.",
+                    "type": "integer"
```
m-mohr:
```
"minimum": 0
"maximum": 100
```
should be added.

m-mohr:
By the way, I ran validation by calling `npm run validate_extension -- -s "../extensions/json-schema/stac-eo-extension.json" -d "../extensions/examples/landsat8-sample.json"`
Unfortunately, the landsat samples are not valid (missing eo:platform, eo:constallation and eo:instrument). We should add these fields to the examples, make sure they all validate and also add the above command to the validation readme.

extensions/json-schema/stac-eo-extension.json
jeffnaus: If we use relative references then I think we will need all the JSON schemas to live in the same path, i.e. we able to directly construct the path between the relative references.  We can do this but we'd have to change the layout to put all JSON schema in one path (like all in one folder), but then you'd either need to put your custom extension schemas there or everybody who uses your custom schema would need to build the appropriate reference resolver.  I really put in the absolute reference to get out of the need for custom reference resolvers. 

The advantage I see with this inheritance is that we don't need to know the relationship between the various schemas in order to perform the validation.  The idea here is that the schema itself says that this extension is extending these various schemas and adding in the specified components.

For instance you could write a schema that requires the item to be both a EO item and a item with the start/stop time extension, plus some other custom fields...  The schema would include these various definitions from the extension's schemas and enforce this without the end validation user needing to know to check these other schemas.

m-mohr: I understand all your points made here, but it just won't work with the validator we are using (see https://github.com/jessedc/ajv-cli/issues/29 ) and therefore we can't use it in CI. We either need to use another validator or remove the dependency to items for now.

jeffnaus:
@m-mohr I've added the min/max values you suggested.  Thanks again for the great proofing you are doing!

jeffnaus:
In testing I noticed that I had the eo:cloud_cover and other fields inside of the bands object and not inside of the properties objects.  I've fixed this in the latest commit.

m-mohr:
By the way, the related issue is #94.

m-mohr:
Needs to be adopted to the changes introduced by #247 (bands as arrays).

m-mohr:
I'll merge this to a development branch in this repository so that we can update the JSON schema by @jeffnaus to the most recent developments.

Issue #194, opened 2018-08-23
Create a generic 'dataset' specification, and use that to replace 'root catalog' concept.


cholmes:
There was lots of good discussion at STAC sprint 3 and continuing in github after on collection / dataset level metadata. 

Not everything is 100% sorted out, but we're homing in on a some good consensus.

The 'dataset' specification will be a more formal definition of the descriptive fields of the catalog (defined by the static catalog in the first sprint, and not really revisited much). This will be its own mini-spec, in its own folder, and other projects like GEE and OpenEO will be able to reuse it. The bulk of this work is being done right now in #164 

This will likely replace the 'root catalog' concept in the static catalog, which was the idea that an item would link to the catalog that had meaningful information about it, instead of just the catalog organization that's immediately above it. 

This will likely also obviate the need for the collection extension, so #174 may just result in removing it, if dataset covers all the needs of it. 

Related issues + PR's
#81 #164 #35 #152 #137

hgs-msmith:
I've attached a UML diagram as an attempt to summarize the different metadata model approaches we have discussed. We can use this during today's (9/24/18) conference call.

[STAC-model.pdf](https://github.com/radiantearth/stac-spec/files/2411240/STAC-model.pdf)

hgs-msmith:
Here is an updated model - my attempt to incorporate the discussion during the 9/24/18 conference call.
[STAC-model-060.pdf](https://github.com/radiantearth/stac-spec/files/2438404/STAC-model-060.pdf)

m-mohr:
Why is contact, provider etc still in the catalog? That's not the current state AFAIK. The catalog should just hold name, description and links (and properties?!). All tge additional fields in catalog belongs to the dataset.

hgs-msmith:
Updated version, reflecting changes as of 11:00am EDT 10/2/18.
[STAC-model-060.pdf](https://github.com/radiantearth/stac-spec/files/2439252/STAC-model-060.pdf)

m-mohr:
By the way, we may need/want to update the extents object later, see https://github.com/opengeospatial/WFS_FES/issues/168

hgs-msmith:
Here is an updated diagram with the changes noted by @m-mohr: 
[STAC-model-060.pdf](https://github.com/radiantearth/stac-spec/files/2451595/STAC-model-060.pdf)

and +1 on pressing the extents updates with WFS.

hgs-msmith:
[...]

Issue #43, opened 2017-12-05
Suggestions for remote sensing datasets


simonff:
Background: in the Google Earth Engine catalog, we have a number of remote sensing raster datasets updated daily.  Many of them are not observation results - they can also be, for example, global hourly climate model outputs with a pixel size of 0.1 degree, containing temperature, precipitation and other bands.

We have hit a number of pain points browsing catalogs on medium and large sites, so here are some things I wish catalogs included, with more important fields first.

1. Band type - observational, classification or bitmask. This is important for pyramiding during ingestion when one higher-level pixel is created from several low-level ones. Observational bands have continuous physical values that should be averaged during pyramiding. Classification bands (eg, forest/non-forest) usually contain integer values between 0 and N representing different land classes, and averaging them makes no sense. For the, we just pick the majority pixel value during pyramiding. Bitmask bands should just use random sampling.
2. Fill (nodata) value. It's sometimes specified in gdal metadata, but not always, and may even be wrong.
3. File modification date. Some providers regenerate files in-place several times.
4. Product version. Many providers release several versions of the same product, and it would be convenient to have a dedicated version field, rather than hunt for it in the title string.
5. Dedicated fields for band units (at least as a freeform string).
6. Band scale and offset. Even observation bands sometimes specify scale, and climate datasets use scale/offset a lot.
7. For categoric bands, it would be nice to encode the class values, names, and maybe desired colors. (Eg, value=0, class=Forest, color=00ff00; value=1, class=Water, color=0000ff)
8. Ditto for bitmask bands. They can have a complicated structure, and describing it needs a separate bug.

The rest of suggestions contain some fields we have added to our own dataset descriptions. They reflect the parameters that we think users often want to see in search results or use in queries. That is, I don't need them as much when ingesting data from other's catalogs, but I think I should provide them to improve user experience.

9. Product DOI and citation(s).
10. Per-band resolution (eg, 10 m or 0.25 degrees)
11. Instrument name. This is different from platform name - one platform (satellite) may have multiple instruments.
12. Product periodicity (eg, hourly or daily climate data) or revisit period (eg, 16 days for Landsat 8).
13. List of tags/keywords to help searching.

cholmes:
Hey @simonff - apologies for the super slow response, I thought I had written back here. But I _really_ appreciate all this input. I believe _most_ of this feedback will be relevant for an 'Earth Observation Profile' of STAC, which I hope we can get to soon. cc @matthewhanson 

[...]

Issue #220, opened 2018-08-29
Specify state how mature extensions and specs are


m-mohr:
We should add some classification to extensions and spec pages stating how mature they are.

Copied over from #202 (@cholmes):
> I think it'd be good to have a bit more guidance on how mature the 'official' collections are. I'd either do four categories - 'official', 'incubating', 'vendor' and 'proposed', and likely not put anything in 'official' yet, but explain that they are mature extensions that have been implemented in X number of catalogs / software and will aim to not change any fields (though could add).
> 
> Other route, which I'm totally fine with too, is to add a field to the table called 'maturity' or something, with a few classifications - like 'alpha', 'beta', 'stable' or something. I think we could introduce some metric, like 'number of implementations' if we want it to be more quantitative.

and

> On the 'maturity' front I think we should also make it pretty obvious near the top of each individual document how 'mature' it is. The start end date time spec is pretty immature - we just put it out and there are no implementations of it yet. It'd be good to make it clear that people are using something that is not super widely accepted yet. This can help them prioritize which extensions to actually make use of, etc.

m-mohr:
The vendor level could also be called candidate. Candidate of getting official. But that's just a quick idea. Community sounds also fine, but we could also divide them in community and official with each having the four levels of maturity.

Vendor (or community) was exactly meant as (usually) being proposed by others from external sources and I think the levels before should also be vendor/external. External extensions could also be in a branch in this repo, but can be linked from our master/dev branches in the vendor table.

But your suggestions sounds also like a good way of handling it. I'd give my +1 on both.

cholmes:
Closed with #254

PR #245, opened 2018-09-07
[WIP] Adding more fields to the bands + Added per asset EO data.

m-mohr:
+ Clarified how to list the bands available per asset.

The new fields were already discussed a little in #164:

* gain vs. scale => took scale for now, at least for me that's more intuitive and GDAL uses them, too.
* @fredliporace "gain (scale) and offsets may have distinct values depending on the date, see the case for LS4 and LS5. How should that cases be represented?" => Per asset, see below.
* @matthewhanson scale/offset/unit per asset: I added a section about additions to the Item Asset Object to document the "undocumented" feature that assets could list the available bands by name. I changed this structure from string to object to allow additional entries such as scale, offset, ... per asset. This structure may change depending on the outcomes of #142.

Related issues: #94, #43, #194, 142
Related PRs: #240


extensions/stac-eo-spec.md
```diff
@@ -82,6 +88,51 @@ numbers of several popular instruments.
...
+```
+...
+  "assets": {
+    "thumbnail": {
+      "href": "https://eo-bucket.com/example/EXAMPLE_20180713_057_122.jpg",
+      "type": "image/jpeg"
+    },
+    "B5": {
+      "href": "s3://eo-bucket/example/EXAMPLE_20180713_057_122_L2_BAND5.tif",
+      "type": "GeoTIFF",
```
fredliporace: I suggest using the recommended mime-type syntax for examples, "image/geotiff+cog" in this case, with 'format' removed.

extensions/stac-eo-spec.md
```diff
@@ -49,15 +49,20 @@ there is no valid EPSG code.
...
+| name                | string   | The name of the band (e.g., "B01", "B02", "B1", "B5", "QA"). |
+| common_name         | string   | The name commonly used to refer to the band to make it easier to search for bands across instruments. See below for a list of accepted common names. |
+| description         | string   | Description to fully explain the band. [CommonMark 0.28](http://commonmark.org/) syntax MAY be used for rich text representation. |
+| gsd                 | number   | Ground Sample distance, the nominal distance between pixel centers available, in meters. See `eo:gsd` for more information. Defaults to `eo:gsd` if not provided. |
+| accuracy            | number   | The expected error between the measured location and the true location of a pixel, in meters on the ground. |
+| center_wavelength   | number   | The center wavelength of the band, in microns.               |
+| full_width_half_max | number   | Full width at half maximum (FWHM). The width of the band, as measured at half the maximum transmission, in microns. |
+| resolution          | number   | Spatial resolution of the band, in meters.                   |
```
hgs-truthe01: How is this different then gsd?

extensions/stac-eo-spec.md
```diff
@@ -49,15 +49,20 @@ there is no valid EPSG code.
...
+| nodata              | [number] | The no data value(s).                                        |
+| unit                | string   | Unit of measurements, preferably following the singular unit names in the [UDUNITS2 database](https://ncics.org/portfolio/other-resources/udunits2/). |
```
hgs-truthe01: Is this intended to describe the unit for the pixel values? This would only be relevant for radiance right (I usually ask one of our scientists these questions)? If the band has classification or reflectance, I don't think we would put anything here?

extensions/stac-eo-spec.md
```diff
@@ -49,15 +49,20 @@ there is no valid EPSG code.
...
+| offset              | number   | Offset to convert band values to the actual measurement scale. Defaults to `0`. |
+| scale               | number   | Scale to convert band values to the actual measurement scale. Defaults to `1`. |
```
hgs-truthe01: offset and scale are pretty generic terms, and I often associate them with projection information. Perhaps we need a more complete description here? I assume this is intended for converting the digital numbers to radiance units?

extensions/stac-eo-spec.md
```diff
@@ -86,12 +91,24 @@ numbers of several popular instruments.
...
+| eo:bands   | [Asset Band Object] | Lists the bands available in an asset and any asset specific band data. |
+
+### Asset Band Object
+
+| Field Name | Type   | Description                                                  |
+| ---------- | ------ | ------------------------------------------------------------ |
+| index      | number | The band index as specified in the Band Object                |
+| unit       | string | Unit of measurements, preferably following the singular unit names in the [UDUNITS2 database](https://ncics.org/portfolio/other-resources/udunits2/). |
+| offset     | number | Offset to convert band values to the actual measurement scale. Defaults to `0`. |
+| scale      | number | Scale to convert band values to the actual measurement scale. Defaults to `1`. |
```
hgs-truthe01: Is the asset band object just a duplicate of the band object? I thought we were just specifying a lookup number here and we would define the metadata in the Band Object?

extensions/stac-eo-spec.md
```diff
@@ -154,11 +181,23 @@ Planet example:
...
+          "index: 0
+        },
+        {
+          "index: 1
+        },
+        {
+          "index: 2
+        },
+        {
+          "index: 3
```
hgs-truthe01: Looks like the closing quotations are missing.

hgs-truthe01:
I also want to allow for some discussion on what level of metadata we need to define as part of the STAC spec. In my opinion, this metadata is going beyond cataloging into a more comprehensive model.

Do we want STAC to replace the vendor supplied metadata files? Or are we just trying to add enough to facilitate search?

I'm more on the search side, but I don't think that precludes implementations from adding this metadata. I am just not sure it should live in the core stac spec?

@cholmes Is there a STAC philosophy to help decide what goes in core or a "primary" extension vs a third party extension?

cholmes:
I definitely lean towards a philosophy of 'search' being the main goal of STAC, certainly not to replace vendor supplied metadata files. Though an ideal world to me would be that vendor metadata files _are_ STAC compliant, and we let them add lots of specific things to it. 

For me 'core' is very clear - it's really the minimum that can be extended in lots of different ways. And then extensions I've felt people can 'do whatever they want'. The place I haven't thought about a lot is 'primary' extensions. I'm inclined to say that the bar is quite high for anything to be considered a 'primary extension'. Like I wouldn't consider EO extension anywhere close, until it has 10+ implementations. And it should end up as the lowest common denominator between anyone making EO data available - the common fields everyone uses - to be any sort of 'official extension'. 

I put more thoughts on #220 - would be great if you could sound in on that @hgs-truthe01 I'll try to prioritize working on that one.

Within the frame that EO is still quite young as an extension, I think more fields is ok, as long as they're optional. We'll see what people actually make use of in the wild, but it can be ok to specify ideas so we don't end up with people making very similar but slightly different things.

extensions/stac-eo-spec.md
m-mohr: Thanks for spotting this mistake on my side while merging. Wasn't meant to be in here as it's not different. Removed.

extensions/stac-eo-spec.md
m-mohr: Yes, that's the unit for the pixel values and is related to offset and scale. It may be used radiance, but derived products may also use it other measurements.

extensions/stac-eo-spec.md
m-mohr: My answer here is basically what I said regarding the unit. I'm totally open for better descriptions here.

m-mohr:
How would we separate them if those things should not get into the EO extension? Should the EO extension be renamed to "Basic EO extension" and a "Comprehensive EO extension" be added so that "eo:offset" then becomes "eeo:offset". This is probably similar to what we are discussing in #217 and others. Where to put/group things... it's really hard to distinguish and decide, but open for discussions. :)

m-mohr:
Okay, I think I'll close this PR and create this as a new "EO for processing" extension, but still there is a need to discuss how to extend extensions (I need to add fields to eo:bands) and how to extend the Band references in the assets.

PR #247, opened 2018-09-14
eo:bands array

hgs-truthe01:
After discussion, we are now proposing that eo:bands be represented as an array.

This will be particularly helpful when searching for items with band definitions that don't have well known names.

I also updated the extension example to use the new catalog.json inherited metadata model instead of the collections extension.


hgs-truthe01:
I realized the eo extension didn't discuss the band indexes in the assets. So I added a section describing that as well.

extensions/examples/L1T-catalog.json
```diff
@@ -0,0 +1,113 @@
...
+    {
+      "href": "http://landsat-pds.s3.amazonaws.com/L8/landsat8-sample.json",
+      "rel": "item"
+    },
+    {
+      "href": "http://landsat-pds.s3.amazonaws.com/L8/L1T-catalog.json",
+      "rel": "self"
+    }
+  ],
+  "properties": {
```
m-mohr: I wouldn't change to the new catalog.json inherited metadata model in this PR as this is not the topic of this PR. That should be separated.

extensions/examples/README.md
```diff
@@ -1,8 +1,8 @@
-## Landsat Example with EO and Collection extension
+## Landsat Example with EO and common metadata extension
```
m-mohr: Why is it now a "common metadata extension"? Did I miss something yesterday?

extensions/stac-eo-spec.md
```diff
@@ -21,7 +21,7 @@ It is not necessary, but recommended to use the [Collections extension](stac-col
...
 | eo:instrument    | string                   | **REQUIRED.** Name of instrument or sensor used (e.g., MODIS, ASTER, OLI, Canon F-1).                                                                                                                                                                      |
-| eo:bands         | Map<string, Band Object> | **REQUIRED.** This is a dictionary of band information where each key in the dictionary is an identifier for the band (e.g., "B01", "B02", "B1", "B5", "QA").                                                                                              |
+| eo:bands         | List<Band Object>        | **REQUIRED.** This is a list of the available bands where each item is a Band Object.                                                                                                                                                                      |
```
m-mohr: We usually list arrays as `[Band Object]` in the other spec files.

extensions/stac-eo-spec.md
```diff
@@ -82,13 +84,104 @@ numbers of several popular instruments.
 | lwir11      | 10.5 - 11.5     |           |           | 10        |            | 31    |
 | lwir12      | 11.5 - 12.5     |           |           | 11        |            | 32    |
 
+
+#### Associating assets with bands
```
m-mohr: I added that in #245 as well. I'd suggest mixing both approaches, I have the format description, you have the more human-readable description.

So, I'd suggest changing this a little to:
```
## Associating assets with bands

Asset definitions that contain band data should reference the band index. Each asset should provide a "eo:bands" property that is an array of 0 based indexes to the correct Band Objects.

### Item `Asset Object` fields
| Field Name | Type     | Description                                  |
| ---------- | -------- | -------------------------------------------- |
| eo:bands   | [number] | Lists the band names available in the asset. |

See [landsat8-merged.json](examples/landsat8-merged.json) for a full example.
...
```

What I also didn't thought of yesterday, that we wanted to add additional values to this field (for example offset and scale, see #245). So this would change the structure of this field a bit, but shouldn't be a big problem.

hgs-truthe01:
@m-mohr I got a little carried away trying to see what the new catalog format would look like. I agree that it should be done separately, so those parts have been reverted.

m-mohr:
Thank you for fixing the issues. I think there's one left (see comment). Other than that it's ready for approval.

hgs-truthe01:
@m-mohr I think I addressed them all. Was there a specific comment you are referring too?

hgs-truthe01:
@m-mohr I didn't notice that I had done that in the example snippets. It should be corrected now.

As a side note, I don't think we should be putting any extensions outside of the properties, but this is not the right place to discuss that!

m-mohr:
> @m-mohr I didn't notice that I had done that in the example snippets. It should be corrected now.

Thanks. => Approved.

> As a side note, I don't think we should be putting any extensions outside of the properties, but this is not the right place to discuss that!

I fully agree with that, but there were these concerns that objects/arrays could not be viewed correctly in legacy(?) GIS systems?!

Issue #251, opened 2018-09-21
Use valid IANA or proper unregistered media type for Item Spec type attribute


philvarner:
The "dev" version of the Item Spec has as valid entries for the `type` attribute these two entries:

> | image/geotiff                  | Georeferenced TIFF. _We would like to register image/geotiff as a mime type with IANA._ |
>| image/geotiff+cog              | Cloud Optimized GeoTIFF                                                                 |

I believe we should replace these either or both of (1) a valid IANA type and/or (2) a properly-defined unregistered type, until a point at which the IANA has actually accepted the recommended one above. 

I believe the best option is allow both, as:

>| `image/tiff` or `image/x.geotiff`     | GeoTIFF TIFF file with standardized georeferencing metadata|
>| `image/tiff` or `image/x.geotiff-cloud-optimized` | Cloud Optimized GeoTIFF   

image/tiff is the mime type that `file` returns for a GeoTIFF.  Placing `x.` in front of the type is the correct way to define an unregistered type.  `+cog` looks nice, but `+cog` is not a valid subtype like `+json` is, and even if a subtype were registered, the full name would be inverted, e.g., `image/cog+tiff`, as the less-narrow type is tiff.

The `image/tiff` spec has an `application` parameter which was originally intended to help fax machines render images, which could possibly be repurposed.  However, the valid values for this parameter are also part of a standard, so putting an arbitrary value like `geotiff` or `cog` in it would be a violation of that spec.

m-mohr:
Has been merged.

mojodna:
[puts pedant hat on]

https://serialseb.com/blog/2011/02/01/minting-new-internet-media-type-identifiers/ has some opinions on not using the `x.` prefix in favor of using a vendor prefix (would "STAC" be the vendor?)

From a consumer point of view, I want (need) to be able to distinguish between GeoTIFFs (which might be COGs?) and GeoTIFFs that are explicitly stated to be COGs. However, my pedantic side wants to use the same media type for both with some alternative way to distinguish between them (`cog+` or with a parameter like `cloud-optimized=true`).

With those in mind, my updated proposal is:

1. `image/vnd.stac.geotiff` for generic GeoTIFFs
2. `image/vnd.stac.geotiff; cloud-optimized=true` for COGs

By using a `stac` vendor prefix, we can legitimately play faster and looser without introducing conflicts with other users of similar prefixes. It's also clear that distinctions exist for STAC purposes.

[...]

PR #252, opened 2018-09-21
change recommended media type for geotiff and cog, plus some wordsmithing

philvarner:
https://github.com/radiantearth/stac-spec/issues/251
change recommended media type for geotiff and cog
wordsmith media type description
replace use of 'MIME type' with 'Media type'

### DIFF
--- a/item-spec/item-spec.md
+++ b/item-spec/item-spec.md
@@ -87,7 +87,7 @@ allowed to add additional fields such as a `title` and `type`.
 | ---------- | ------ | ----------------------------------------------------------------------------------------------------------------------------------- |
 | href       | string | **REQUIRED.** The actual link in the format of an URL. Relative and absolute links are both allowed.                                |
 | rel        | string | **REQUIRED.** Relationship between the current document and the linked document. See chapter "Relation types" for more information. |
-| type       | string | MIME-type of the referenced entity. 																								|
+| type       | string | Media type of the referenced entity. 																								|
 | title      | string | A human readable title to be used in rendered displays of the link 																	|
 
 #### Relation types
@@ -122,7 +122,7 @@ or streamed. It is allowed to add additional fields.
 | ---------- | ------ | ------------------------------------------------------------------------------------- |
 | href       | string | **REQUIRED.** Link to the asset object. Relative and absolute links are both allowed. |
 | name       | string | The display name for clients and users.                                               |
-| type       | string | MIME-type of the asset (see chapter on MIME types below).                             |
+| type       | string | Media type of the asset (see chapter on Media Types below).                             |
 
 #### Asset types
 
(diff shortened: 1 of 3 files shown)


item-spec/item-spec.md
```diff
@@ -132,31 +132,38 @@ The following types are commonly for assets and are used as key for the Asset Ob
...
 
-A set of commonly used MIME types for STAC items include the following:
+[Registered](https://www.iana.org/assignments/media-types/media-types.xhtml) Media Types are
+preferred. In cases where custom vendor-specific media types are necessary, they should
+use the `vnd.` prefix. 
 
-| MIME type                      | Description                                                                             |
+STAC Items that have sidecar metadata files associated with a data asset (e.g, `.tfw`, Landsat 8 MTL files) 
+should use media types appropriate for the the metadata file.  For example, if it is a plain text file, then `text/plain`
+would be appropriate; if it is an XML, then `text/xml` is appropriate.
```
m-mohr: Should we somehow mention that `application/xml` can be used as synonym for `text/xml` by applications? See [RFC 7303](https://www.ietf.org/rfc/rfc7303.txt) for details.

item-spec/item-spec.md
```diff
@@ -132,31 +132,38 @@ The following types are commonly for assets and are used as key for the Asset Ob
...
-| image/jp2                      | JPEG 2000                                                                               |
-| image/png                      | Visual PNGs (e.g. thumbnails)                                                           |
-| image/jpeg                     | Visual JPEGs (e.g. thumbnails, oblique)                                                 |
-| text/xml                       | XML metadata                                                                            |
-| application/json               | JSON metadata                                                                           |
-| text/plain                     | Plain text metadata                                                                     |
-| application/geo+json           | GeoJSON                                                                                 |
-| application/geopackage+sqlite3 | GeoPackage                                                                              |
+| `image/tiff` or `image/x.geotiff`     | GeoTIFF TIFF file with standardized georeferencing metadata                             |
```
m-mohr: I think I'd remove `image/tiff` here and in the next line. I think we really want to know whenever we have a GeoTiff or COG to be able to make use of their "features".

Also, I am wondering how using `x.` complies with this sentence above: `In cases where custom vendor-specific media types are necessary, they should use the `vnd.` prefix. ` For me that sounds like it should be `image/vnd.geotiff`.

m-mohr:
I added some comments to the source code.

item-spec/item-spec.md
philvarner: I would categorize that as the difference in x. and vnd. as between an open "standard"(e.g., COG, even thought there's not standards body-approved RFC) and a proprietary format from a specific vendor (e.g., ESRI Shapefiles, even though it's a mostly-open spec).  In the first case, it would use x.foo, in the second case it should use something like application/vnd.bar.   

The examples I have are Microsoft Office files- "old" .xls files use `application/vnd.ms-excel`, newer .xlsx use `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`.  Even though they're technically an open standard, they only became that after MSFT fully developed them and then got a standards body to approve them, but for all intents and purposes they're a vendor specific format.

philvarner: I feel it's important not to **force** users into using an unregistered media type, particularly one that I just made up 😬 Maybe the `type` attribute should be an array instead of a single value, so that a provider can specify as many types as they wish? That way, in the future, when `image/geotiff+tiff` and `image/cog+tiff` are IANA-registered, they can be added to the array without breaking existing applications that are expecting `image/tiff` or `image/x.geotiff`

philvarner:
@cholmes I added a reference MIME types.  I'd be glad to add it to more places if you feel the need.

item-spec/item-spec.md
philvarner: Added a reference to `application/xml`

philvarner:
@cholmes @m-mohr do you have any further feedback about the media types and documentation?

PR #256, opened 2018-10-01
wordsmith micron to standard term micrometres

philvarner:
To quote wikipedia "The micrometre (International spelling as used by the International Bureau of Weights and Measures) or micrometer (American spelling), also commonly known by the previous name micron..."


matthewhanson:
I can't seem to add my review here

Issue #195, opened 2018-08-23
Update spec to reflect more overlap between static catalogs & STAC API's


cholmes:
One of the biggest advances at STAC Sprint 3 was the idea that dynamic catalogs (STAC API) can implement the 'static catalog' spec - they don't have to be limited to just the 'stac api' part of the spec. This lets us layer on functionality in a more coherent way.

The core idea is that a 'level 1' (name tbd) implementation is just the link structure of the 'static catalog', but it can be implemented by a dynamic server, returning a catalog json (as defined in static catalog) at the /stac endpoint. It divides up its sub-catalogs however it wants, but this enables it to be used directly in STAC Browser. This is issue #152 and has a first pull request at #172 which defines an OpenAPI doc for the response. 

So a capable server can start with implementing just the /stac link structure (catalogs and sub-catalogs) for browsing. And then it can layer on the /stac/search endpoint to enable return of items based on search criteria. And then it could choose to be WFS 3 compliant, by setting up proper /collections as well as /api and /conformance endpoints. 

We should update our core documentation to reflect this. Right now a static catalog and a dynamic catalog / stac api are seen as two different implementations that share return of a STAC Item. Instead the browse link structure should be described as something that can be implemented with either an object store / online file folders, or by an API that is returning the links dynamically. And then the search API can be layered on top (and indeed a third party search could make an object store catalog searchable - it doesn't have to be the same service). This link structure should be specified in both an openapi 3 document as well as a json-schema. And a validation tool should be provided that can check a catalog and its items for conformance. Cc @jbants 

This may mean we update our 'folder structure' a bit, and perhaps even name things a bit differently. Perhaps use 'browse' instead of 'static' in some way. And 'search' instead of 'dynamic...

This issue is open for ideas on how to talk about it, and once we get the bulk of open PR's merged I'll take a crack at it.

Related #152

cholmes:
Closed with #258

Issue #152, opened 2018-08-14
Define where to put catalog level metadata in a STAC API


cholmes:
For static catalogs we have fields like 'tags' and 'contact', as well as proposals like 'homepage' in #137. It is less clear where to put this type of metadata in a stac api.

Static catalogs tend to have a 'catalog' field, that refers to its root catalog that has this field. But that root catalog is also the 'starting point' to point someone at to learn about the dataset and dive in. There should be a well known place in the STAC API that has that info. It's likely either a set of json fields in /collections/{collectionId} or a little document that hangs off it, like /collections/{collectionId}/catalog.json 

The latter likely makes more sense, especially if there's a nice way to have it work the same for static and dynamic catalogs.

Related to #81 as well as concepts around collection level metadata.

cholmes:
This will be done with #172 - the conclusion was that STAC API's should have a /stac endpoint that has a catalog structure that follows the static spec.

cholmes:
Not sure why I closed this - should be closed when #172 is fully merged and the concept is fully explained.

cholmes:
closed with #172 and #258

PR #258, opened 2018-10-02
[WIP] Structure and cleanup

cholmes:
Large PR that focuses on the overall structure and flow of the spec, aligning with all the great new changes, and doing lots of different clean up. 

In particular, it focuses on #195 #194 and #152 

It is marked Work In Progress as there is more to do, but making a PR so people can see what is coming.


m-mohr:
I think this PR should point to dev and not master?

cholmes:
Ah, yes. Changed, thanks!

catalog-spec/catalog-recommendations.md
```diff
@@ -4,8 +4,89 @@ This document represents a number of ideas for options and requirements that are
...
+    "l8:geometric_rmse_model_x": 5.96,
+    "l8:geometric_rmse_model_y": 4.654,
+    "l8:geometric_rmse_verify": 5.364,
+    "l8:image_quality_oli": 9
+
+  },
+
+```
+
+This could also extend to product level metadata:
```
m-mohr: Why is it 'product' level? Overall, the following example seems confusing and outdated... bands vs eo:bands, other eo fields etc.

catalog-spec/catalog-recommendations.md
```diff
@@ -61,7 +142,7 @@ of canonical product definitions online, that all static catalogs using a partic
 able to reference.
 
 
-## Product
+### Product
```
m-mohr: Products are outdated => Dataset

catalog-spec/catalog-recommendations.md
cholmes: I didn't get a chance to look in depth at everything that was in this document. It's in the repo now, so was just trying to make it a bit better. If there's not time to really go over it then we can just remove it. Best thing to do might be to move this particular discussion to an issue, since there's not really a great 'recommendation' there. I just wanted to make sure this bit wasn't lost, as it captured some of the previous ideas / options.

catalog-spec/catalog-recommendations.md
cholmes: Well I think product was more just supposed to capture the asset level metadata, not the dataset level stuff. I agree that it should be added at the dataset 'level', but we don't have the clear proposal for how to do that. I'm not sure if we should require binding it to the dataset level - at Planet we have a lot of assets and it may make sense for an Item's asset to refer to its 'product definition' (asset metadata) dynamically, instead of saying it needs to be enumerated at the dataset level.

Though not sure if this is the best place to discuss this - this PR is just an attempt to make various bits better...

cholmes:
Ok, this is good for review.

cholmes:
Ok, this is good for review.

Issue #179, opened 2018-08-15
Tracking Item Provenance


omad:
# Background
Many applications wish to be able to track detailed information about the source data and production algorithms which have been used to produce the data in this Item.

We need to balance this requirement between keeping it simple while still having enough details to be useful.

# Tracking Data Sources

A simple way to track source data would be to include additional links in the existing `links` section of `Items`, using new `rel` attributes, and potentially new `type` attributes. 

New `rel` names:
- `derived_from`
    This would provide a link to the exact `Item` or set of `Items` that this used as input data in it's production.

    Extras:
    - It can be useful to indicate a bit more information about what the source item is, or how it was used. One way of doing this would be adding a `type` attribute on the link. In HTML this is typically a MIME type, but we could simply use descriptive text. We could also use a different attribute name.
    - Do we want to provide hints about whether we're linking to STAC Items (Ideal!) or some other type or resource?

- `software`
    A link to a Git repository or other useful link to software used when producing this data.

    Options:
    - Do we need a way to represent a specific version of software. Could be extra attributes on the link, or simply providing a more specific `href`. 

# Example

```json
{
    "links": [
        { "rel": "self", "href": "https://www.fsa.usda.gov/my-real-home/naip/30087/m_3008718_sw_16_1_20130805.json" },
        { "rel": "root", "href": "https://www.fsa.usda.gov/programs-and-services/aerial-photography/imagery-programs/naip-imagery/catalog.json"}

        { "rel": "derived_from", "href": "https://ga.gov.au/sr/surface-prod-askj2ka/item.json",
          "type": "surface-reflectance"},
        { "rel": "derived_from", "href": "https://ga.gov.au/sr/surface-prod-KKaslkn2/item.json",
          "type": "dsm" },
        { "rel": "derived_from", "href": "https://ga.gov.au/sr/surface-prod-KKaslkn2/item.json",
          "type": "dsm" },
        { "rel": "software", "href": "https://github.com/GeoscienceAustralia/fc",
          "sha": "v0.6.2"}
    ],
}
```

# Storing detailed information about data production (Optional Extension)

In some cases even more detailed information might be desirable. This is a quick placeholder example, but needs a *lot* more work before being included as a STAC Extension.

```json
{
    "production_information": {
        "software": {
            "href": "https://github.com/GeoscienceAustralia/fc",
            "sha": "v0.6.2"
        },
        "environment": {
[...]

PR #259, opened 2018-10-02
Added the derived_from relation

cholmes:
See #179 

We likely should also create an extension that more fully spells out some of the ideas for this field, as well as additions to make the links more useful. Like explain that it could be used to link to non-STAC metadata, etc. But wanted to get this in so people could start to use it.

### DIFF
--- a/catalog-spec/catalog-spec.md
+++ b/catalog-spec/catalog-spec.md
@@ -150,6 +150,5 @@ The following types are commonly used as `rel` types in the Link Object of a Dat
 | child   | URL to a child [STAC Catalog](../catalog-spec/).                                                                                                                                                                                                                                        |
 | item    | URL to a [STAC Item](../item-spec/).                                                                                                                                                                                                                                                      |
 | license | The license URL for the catalog SHOULD be specified if the `license` field is set to `proprietary`. If there is no public license URL available, it is RECOMMENDED to supplement the STAC catalog with the license text in a separate file and link to this file.                           |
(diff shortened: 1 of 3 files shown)


cholmes:
@omad - your review on this would be great too.

m-mohr:
I'd recommend to add this also to datasets and catalogs. They could also be derived from anothet dataset / catalog.

cholmes:
@m-mohr - just added it in, but I'm not sure if I really love it. I said that you could derived a dataset / catalog from an Item as well. But then I wonder if it's a bit too flexible... Like if we limit it to just being derived from an Item then someone who follows the rel link can expect an Item. And more Items can be handled with more links. 

And then the case where an Item is derived_from a catalog: In the case of a mosaic that points to a catalog of exactly the items that make it up that makes sense to me. But I fear it'll often be used to just point at a dataset that would have lots of data that isn't actually used. So it'd become more meaningless.

So I think I actually want to start with just having derived_from in Item. I'm totally cool if people reuse it in dataset / catalog, but I don't want it to be so broadly defined.

If you do really want to use it for datasets the one thing I could go for is only including it in dataset, and defining it to only be pointing at other datasets. So basically the relation would be in context - a dataset can be derived_from a dataset, and an Item can be derived_from an Item. And I'm inclined to leave off Catalog, as I don't really know the use case.

m-mohr:
Sorry, that is a misunderstanding. Items should only be derived from items as you said! But datasets could be derived from other datasets, so that spec should also allow that. So I just wanted to get in what you explained on

cholmes:
Ok @m-mohr - made the changes, I think everything is good for review now, thanks!

m-mohr:
We need another reviewer with write access as @omad s review doesn't count towards the 2 approving requests as he has no write access to the repo.

PR #261, opened 2018-10-04
Change eo:epsg and eo:cloud_cover types

philvarner:
Two changes:
1) change type of  eo:epsg to string.  Even though the value of the epsg code is an integral number, it is not a numeric value, but rather a string that happens to be comprised of numbers.  

2) change type of eo:cloud_cover to number, since it's reasonable to have decimal percentages of cloud cover -- users shouldn't be forced into integer percentages.


m-mohr:
1. For me it doesn't really care whether the data is updated regularly or not as that won't change the schema, I hope.
2. I looked at it with native MS Access and it is definitely not a string, but a number. Also, all the other dumps are using integer. Oracle, MySQL and PgSQL all have it defined as INTEGER NOT NULL. 

So if we really change it to string then the Schema should add a regexp to ensure its only numeric, i.e. `^\d+$`. But for me it's still a number as the official source defined it as integer in their schemas.

philvarner:
Okay, i'll resubmit this PR with just the cloud cover change.

Issue #260, opened 2018-10-03
Dataset -> Collection


matthewhanson:
We've been through this before, but now with the latest suggested changes every `Item` will be required to be part of a `Dataset`. A root catalog will point to 1 or more `Datasets`.

But, in a WFS compliant STAC dynamic API, `Datasets` fall under the /collections endpoint.

I think we should just change the name of Datasets to Collections. The extension for shared metadata would be called 'Collection properties'.

And as @m-mohr pointed out on Gitter, there are existing [IANA relations](https://tools.ietf.org/html/rfc6573) named 'collection' and 'items', so we can use those as the rel types in links.

@cholmes

m-mohr:
As said on Gitter, I strongly support to rename dataset to collection as it would be inline with WFS (and openEO).
There are the IANA rel-types collection and item, too, which we could use. I think this is nice because we re-use existing things. So the rel types parent/root can point to either catalogs or collections/datasets and collection always points to the collection/dataset. That would make a merge also much easier.

m-mohr:
In addition, collection is unambiguous and dataset is ambiguous according to the following list of synonyms used by organizations from the CEOS OpenSearch Best Practice Document:

* For STAC Item: dataset (ISO 19115), dataset (ESA), granule (NASA), product (ESA, CNES), scene (JAXA)
* For STAC dataset: dataset series (ISO 19115), collection (CNES, NASA), dataset (JAXA), dataset series (ESA), product (JAXA)

cholmes:
So I think overall I like it, since if it all works right then it leads to a really nice integration with WFS3. But I have two fairly major concerns:

1) Since our Dataset is subtly different than the WFS Collection response document I worry about it being more confusing to call ours 'collection' too. Like it seems to make it a bit easier to me to explain a standard WFS has its 'collections' endpoints and STAC defines Datasets, which are some additional fields that add on to that WFS endpoint. Like I'd be very happy if our collection/dataset concept was truly the same, and then our 'collection spec' would be very short and would just explain how to use the WFS concept for STAC. But I see value in what STAC is doing, which is to extend the WFS collection with 'more' to emphasize that those fields in Dataset can be used for search *of* datasets. This is the direction I want CSW 4.0 to go - some defined additional fields on WFS Collections, and then use those same fields to define search of collections.

[...]

Issue #257, opened 2018-10-01
Update OpenAPI definition to catalog / dataset definition


cholmes:
The OpenAPI definition still has the old Catalog definition, with keywords, provider, etc. See https://github.com/radiantearth/stac-spec/blob/dev/api-spec/definitions/STAC-standalone.yaml#L328 (though I'm not sure if that's where the spec is defined or if that file is built from others). 

It should be updated to be in line with what we have in dev for catalog. Could have the return example be a dataset, to show a fuller return. 

Should also try to communicate a bit more in that OpenAPI snippet, like that /stac can be a catalog that links to other catalogs, or a dataset definition, or even alternate 'views' in to the data.

m-mohr:
Let's make this once we have #262 fully in. It's a bit of work to be done and I'd like to have that PR finalized first.

cholmes:
Sure, no problem - I just wanted to be sure the issue was in so it's sure to get done.

m-mohr:
I'm still waiting got #268 and #269 to get discussed and merged as these also require changes in the API.

m-mohr:
Should be updated with PR #271 , but needs to be reviewed carefully due to the huge amount of changes.

Issue #225, opened 2018-08-29
Single provider for datasets


simonff:
I suggest shifting to single object to reduce complexity. (In the internal EE catalog, I started with providers a list and found that this is not necessary in 99.9% of cases, and the remaining 0.1% is better handled by provenance.) I'm open to revisiting this, but I'd like to see more examples first.

m-mohr:
Having multiple providers for me are about giving proper credit to all parties that put effort into a dataset. That's more a "political" thing than always useful for a user, but is also a part of provenance, I think.

That said, I think this is part of a much broader discussion about how we organize provenance information, so #179 and #226 are closely related issues. Let's imagine we have a standard for describing applied processes (#179), provider information (this issue) and source information (#226). How would we like to have this organized? We probably want to encapsulate them some how. So maybe let's think about that and design is as we want it to be and just let the missing things out so we can add them later to core or via extension.

Regarding examples: Don't you already have examples in your GEE catalog? For example [1](https://developers.google.com/earth-engine/datasets/catalog/FIRMS), [2](https://developers.google.com/earth-engine/datasets/catalog/JRC_GSW1_0_MonthlyRecurrence) and [3](https://developers.google.com/earth-engine/datasets/catalog/LANDSAT_LC08_C01_T1_TOA) - and I just looked at a few of them.

simonff:
https://developers.google.com/earth-engine/datasets/catalog/LANDSAT_LC08_C01_T1_TOA (provider=USGS/Google) represents light processing (applying TOA) done by Google, and adding ourselves as an extra provider was the least bad option given that we don't show processing chain.

https://developers.google.com/earth-engine/datasets/catalog/JRC_GSW1_0_MonthlyRecurrence (provider EC JRC / Google) was created as a single product as collaboration between Google and JRC. On one hand, we want such products to be findable by querying for each organization individually, on the other hand, there's only a single end product with a single homepage (http://global-surface-water.appspot.com/). So adding two independent Provider objects is not right either. We could handle this by adding multiple organization names within a single Provider, to more accurately model the real situation.

[...]

Issue #184, opened 2018-08-20
Dataset hierarchy


simonff:
This is not a concrete proposal yet - just a place to record some thoughts on catalog organization. So far we have considered Datasets as fully independent objects, though existing in a hierarchy of catalog nodes.

In reality, some Datasets will be tightly grouped together. In EE, I represent this internally as parent and grandparent folders that may contain common attribute to avoid repeating them, though this is not visible to the user yet.

Example 1. A dataset with multiple versions (eg, GLIMS glacier data: http://www.glims.org/download/) has the same attributes across all versions. 

Example 2. Landsat can be considered as a single top-level dataset (forgetting about DN vs TOA vs SR for now). Zooming in, there are multiple Landsat subdatasets - from Landsat 1 to Landsat 8. Zooming in further, each of the subdatasets contain RT, T1 and T2 tiers (https://landsat.usgs.gov/what-are-landsat-collection-1-tiers). (To make this even worse, L7 is divided into SLC on and SLC off modes, and L8 may contain assets with only OLI or only TIRS bands). While all these datasets can be linked via keywords, some hierarchy would be useful here, but because its structure is very ad hoc, it's not clear how to generalize this well. 

Example 3. MODIS Terra and MODIS Aqua.

m-mohr:
I think this is mostly(?) tackled by the dataset extension being introduced as extension to catalogs. After #239 has been merged, it should be clear that it is possible to have a hierarchy of datasets. What we still need to define, how inheritance between datasets should be working exactly or whether that is not a thing in STAC and everything needs to be repeated.

m-mohr:
In the last months we concluded quite often that we don't do inheritance in STAC. Therefore, this issue is solved. Collections and Catalogs can build a hierarchy, but need to repeat data.

PR #262, opened 2018-10-05
Dataset improvements

### LINKED ISSUE #78: Better define what should go in 'provider' field
The 'provider' field is so loose as to not be so useful. We should improve that incrementally, and also figure out the long term plan.

In the short term that means picking a definition, either the organization providing or the sensor/instrument, and then having all the examples be consistent (right now we have 'landsat' and 'planet' in the examples, which aren't really the same level). We should then update the spec narrative to be clearer.

In the long term should perhaps figure out some way to enumerate the options, so we don't have 'landsat', 'landsat8', 'usgs', 'nasa', etc all representing the same data.

This came up in #75

### LINKED ISSUE #136: Consider versioning STAC Items
I think we discussed versioning at the first sprint and decided against it for some reason.

It does seem like it would be useful now for clients to understand what is changing.

Or alternatively we commit to getting pretty stable pretty soon. Though even if we are stable for a number of years at a 1.0 then it does seem useful to be able to go to 2.0. 

If we're for versioning then the question is where - does every item need it? Or do we do it at a collection / catalog level.

m-mohr:
Based on the last meeting (see recap: https://groups.google.com/forum/#!topic/stac-spec/A-7_e-eo3oU) I made this PR to implement the changes (related issues: # #184):
* Removed provider and license from Items. Fixes #78, #36 and may make #225 obsolete.
* Make datasets required for Items. Somehow fixes #136.
* Added a type to the provider object to specify whether the provider is a producer, processor or host. (Newly introduced by me, please discuss.)
* Removed the host, organization info is now in the provider and further info will be added at some point by the storage profiles (see #148).
* Removed the reference to SemVer from the version field (SemVer is pretty software related - was feedback from within openEO).
* Some minor fixes

Doesn't yet implement it to the API (needs to be solved with #257) and the decision on the name (#260) is still due.


dataset-spec/dataset-spec.md
```diff
@@ -21,10 +21,9 @@ Implementations are encouraged, however, as good effort will be made to not chan
 | title       | string            | A short descriptive one-line title for the dataset.          |
 | description | string            | **REQUIRED.** Detailed multi-line description to fully explain the entity. [CommonMark 0.28](http://commonmark.org/) syntax MAY be used for rich text representation. |
 | keywords    | [string]          | List of keywords describing the dataset.                     |
-| version     | string            | Version of the dataset. [Semantic Versioning (SemVer)](https://semver.org/) SHOULD be followed. |
+| version     | string            | Version of the dataset.                                      |
```
cholmes: Why is semver being removed? Not against removing, but don't remember discussing it.

dataset-spec/dataset-spec.md
```diff
@@ -52,28 +51,14 @@ The coordinate reference system of the values is WGS84 longitude/latitude.
...
 ### Provider Object
 
-The object provides information about a provider. A provider is any of the organizations that created or processed the content of the dataset and therefore influenced the data offered by this dataset.
+The object provides information about a provider. A provider is any of the organizations that captured or processed the content of the dataset and therefore influenced the data offered by this dataset. May also include information about the final storage provider hosting the data.
 
 | Field Name | Type   | Description                                                  |
 | ---------- | ------ | ------------------------------------------------------------ |
 | name       | string | **REQUIRED.** The name of the organization or the individual. |
+| type       | string | The type of provider. Any of `producer`, `processor` or `host`. |
 | url        | string | Homepage of the provider.                                    |
```
cholmes: Perhaps we make a recommendation here that the homepage should ideally include a way to find a contact about the dataset? To help explain a bit why we dropped 'contact' info.

item-spec/item-spec.md
```diff
@@ -110,7 +108,7 @@ link is required to be absolute.
 
 #### Datasets
 
-Items are *strongly recommended* to provide a link to a dataset definition.
+Items are *required* to provide a link to a dataset definition.
```
cholmes: Maybe it's a bad idea, but I was thinking we could bring the *requirement* to link to a dataset in at the Catalog level (or the Dataset level). Like let the item spec stand alone, and just have a strong recommendation. And then for Catalog it could add that any Item referenced from the Catalog should have a Dataset.

As I've been explaining the set of specs I've been thinking about trying to make sure they could be used standalone. And requiring a dataset link in an Item does make it less able to stand on its own. 

And I think we should put a bit more color here on why this is so strongly recommended. Explain that it provides the license and provider information, as well as additional information about the set of Items.

dataset-spec/dataset-spec.md
m-mohr: That's feedback from within openEO (see my PR description), this does not originate from our telco. SemVer is very software centric and doesn't really always suit to dataset versioning.

cholmes: Cool, sorry, read the description quickly. Sounds reasonable to me.

dataset-spec/dataset-spec.md
m-mohr: Okay, will add that next week.

dataset-spec/dataset-spec.md
```diff
@@ -52,28 +51,14 @@ The coordinate reference system of the values is WGS84 longitude/latitude.
 
 ### Provider Object
 
-The object provides information about a provider. A provider is any of the organizations that created or processed the content of the dataset and therefore influenced the data offered by this dataset.
+The object provides information about a provider. A provider is any of the organizations that captured or processed the content of the dataset and therefore influenced the data offered by this dataset. May also include information about the final storage provider hosting the data.
 
 | Field Name | Type   | Description                                                  |
 | ---------- | ------ | ------------------------------------------------------------ |
 | name       | string | **REQUIRED.** The name of the organization or the individual. |
+| type       | string | The type of provider. Any of `producer`, `processor` or `host`. |
```
cholmes: We should also probably explain a bit more on what 'producer', 'processor' and 'host' mean. Doesn't need to be in this table, could be just below or above. But just a bit of explanation to help people make a decision about which to use.

matthewhanson:
I find it a bit annoying that `Item` has an ID, but for Datasets we have `name` rather than ID. 

Can we change it so that both have a unique Identifier that are both called `id`?

item-spec/item-spec.md
m-mohr: I think that idea doesn't work out as it doesn't really care where things are written down as required - they are still required then. So if we don't want close coupling of things, we can just say that they are strongly recommended and must avoid using requirements between the specs at all. Sure, we can add that items that are linked from in a dataset are required to link back to its dataset, but that's a different thing for me than what you wrote in your recap: "This also lead to requiring a dataset link in every Item". So let's make it that items linked from a dataset requre a link back. I'll work on that next week.

dataset-spec/dataset-spec.md
m-mohr: Okay, I'll add explanation about it next week.

m-mohr:
@matthewhanson I think we discussed that earlier and the problem is that we are following different specifications here. GeoJSON required using "id" and WFS3 requires using "name". I don't like it, too, but that's how it is. We have this trouble also in other parts of the spec as "name" is used sometimes synonymously for IDs or titles. Only id and title are pretty much self-explanatory.

matthewhanson:
Ok, yeah that's too bad.  It might be too late in the game to ask WFS folks to change to ID, which is clearly a better term as it's specific. name is not. 

At any rate, for other parts of STAC where we can we should avoid `name` entirely and just use `id` and `title`.

m-mohr:
@matthewhanson Just try it and open an issue? I thought the same for some changes I proposed, but it seems as if they are getting adopted in WFS. We can only win ;-)
The other place that immediately comes to mind in STAC is the asset name, see my (last) comment in #139.

m-mohr:
Will be changed to id, see the issue :)

dataset-spec/dataset-spec.md
m-mohr: Added the explanation. Please check whether it needs more explanation or not.

item-spec/item-spec.md
m-mohr: Changed, please review it again.

m-mohr:
All requested changes should be incroporated, please re-check whether everything is as expected.

cholmes:
I just commented on the WFS issue - I'm not fully convinced Clements fully understood that we wanted to change the 'name' in the response document, not just in the url template names. I remember the discussion about the template names, which I think @hgs-msmith brought up, but don't remember them discussing the response context.

But I suppose we can just go ahead with the change and advocate on that issue, as there's pretty good arguments for it and WFS 3.0 isn't locked in yet. Let's just keep tracking the issue and sounding in, and I'm ok for us to go ahead with 'id' (it just may be that WFS+STAC implementations might have to put in both name and id until it works out).

cholmes:
Oh, also note that no iterations of this PR actually fixed #136 - the 'version' in dataset is not the version of the STAC specification they are following, it's the version of the dataset.

We'd need to add in the `stac_version` parameter in to dataset, or catalog.

m-mohr:
WFS id vs name: I actually thought changing the URL template would imply also changing the response, otherwise it would be even more inconsistent than before. I think it is fine to follow the route @cholmes suggested.

> I can't seem to find the discussion about the required dataset

Its in the chapter about the relation types. It's not a big discussion, but more a note that using the `item` rel type makes it required for the linked item to refer back to its dataset. Do you see it now?

> it's a bit less than what we talked about on the call

But you still think it is good as it is now or should we still change/improve something?

Regarding stac_version (#136): I was not sure whether I was excpected to add that or not, but I can make a separate PR for it.

cholmes:
> I can't seem to find the discussion about the required dataset
> 
> Its in the chapter about the relation types. It's not a big discussion, but more a note that using the item rel type makes it required for the linked item to refer back to its dataset. Do you see it now?

Oh I just meant I couldn't find our github discussion to respond in the thread. I found the edits in the spec - they look great. And yes, I think it's good as it is right now.

> Regarding stac_version (#136): I was not sure whether I was excpected to add that or not, but I can make a separate PR for it.

No, I wasn't expecting you to add it. But a separate PR on it would be great.

m-mohr:
> Oh I just meant I couldn't find our github discussion to respond in the thread. I found the edits in the spec - they look great. And yes, I think it's good as it is right now.

@cholmes Were you looking for the discussions in #174 and #136 ?

PR #268, opened 2018-10-08
#136:  Added the STAC version to catalog and dataset

m-mohr:
Added the STAC version as field to catalog and dataset (implements #136).


m-mohr:
@hgs-msmith I agree that there should be just a single STAC version across a catalog, but I would rather add that requirement to the documentation of the field. We don't know whether the top level element is always a STAC Catalog, it could also be a STAC Dataset. And anyway, Dataset inherits the fields from Catalog - that's another issue. We can't use `version` in Dataset as `version` is already used for the Dataset version, so we always used `stac_version` in discussions as far as I can remember.

m-mohr:
I added a paragraph to catalog and dataset spec that disallows mixing STAC versions. Nevertheless, I think all catalogs and datasets should hold the stac_version so that you can easily split a sub catalog and still have the STAC version available.

Issue #243, opened 2018-09-06
Bitmask/QA bands


simonff:
Pixels of integer raster bands sometimes contain bit-packed values - eg, MODIS MOD09GA: https://developers.google.com/earth-engine/datasets/catalog/MODIS_006_MOD09GA

Such bands are sometimes called QA bands, as they often contain per-pixel QA information.

Bitmask usually take up up to 4 bits, and some bit values can be unused.

Examples from the MOD09GA description:

Example A, single bit 4:
Bit 4: Digital elevation model quality flag (0 = Valid, 1 = Missing/inferior)

Example B, two bits 0 and 1:
00 (0): corrected product produced at ideal quality all bands
01 (1): corrected product produced at less than ideal quality some or all bands
10 (2): corrected product not produced due to cloud effects all bands
11 (3): corrected product not produced due to other reasons some or all bands may be fill value

In Earth Engine, we represent them with the following protocol buffer messages:

message BitmaskValue {
  int32 value = 1;
  string description = 2;
}

message BitmaskPart {
  string description = 1; // Meaning of the bit pattern.
  int32 first_bit = 2; // 0-based index of the lowest bit
  int32 bit_count = 3; // Number of bits in the pattern
  repeated BitmaskValue values = 4; // Explanation for possible bit values.
}

message Bitmask {
  repeated BitmaskPart bitmask_parts = 1;
}

The MOD09GA examples above are implemented like this (syntax is YAML):

Example A

```
 bitmask:
      bitmask_parts:
...
   - description: Digital elevation model quality flag
        first_bit: 4
        bit_count: 1
        values:
        - value: 0
          description: Valid
        - value: 1
          description: Missing/inferior
```

Example B:

```
 bitmask:
      bitmask_parts:
      - description: MODLAND QA bits
        first_bit: 0
        bit_count: 2
        values:
        - value: 0
          description: Corrected product produced at ideal quality
            - all bands
        - value: 1
          description: Corrected product produced at less than ideal
            quality - some or all bands
        - value: 2
          description: Corrected product not produced due to cloud
            effects - all bands
        - value: 3
          description: Corrected product not produced for other reasons
            - some or all bands, may be fill value (11) 
```

cholmes:
I think this is just an asset in our representation, which can be included. Could be good to write up more about it, but don't think it is a core stac spec thing.

Issue #242, opened 2018-09-06
Categorical (landcover) bands


simonff:
For categorical (landcover) raster bands, each pixel value is usually associated by the provider with a color (RGB color string or a common name) and a human-readable string explainig the meaning of the value.

Example:
0, black, ocean
1, blue, freshwater lake
2, 0000dd, river
...
255, aabbcc, urban

This can be represented by list of object named, say CategoricalValue. Each object has an integer field for the pixel value and two string fields. The color field can be optional.

m-mohr:
This should be covered by the new classification extension: https://github.com/stac-extensions/classification

PR #270, opened 2018-10-09
Allow asset specific band metadata

m-mohr:
As discussed yesterday, it is useful to allow asset specific band metadata, for example for the new extension "EO for processing and ingestion" (eop), which will add unit, scale and offset as asset specific metadata. This is currently not possible, but I think it is helpful to allow this generally in the EO extension.

This PR is meant to discuss the need and potential solutions, with me preferring this solution. Bases on PR #245 and may potentially be used for issues #243, #242 and the gain/scale, offset, unit addtions for eop, which have no separate issue.

**PR Checklist:**


hgs-truthe01:
I'm still confused on why we want to put metadata in the assets section? Any metadata outside of properties breaks the common metadata model. If you put the unit scale and offset values in properties["eo:bands"] what won't you be able to do? I'm pretty uncomfortable with having some metadata in properties and some in assets.

m-mohr:
@hgs-truthe01 After thinking about this again for a while, this actually might not be required any longer due to that fact that we changed from dicts to arrays.
But let me first ask you how you would solve this at the moment:
You have a STAC item with two assets, both have bands that share the same names and basically the same characteristics, but differ in two values such as scale and offset. How would you model that? That's basically the use case that came up previously in a discussion with @matthewhanson and @fredliporace.

matthewhanson:
@hgs-truthe01 
The problem as @m-mohr mentions is that you can have values that vary by asset and item.

For example, with Landsat there is a scale and offset given for each band, see here:
https://landsat-pds.s3.amazonaws.com/c1/L8/086/240/LC08_L1GT_086240_20180827_20180827_01_RT/index.html
Scroll down to RADIOMETRIC_RESCALING and you'll see RADIANCE_MULT_BAND_X and RADIANCE_ADD_BAND_X, along with values for converting to REFLECTANCE as well.

With Landsat-8 it's true that these values are constant across all assets, but that was not always the case. Landsat-7 and earlier have values for scaling that change for each asset. I've also seen this occur in other data sources as well, so it's not limited to Landsat.

hgs-truthe01:
Another question I think needs to be resolved. Is this a valid item:

```
{
  "assets" :{
    "B1": {
      "href": "http://landsat-pds.s3.amazonaws.com/L8/153/025/LC81530252014153LGN00/LC81530252014153LGN00_B1.TIF",
      "type": "GeoTIFF",
      "required": true,
      "eo:bands": [
        {
          "index": 0,
          "gsd": 15
        }
      ]
    }
  },

  "eo:bands": [
    {
      
      "name": "1",
      "common_name": "coastal",
      "gsd": 30.0,
      "accuracy": null,
      "wavelength": 0.44,
      "full_width_half_max": 0.02
    }
  ]
]
}
```

Why would we put specific eo fields in one place or the other? What happens if they are in both places?

hgs-truthe01:
So, the fact that gsd isn't allowed in assets is exactly the "special" rules that I was trying to avoid with the common metadata model.

I think what you are asking for here is essentially a deep merge on eo:bands between item and asset levels. But with rules on what fields go where.

What I would propose is that we don't have special rules and we make use of the commons extension. This might require a small change to commons extension to say that we should use a deep merge when applying common properties though.

collection level:
```
{
  "properties": {
    "eo:bands": [
        {
          "center_wavelength": 700,
          "gsd": 15,
          ...
        }
      ]
}
```

item level:
```
{
  "properties": {
    "c:id": 
    "eo:bands": [
        {
          "offset":10,
        }
      ]
}
```

I think this would solve the use case you are describing, but if there really is asset specific metadata, then I would suggest we need to add assets[0].properties and follow the same rules as the commons from collections to items to assets.

matthewhanson:
That doesn't quite do it either @hgs-truthe01  because  you could have an two different assets that reference the same spectral band, but with different scales and offsets.

Case in point again is Landsat, which has one set of scale/offset to get the into Reflectance, and and another scale/offset to get the data in Radiance. These could be two different assets but both reference the same spectral band.

In a similar fashion you might have the actual data file which contains RGB and requires scale/offset to get into TOA reflectance, and then a true color image which also contains RGB but doesn't have scale of offset because the values have already been scaled to a byte.

I definitely would like to use behavior like in the Commons extension do be able to handle this. Because sometimes you might have to specify these fields per asset, and other times they will be the same for the same asset across all Items. But what we're talking about really isn't band specific information though, it really needs to be tied to the assets.

I'd like to wait on this until the next release when we work on an asset definition, I don't feel comfortable pushing this into this release because I think there's still plenty to discuss.

m-mohr:
> What I would propose is that we don't have special rules and we make use of the commons extension. This might require a small change to commons extension to say that we should use a deep merge when applying common properties though.
> [...]
> I think this would solve the use case you are describing, but if there really is asset specific metadata, then I would suggest we need to add assets[0].properties and follow the same rules as the commons from collections to items to assets.

This sounds reasonable, but also a bit complex. In the end this means that for example you can have a datetime per asset for example and having multiple of them would lead to a daterange in the Item etc.

> I'd like to wait on this until the next release when we work on an asset definition, I don't feel comfortable pushing this into this release because I think there's still plenty to discuss.

Sounds good to skip this for 0.6.0 - leads to us changing the bands and assets in multiple versions drastically, but it's a 0.x version, right? ;-)

hgs-truthe01:
I'm also on board for a 0.7 discussion on this.

>  This really isn't band specific information
That says to me we shouldn't put it in an eo:bands key right?

cholmes:
Moved to 0.8.0, as asset definition stuff is going there.

matthewhanson:
Talked about this in meeting, and discussed how there hasn't seemed to be anyone with a compelling use case that is driving this anymore. We are closing this for now, and can revisit the issue if there are future use cases that come up.

PR #271, opened 2018-10-09
#257: Update OpenAPI definition to catalog / dataset definition

m-mohr:
Fixing #257. Please review it carefully, not sure whether I missed something.

I'll also address #139 here and implement a title for links. This also makes links more consistent with assets.

It adopts to the smaller catalog spec and adds the collection/dataset fields to the WFS-based /collections/... endpoints.


m-mohr:
@cholmes I addressed both issues, but of course descriptions and explanations could still be improved further in a seperate PR.

We now have two reviews and would be ready to merge, but I'd still like @hgs-truthe01 to review this PR before as he was the person behind most of the work for the API spec. Do you have some spare time for this, Tim?

m-mohr:
Collection is not an extension either, should I move it to fragments, @hgs-truthe01?

m-mohr:
Okay, thanks.
 I agree fragment organization doesn't seem very intuitive. I may just vote to put everything in one directory or so ;)

PR #275, opened 2018-10-09
replace collections extension with commons extension

matthewhanson:
Changes the Collections extension with the Commons extension. 

This allows putting properties that belong to an `Item` into it's `Collection` and have it apply to all member `Item`s.


m-mohr:
Not reviewed yet, but I'd like to have all extensions in folders. Therefore I'D suggest moving the file in a commons folder. Same. Should be done with the EO extension, but that can be a separate PR of course.

extensions/commons/README.md
```diff
@@ -0,0 +1,130 @@
+# Commons Extension Spec (`c`)
```
m-mohr: Why is there the prefix 'c' mentioned? The extension page says there's no prefix.

matthewhanson:
@m-mohr I did move it into a 'commons' folder.

extensions/commons/README.md
matthewhanson: ah, thanks, fixing...

m-mohr:
Thanks for moving it! I resolved the merge conflicts with another PR and added minor improvements.

matthewhanson:
@m-mohr examples added and README updated

PR #281, opened 2018-10-11
add swir12 band common_name

philvarner:
Add swir12 common_name for band 1.23 - 1.25.  MODIS uses this range exactly for band 5, but it might be expanded if there are other instruments that use a similar range.


matthewhanson:
@philvarner How common is that band though? The 1.6 and 2.2 are the main atmospheric windows and most sensors with SWIR bands will be in one of those two windows.

There are lots of other bands that wouldn't have a common name, DG's Worldview 3 has 8 SWIR bands, with multiple narrower bands that fit into each window, so the common_name really wouldn't apply in that case.

I'm all for expanding common_names, but when you start dealing with hyperspectral imagery and more unusual bands I don't see them being used much.

m-mohr:
I think common names should have at least three (or more?) products to be "common". So if this is only used in MODIS it shouldn't be a common name. Are there more examples than MODIS?

philvarner:
I think it depends on whether "common" is referring to it being common across many instruments, or the commonly used name for that spectra.  I think it's useful in either case.  But, if you don't think it will be useful, I'm not going to push very hard for it.

jbants:
I could see common names being used for radar bands and the different polarizations. The IEEE has set some fairly common designations for specific wavelengths. For example, L, C, and X band.

matthewhanson:
Waiting on adding additional common_names until there are more examples. Closing, feel free to re-open @philvarner with additional info.

More than happy to provide a larger set of band names as it will allow algorithms to be used generally across multiple datasets.

PR #290, opened 2018-10-15
Relation type consistency across specs

m-mohr:
1. The child rel type for catalogs should also allow to link to collections, otherwise collections in catalogs would not really work.
2. There are also other discrepancies for the root/parent/child rel types in the catalog and collection specs. Tried to fix them.
3. Removed requirement to have an absolute URL in as root in the catalog. We don't follow this in our examples anyway and it is not required in collections and items. Is that okay or should we aim for something different here?

Please review carefully. Thanks!

### DIFF
--- a/catalog-spec/catalog-spec.md
+++ b/catalog-spec/catalog-spec.md
@@ -214,12 +214,12 @@ The following types are commonly used as `rel` types in the Link Object of a STA
 | Type    | Description |
 | ------- | ----------- |
 | self    | **REQUIRED.** _Absolute_ URL to the catalog file itself. This is required, to represent the location that the file can be found online. This is particularly useful when in a download package that includes metadata, so that the downstream user can know where the data has come from. |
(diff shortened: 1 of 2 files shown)


hgs-truthe01:
I don't remember the whole conversation, but I think removing the absolute URL requirement might be an issue? I don't personally have a strong opinion, but I think that might have been the compromise that we came to in the Menlo park sprint when discussing relative urls. I think the idea was that the root was absolute so to anchor everything? Feel free to correct me if that is wrong, but it might be a good idea to do that in a separate PR.

m-mohr:
@hgs-truthe01  As far as I remember from the Menlo Park sprint, we decided to require absolute `self` URLs to resolve the relative URLs in the other links/anchor everything. I think that makes sense and in items the `root` was not required as absolute URL, too. So I think this was just an outdated catalog specification. Please correct me if I am missing something here.

hgs-truthe01:
@m-mohr That sounds right to me. Mostly I just hesitate to have a change ride on a bug fix last minute for the release. I don't feel a need to hold up the PR though.

Issue #302, opened 2018-10-17
change provider 'type' to 'role'


cholmes:
Discussion on gitter seems to agree that 'role' is a bit more descriptive for the provider array.

matthewhanson:
+1

m-mohr:
Another question that was raised by @mojodna : How to specify the licensor. Therefore I propose to add a role licensor and make the type/role/roles an array. 

Example:
```
  "providers": [
    {
      "name": "European Union/ESA/Copernicus",
      "roles": [
        "producer",
        "licensor"
      ],
      "url": "https://sentinel.esa.int/web/sentinel/user-guides/sentinel-2-msi"
    }
  ],
```

Made a PR: #303.

PR #303, opened 2018-10-18
#302: Provider type to roles, string to array, added licensor, added descripion

m-mohr:
As discussed on Gitter and GitHub (#302).

We mostly agreed on type => role and the other changes are a proposal by me based on feedback from @mojodna that asks to allow specifying the licensor. And also, string was already flawed before, because a provider could be everything from the list of options.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
--
+- Provider Object in collections: added a `description` field., renamed `type` to `roles`, change `roles` from string to array and added a `licensor` role.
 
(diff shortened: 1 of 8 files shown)


mojodna:
This makes total sense to me. I was originally thinking that keeping `licensor` closer to `license` would be nice, but both `license` and `providers` are required Collection properties.

m-mohr:
The license url for license=proprietary is in links, so things get split anyway. Also in other cases, so I don't think that's a big problem.

cholmes:
Discussed on call on 10/23 - @m-mohr to add 'description' here, for people to add in more information.

m-mohr:
Added, please re-review it ;-) @cholmes @mojodna @matthewhanson

matthewhanson:
re-reviewed, I think it's good to go

PR #311, opened 2018-10-18
[WiP] Markdownlint + light editing

mojodna:
Markdownlint yells at me when I open these, so I'm simultaneously (re-)reading everything and cleaning up the formatting (and phrasing in some places) as I go.


api-spec/definitions/STAC-collections.fragment.yaml
```diff
@@ -28,7 +28,7 @@ components:
         version:
           title: Collection Version
           type: string
-          example: 1
```
m-mohr: I was told by a dataset creator that the Major.Minor.Patch is very software-centered and datssets are usually not versions like that. So the single 1 instead 1.0.0 was on purpose.

mojodna: Ah, makes sense.

Are they versioned by date? I was thinking of using an example that's obviously a string (even if it's declared so immediately above).

m-mohr: As far as I know it depends on the provider, but an incrementing ID or date seems more reasonable than the Major.Minor.Patch.

m-mohr:
I think that should go into the 0.6.0 RC2 release!

matthewhanson:
@mojodna is this no longer a WiP and ready to merge then?

mojodna:
I’m still working on it, but it represents incremental improvement, so it could be merged now with a follow up for the rest.

PR #312, opened 2018-10-19
Improvements to the provider URL.

m-mohr:
Slightly improved the description of provider[...].url as it seems more reasonable to link to a dataset page rather than the front page of a provider. Was suggested by @simonff in a private conversation.


simonff:
+1

PR #315, opened 2018-10-23
Use STAC vendor prefixed GeoTIFF media type

mojodna:
Using the STAC vendor prefix allows us to define "GeoTIFF" in a way that's intended for use with STAC and doesn't force our interpretation on other users of a similar media type.

See #251 for additional context.

This was not specified in 0.5.x and differs from 0.6.0rc1, but seems like it belongs in 0.7.0.

### DIFF
--- a/item-spec/item-spec.md
+++ b/item-spec/item-spec.md
@@ -155,8 +155,8 @@ Common STAC Item Media Types:
 
 | Media Type                      | Description                                                                              |
 | ------------------------------ | ----------------------------------------------------------------------------------------- |
-| `image/tiff` or `image/x.geotiff`     | GeoTIFF TIFF file with standardized georeferencing metadata                        |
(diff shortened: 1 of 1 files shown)


m-mohr:
More a question than a review, but why are we using vnd.stac.geotiff and not just vnd.geotiff? I could re-use vnd.geotiff in openEO for example, but vnd.stac.geotiff seems to be STAC specific and feels a bit odd as we are not standardizing GeoTiff in any form. (Same applies to COGs).

mojodna:
I was thinking of it being STAC-specific as a feature for the time-being.

I'm hesitant to be more general because standardizing GeoTIFF has been a seemingly fraught process that still hasn't borne fruit.

mojodna:
@hgs-truthe01 good catch, thanks! search/replace fail.

Issue #309, opened 2018-10-18
Add Asset information about the use of the assets.


mojodna:
> Relationship between the current document and the linked asset. See chapter "Relation types" for more information.

`rel` is currently implicit in the object key (e.g. `thumbnail`), but should also be included in individual Asset definitions, especially since they may be duplicated, e.g. multiple "analytic" sources.

This would help with #308.

If this is a good idea, I can put together a PR for it.

m-mohr:
I think this is something we could add better in 0.7.0 and decide together with the discussion on asset dict vs. array and the proposed asset schemas in collections.

I think between release candidates we should try to not add too many new 'features'.

cholmes:
Yeah, this does seem like a good idea to me, but similar to what @m-mohr mentions, it leaves me wondering again about array vs dict. Part of the appeal of the dict to me is that the key would be the relationship. So if we need another field that will be the same for most implementations then that's less appealing. I was imagining we would just discuss asset definition in 0.7.0, but I suppose it does make sense to revisit array vs dict for asset in that context. But I do think that will truly be the last time we can change this one.

I am switching this to 0.7.0, though I'm open to arguments it should be 0.6.0-RC2. But seems like we should think it through more thoroughly.

matthewhanson:
I do not think it should be in 0.6.0-rc2. It's part of the larger discussion of asset definitions, which also includes the whole dict VS array chat.

I personally would like to see more examples beyond thumbnail for which a rel type would be used. Because otherwise I see the asset key as serving this purpose. 

I don't think multiple thumbnails is a great argument for it, because we still want to know which specific thumbnail to serve up automatically. Additional non-standard thumbnails can be specified as such via the key (e.g., thumbnail-1, or thumbnail-swir) as well as the title field.

matthewhanson:
I don't think 'rel' type is what we after here.  I can see the need for describing something as a visual asset, or a thumbnail, or an analytic asset, or metadata, but that seems different that what 'rel' is after, and I wouldn't want people to go and use all of the IANA rel types here.

I perhaps see a related field, like "type" which would have an enumerated list:

- thumbnail
- visual
- preview
- analytic
- qa
- metadata

m-mohr:
Agreed. Throughout the spec we usually use type for media types, so I would use another keyword. Maybe something like use or application makes more sense. Native speakers, please propose better words if there are any. ;)

matthewhanson:
[...]

Issue #308, opened 2018-10-18
Determining "primary" asset


mojodna:
When consuming a STAC `Item` without detailed knowledge of vendor-specific properties or filename conventions, I'd like to know which asset should be treated as the "primary" one.

Specifically, when pointing STAC Browser at a catalog, I'd like to make an informed choice about which asset to provide an interactive preview for. Since `assets` is an object, there's no "first" item to use as a default.

We have the `thumbnail` key (which should also be `rel=thumbnail`?). Should we also introduce a `primary` key (and `rel`) for this purpose?

(Additional keys / `rel`s are already relegated to extension / vendor-specific use, which should handle use-cases like "analytic")

mojodna:
In the case of datasets with bands split into individual files (Landsat 8), multiple assets may be "primary".

m-mohr:
What I don't really get: Is primary just a special thumbnail or could it also be used for other purposes? I'm wondering whether we could come up with a more intelligent way of determining the "primary" asset. We would need it anyway for cases without a primary in the assets - or should primary be required? Not sure whether there is always a primary asset though. I am also wondering how we would name the keys for multiple thumbnails. `thumbnail`, `thumbnail2`, etc? I agree that asset rel seems very useful (#309) as the asset keys are somewhat restricted. I think we should wait for the asset discussions in 0.7.0 (dict vs. list, asset schemas, etc). We may find solutions there (e.g. using arrays).

mojodna:
"primary" would be the primary dataset(s) (not thumbnail) associated with an `Item`--all other items would be derivatives / sidecars, e.g. QA band, etc. Maybe "visual" is actually what I'm looking for here.

"analytic" would be another `rel`, at least from the Planet perspective.

cholmes:
I definitely see a need for this. I see it as a 'hint' that a data provider gives to help clients determine what to 'display'. It is more than just a thumbnail, ideally it's the full Cloud Optimized GeoTIFF that can be fully browsed. Indeed having STAC refer to COG's to me is one of the most compelling features, as data can be used right there. But you may have an Unusable Data Mask and a 4-band analytic product both as COG's, and you need a way to hint which should be used in a COG slippy map. So I wouldn't see primary as being required, indeed Items with just one asset wouldn't really need it.

I've also wondered about making the dictionary names useful - like defining some common ones. 

But I agree we need to have a fuller discussion for 0.7.0 on asset definitions in general, and keep this bit in mind as something we need to do.

[...]

PR #316, opened 2018-10-23
Document asset rel

mojodna:
`rel` establishes the relationship between the current document and the linked asset.

Common relationships include: "thumbnail" (usually singular), "visual" (1+), "analytic" (1+), "qa" (1+).

This currently complements the asset key, particularly in scenarios where it would need to be parsed to determine its relationship (e.g. `visual-1`, `visual-SWIR`, etc.)

Refs #308 
Refs #309 

**PR Checklist:**

### DIFF
--- a/item-spec/item-spec.md
+++ b/item-spec/item-spec.md
@@ -127,6 +127,7 @@ or streamed. It is allowed to add additional fields.
 | Field Name | Type   | Description                                                                           |
 | ---------- | ------ | ------------------------------------------------------------------------------------- |
 | href       | string | **REQUIRED.** Link to the asset object. Relative and absolute links are both allowed. |
+| rel        | string | Relationship between the current document and the linked asset. See chapter ["Asset relation types"](#asset-relation-types) for more information. |
 | title      | string | The displayed title for clients and users.                                            |
 | type       | string | Media type of the asset (see chapter on Media Types below).                           |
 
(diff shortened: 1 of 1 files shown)


matthewhanson:
I don't think this is sufficient to do what we need, because as you pointed out in #308, a visual product could be spread out over multiple files and this wouldn't be able to depict that because there's no order specified anywhere.

I think these different use cases are better off defined in the Asset definition.

hgs-truthe01:
I'm not sure if there is enough of a difference between type and rel for me. Both of them use thumbnail as a primary example.

m-mohr:
`type` must be a media type according to the spec. `thumbnail` as type would be invalid and should be replaced with `image/...`. Are you referring to another type, @hgs-truthe01 ?

hgs-truthe01:
@m-mohr 

The asset spec says: Media type of the asset (see chapter on Media Types below).
The chapter below is a table with 1 row where type is "thumbnail".

I know we discussed mime types, but it doesn't actually say that and the one type that is listed shows thumbnail. So I think we need to de-conflict the definitions.

m-mohr:
@hgs-truthe01 You are looking at the Asset Type chapter, but further down there's a Media Type chapter. The Asset Types are the property keys of the assets and the Media Types are for the field `type`. So I think you looked at the wrong table/chapter.

Nevertheless, I agree with you that a `rel` somewhat duplicates the property key for the asset.

hgs-truthe01:
@m-mohr I see what you are saying, but that is actually even more confusing then. Asset Type needs to be renamed to something like Asset Name. We have 3 tables in a row that are all "types". If it said "Asset Key Name", "Asset Relation", and "Media Type" it might make more sense.

matthewhanson:
Closed in favor of #637

Issue #310, opened 2018-10-18
Item title


mojodna:
`Item`s currently have required `id`s. In some cases, they may have descriptive titles. We should specify a place for these titles to be provided.

My preference would be to add `title` as a documented property in `properties` (not required, but well-known).

Thoughts?

m-mohr:
I don't see that items really need a title, but I am looking to it from a EO perspective where the granules just don't have titles as they are basically split arbitrarily into granules. How often do Items really have a useful title?

m-mohr:
Well, I think OAM is not such a good example as explained here: https://github.com/radiantearth/stac-spec/issues/307#issuecomment-431287306 - The USGS DEMs seem to be more an ID than a title (if I looked at the same data as you did).

mojodna:
OAM is certainly an edge case, but I still think it's useful to consider how to surface its (core) metadata within STAC without resorting to extensions.

cholmes:
I think Seth is just saying that one option is to make a 'title' extension that adds in the parameter. But that then people would have to know about that extension.

Overall I like it, as a way to inform clients what field to display in say a search result. If it's not there then they can default to the 'id', since it's optional, but it can be a nice descriptive way to get across what should be displayed. And 'title' at the item level gives a nice symmetry with it used in catalog, in links, etc.

I'm feeling a bit conservative about putting a lot of things into RC2, and for most all functionality going forward I do think I prefer things to start in 'extensions', and then migrate to the core - keep the surface area really small to ensure everything there is useful. I do think 'title' should likely become a core field, and I think implementing STAC Browser to make use of it will help prove its case.

Let's discuss on tuesday, but I'm leaning towards put in an extension to RC2, even if it's just a simple one, and if people use it we can put it in core... But I also see an argument that it is a pretty small addition, optional, and aligns with the other 'title' stuff we have.

m-mohr:
Okay, that sounds reasonable.

Thinking more about this, I have a major discomfort with the common metadata model. It seems odd that (for example) id and title would be on different levels. Title is part of the common metadata model and id is not. Same applies for datetime and bbox/geometry. They both specify extents, but the temporal extent is part of the common metadata model and the spatial extent is not. That really feels(!) poorly designed, but of course has a background (GeoJSON compliance).

cc @hgs-truthe01 as you were voting strongly for the common metadata model: How do you feel about it?

mojodna:
From https://tools.ietf.org/html/rfc7946#section-3.2:

[...]

PR #317, opened 2018-10-23
Document 'title' as a commonly-used property

mojodna:
In some cases, `Item`s may have descriptive titles. This documents a key that _may_ be used for such purposes.

Refs #310


matthewhanson:
In most cases I don't see this being used, but in some cases it would be useful, so +1 from me.

m-mohr:
It's a minor change, optional, but is probably not used very often. Though, it allows to add the title from linking catalogs/collections, so overall I'm fine with adding it.

Issue #305, opened 2018-10-18
Adopt camelCase for property names


mojodna:
We've been using snake_case since the beginning. However, WFS 3 uses camelCase for property names, so we should consider aligning for stylistic purposes.

The only properties in the core spec (introduced for 0.6.0, so this could go into the next RC) that would be affected by this are:

* `stac_version` → `stacVersion` (`Catalog`, `Collection`)
* `derived_from` → `derivedFrom` (relation (`rel`) types)

This has implications for extensions, but since they're extensions, they can basically do whatever they want.

cholmes:
Funny, I hadn't even really realized that WFS 3 did camel case, since the properties in the main responses are all one word. 

I think I was the one who did snake case, as stylistically I just have never liked camelCase... Though I do really like aligning with WFS, so I'm conflicted. 

Is there any more context for this? Like why people make these choices in the first place? Looks like there's some discussion at https://www.quora.com/Which-is-better-snake-case-or-camel-case-pascal-case#

m-mohr:
> Is there any more context for this? Like why people make these choices in the first place?

I think this is really specific to the domain/programming language and probably in most cases is just the choice of the original "author" based on their subjective preference. It is the same here: Chris decided for snake_case for STAC, a colleague of mine decided for snake_case for openEO, etc. If you ask WFS3 for snake_case vs. camelCase they'll probably answer similar things.

mojodna:
I don't have an independently strong opinion between the 2. When I do, it's according to local norms of the language I'm using--Python, definitely snakes, JavaScript, camels for life, C#, Pascal is my god--and it's just weird/jarring/an indicator of the author's native language when things don't match.

In this case, WFS3 owns the local norms, so we should follow its lead to keep things internally consistent.

m-mohr:
Closed as discussed in #318.

PR #318, opened 2018-10-23
WFS 3 alignment: camelCase

mojodna:
WFS 3 uses camelCase for property names. This changes the 2 affected core property names (`stac_version` and `derived_from`) to match for consistency.

Refs #305


mojodna:
Agreed that both external and internal consistency is desirable. I skipped the extensions as they've been in use prior to 0.6.0, but with the idea that they'd be aligned longer-term.

cholmes:
Discussed on oct 23 2018 call, decided we are sticking with snake case in core. WFS3 alignment is nice, but it is seen as the api mechanism, while STAC's place is content, so we don't feel like we need to align with their style. Extensions can do what they want.

Issue #307, opened 2018-10-18
Item-level license and providers


mojodna:
For `Item`s that are difficult to group into coherent `Collection`s, we should provide a clearly-defined mechanism for specifying their `license` and `provider`s.

In other words, `license` and `providers` are important enough properties that they should be calculable for a given item, either directly or via the `Collection` that it belongs to. `Item`s that only belong to `Catalog`s are currently orphaned with no license data.

OpenAerialMap is a collection of heterogeneously-licensed data (differing CC licenses) provided by a variety of users. In order to surface licensing data for individual items, a `Catalog` would need to consist of multiple `Collection`s per user, one for each license used by that user. To further complicate things, legacy imagery does not include user associations. Complicating things even *more*, user uploads may have varying [free-form] attribution.

From Gitter:

@cholmes:

>  for OAM I'd have many collections. It can be one 'catalog' at the highest level, and then you'd navigate down to like the user level, and perhaps each user (or user + specific drone?) would be a 'collection'.

@matthewhanson:

> The license thing had come up before, when Jeff at DG mentioned license might vary in a collection.
> I think for OAM there's lots of other good reasons to make different collections for the data (varying platforms, instruments, spectral band definitions, etc), however for some collections it is useful to group together things from multiple sources. I'm currently working on making a collection that represents a specific area, with mixed imagery as member Items, so the license thing is pertinent there as well (it could be DG imagery, or publicly available imagery).
> We could solve this pretty easily by moving license into the Item properties. It could then be specified using the Commons extension in the Collection, or not and just specified per Item.

@m-mohr:

> You are taking a subset of a collection and merge it to another collection, so to me it would be clear that there should be multiple collections involved.
> If we are stripping away license from collections for that reason, we would also need to strip provider. Your combined imagery probably doesn't only have different licenses, but also different providers. And basically, the same applies for the extents. So we are basically back to discussing usefulness of collections as we would strip away all fields that separate collections from catalogs. Really? We just discussed that in depth...

mojodna:
Allowing `license` and `providers` into `Item` `properties` has the additional benefit of making `Item`s more self-contained (no need to look up `rel=parent` or `rel=collection` (if present) or `rel=root` for additional metadata).

m-mohr:
What's the question here? (you added the question label)

[...]

PR #319, opened 2018-10-23
Resurrect Item 'license' and 'providers'

mojodna:
These were previously scheduled to be dropped in 0.6.0. I propose retaining them until we have a better way to handle scenarios where `Item`s and `Collection`s are 1:1.

@cholmes:

> I'm thinking we should consider putting license and provider back in to Item for 0.6.0, since it is something we are actively taking away. I had thought that the collection would cover every use case, but having OAM provide a counter example gives me some pause.

Refs #307


cholmes:
This is the PR for #307. 

Assigning to @m-mohr to start take it over. I think the action is to close this PR and start a new one, but could potentially start from this one?

m-mohr:
superseded by #321

PR #320, opened 2018-10-23
Stac process

### LINKED ISSUE #67: Define the branching / PR / changes 'process'
As the spec stabilizes a bit we should document a process for how a new version is made and how the community reaches consensus on how a suggestion gets incorporated.  

cc @jeffnaus - you said you could help on this one.

cholmes:
Added information about development and release process, should close #67


m-mohr:
Looks fine, but I think there should be a short not in the Release Process that the version numbers need to be updated in several places (at least the top-level README, but potentially also in the API spec docs, examples etc.).

cholmes:
Ah, good point. Will add that.

PR #321, opened 2018-10-23
#307: Single Item extension - Providers and License on the item level

### LINKED ISSUE #307: Item-level license and providers

m-mohr:
That extension name ([mixin](https://en.wikipedia.org/wiki/Mixin), prefix: `mix`) is a bit... strange(?), but describes basically what we are doing here:
* using fields from collections as mixin and
* mixing providers and licenses in a collection ;-)

Additionally, it is pretty generic and allows to extend it with new fields in future if we like.

Fixes #307.

Further changes in #303 would most likely need to be adopted here.


extensions/mixin/README.md
```diff
@@ -0,0 +1,45 @@
...
+Provides a way to specify several fields in Items that usually reside on the collection-level such as license and provider.
+
+- [Example](example.json)
+- [JSON Schema](schema.json)
+
+## Item fields
+
+| Field Name       | Type              | Description                                                  |
+| ---------------- | ----------------- | ------------------------------------------------------------ |
+| mix:license      | string            | Items's license(s) as a SPDX [License identifier](https://spdx.org/licenses/) or [expression](https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60) or `proprietary` if the license is not on the SPDX license list. Proprietary licensed data SHOULD add a link to the license text, see the `license` relation type. |
```
mojodna:
```suggestion
| mix:license      | string            | Item's license(s) as a SPDX [License identifier](https://spdx.org/licenses/) or [expression](https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60) or `proprietary` if the license is not on the SPDX license list. Proprietary licensed data SHOULD add a link to the license text, see the `license` relation type. |
```

extensions/mixin/README.md
```diff
@@ -0,0 +1,45 @@
...
+The object provides information about a provider. A provider is any of the organizations that captured or processed the content of the collection and therefore influenced the data offered by this collection. May also include information about the final storage provider hosting the data.
+
+| Field Name  | Type      | Description                                                  |
+| ----------- | --------- | ------------------------------------------------------------ |
+| name        | string    | **REQUIRED.** The name of the organization or the individual. |
+| description | string    | Multi-line description to add further provider information such as processing details for processors and producers, hosting details for hosts or basic contact information. [CommonMark 0.28](http://commonmark.org/) syntax MAY be used for rich text representation. |
+| roles       | [string]  | Roles of the provider. Any of `licensor`, `producer`, `processor` or `host`. |
+| url         | string    | Homepage on which the provider describes the dataset and publishes contact information. |
+
+**type**: The type of the provider can be one of the following elements:
```
mojodna:
```suggestion
**roles**: The provider's role(s) can be one or more of the following elements:
```

extensions/mixin/example.json
```diff
@@ -0,0 +1,66 @@
...
+      ]
+    ]
+  },
+  "properties":{
+    "datetime":"2018-01-01T13:21:30Z",
+    "mix:license":"PDDL-1.0",
+    "mix:providers":[
+      {
+        "name": "John Doe",
+        "type": "producer",
```
mojodna:
```suggestion
        "roles": ["producer"],
```

extensions/mixin/README.md
m-mohr: Fixed, in both extension and collection spec.

extensions/mixin/example.json
m-mohr: Fixed.

extensions/mixin/README.md
m-mohr: Fixed, thanks.

m-mohr:
Fixed the type => roles issues - were copy&paste errors as #303 was not merged yet.

I'm always trying to make this extensible. We are not introducing the mixin concept itself, but leave the door open to add it later. For now we limit it to license and provider. That said, I personally would prefer to deprecate this extension anyway in 0.7.0 based on the planned improvements to the metadata model.

We are recommending prefixes in our spec, so I'd say we should follow our own recommendation. Nevertheless, if you feel it should be without prefixes, I think I'm fine with it.

m-mohr:
Maybe we could also go with the prefix "item" and "Single Item Extension"?! That would also make a clearer distinction between the collection provider & license and the item provider and license. Maybe that's a better solution than mixin...

PR #322, opened 2018-10-24
add note with link to IANA Link relation types

### LINKED ISSUE #314: Add link rel type for additional info or other link types
In the documentation it's not clear if arbitrary rel types are allowed, or if one must be chosen from the list available.

Either way, I suggest adding at least one new rel type:

- info: This is a link to additional information about this Catalog/Collection/Item

I'm not sure arbitrary values need to be allowed though. If you want to explain something better then you can put that in the `title`.  If we want the current rel types to be a definitive list then we need to at least add in:

- other: a link to other related material

matthewhanson:
resolves #314


matthewhanson:
@m-mohr added

hgs-msmith:
This it is a nit - but did you intend to put this note inside the Links property table?

PR #324, opened 2018-10-24
UML diagram

hgs-msmith:
Added UML diagram - also attempted to capture proposed changes discussed during 10/23/18 conference call.


m-mohr:
I think the UML diagram looks much better and cleaner now, great improvement, thanks. I know the reason behind it, but it could be a little confusing that several of the entities are not in plural forms.

More a general question, but `link:rel=derived_from` is specified as `(1..1)`. Why is that? I think it should be `(0...*)`. Data could be derived from multiple datasets and may not be derived from another dataset (=original dataset).

Also, the rel types `root` and `parent` are not required and should be `0..1`, I think. Only the `self` rel type is required. The collection rel type can't be modeled correctly due to its one way requirement, I think.

hgs-msmith:
@m-mohr, thanks for the input on the cardinalities - have updated the diagram accordingly. I used Visio to create the diagram - it is powerful and convenient, but requires a license. In the future, we may want to move to one of the free online UML tools that uses a markup language to describe the relationships. Then, we could archive the markup file here so that anyone could edit it. Does anyone have a preference on one of these free UML tools?

m-mohr:
Thanks, @hgs-msmith . The cardinality for link:rel=derived_from is now specified as (0..1). Was that on purpose? I still think it should be (0..*). Yes, this is nitpicking from my side.

I'm currently not using any UML tools myself so I can't give any advice.

fredliporace:
@hgs-msmith sorry for the late review. The only issue I found is the need to change 'provider' to 'providers' in collection. I have no current experience on open source UML tools, the last I used was [dia](http://dia-installer.de/shapes/UML/index.html.en) but this hasn't been updated for a while.

PR #329, opened 2018-10-26
Allow null for eo:epsg in JSON schema

### LINKED ISSUE #330: Allow null for eo:epsg in JSON schema
The spec explicitly allows setting eo:epsg to null, but the JSON schema would report the null value as invalid as only integer is set as data type.

m-mohr:
The spec explicitly allows setting eo:epsg to null, but the JSON schema would report the null value as invalid as only integer is set as data type. Fixing this.

Resolves #330.


m-mohr:
Sorry, branched this off the master branch, which is wrong.

Issue #199, opened 2018-08-24
Use continuous integration and define full suite of tests to validate all PR's


cholmes:
Thanks @scisco for doing the bulk of the work on this, and @justinwp for having the same idea and getting it set up with Travis (may want to use some of his checks in CircleCI). He's got CircleCI set up on master and dev. 

Right now it's not doing much, but is working. We likely don't want to do the full validation until we merge the major PR's and update the examples all in one go, instead of having each json schema PR update all the examples just for their change.

- [x] set up circleci
- [x] check the api spec creation #183 
- [x] update all the examples for the new spec.
- [x] turn on CI for json schema checks of items and catalog samples.

m-mohr:
The CI should also make sure that the generated API specs are up-to-date, see #208.

cholmes:
Yes, we need that. That's what I was trying to say in the checkbox above on 'check the api spec creation'. I thought #183 would do that, but it looks like @scisco is going to take another approach.

cholmes:
Discussed on a call today - it would be good to have a validation tool that can reference remote schemas, which isn't in place now. May be good to use the validation tool @jbants is working on. James, if you could link to the python validator from here that could help @matthewhanson get it set up.

jbants:
https://github.com/sparkgeo/stac-validator
@matthewhanson let me know if you want some help.

cholmes:
We also discussed that it would be ok to do this work after RC1, since it hopefully doesn't change the spec at all. If spec changes are needed we'd cut an RC2, but most likely we'd just get the validation working and maybe update examples and release 0.6.0

cholmes:
Changed to 0.6.0 Would be good to update the CI to use the stac validator (as I think it can pull in remote references), and ensure all the examples are up to date, and have that CI in place for all future dev development (so people who make PR's that change the schema need to update the examples too). And we should ensure the OpenAPI doc combining stuff is all working.

Also consider doing formatting. Like #298 on the examples. We could even consider it for the markdown formatting as well... Though that should probably wait till 0.7.0 so we don't have tons of changes after RC1

m-mohr:
As said in #298, it would be useful to write a script that combines the following tasks:
* Check the style of examples (excluding the examples included in the markdown files)
* Check the style of JSON schemas
* Check the validity of JSON schemas
* Check the validity of examples based on the JSON schemas
* Generate API documents from fragments (for api-spec and extensions)

Not sure about Markdown formatting though.

[...]

PR #334, opened 2018-10-29
Update ci config

jbants:
This PR updates the circleci config to use the python [stac_validator](https://github.com/sparkgeo/stac-validator).

This starts to work on #199 as it supports pulling in remote references and can be used locally to validate a STAC. The validator only validates against stac_spec release versions (github tags) right now. An issue has been created to support validating against stac_spec branches and local schemas as well. This would support different CI configurations.


m-mohr:
Thanks! I have a question: You said, it "only validates against stac_spec release versions (github tags) right now". So if we merge that to dev and want to make a PR that changes JSON schemas and examples, wouldn't this fail in the CI and thus block us from changing anything in JSON schemas in dev at all?

matthewhanson:
This looks good to me for now, since as @jbants points out it will only fail on schema changes.

PR #342, opened 2018-11-06
Updated validation readme with python validator

jbants:


m-mohr:
Great, thanks!

Maybe we should explicitly mention that the JS validation is deprecated and will be removed in the next version.

Issue #155, opened 2018-08-14
SAR Content Extension


hgs-msmith:
Define metadata extension for SAR data.

hgs-msmith:
Consider adding the following as SAR-specific metadata:

- Pass direction | ascending vs. descending orbit
- Polarization | HH, VV, HV, VH
- Instrument mode | IW, stripmap, etc.
- Product class | RAW, SLC, GDR, etc. – different levels of processing
- Range resolution | eo:gsd as array?
- Azimuth resolution | eo:gsd as array?

jeffnaus:
I'll, jnaus, will own this and give an example and kick off for comment

Questions:  What is the relationship between this and the EO extension.  can they/ should they have the same parent.

matthewhanson:
There is a lot of fields in the EO extension that apply here. The whole bands mechanism can be used as in EO extension - SAR bands can be specific with wavelength, FWHM, and even common_name can be used to refer to standard radar wavelength bands (e.g., x-band).

Regarding sharing a parent, multiple inheritance isn't something that has been discussed yet, unless you guys talked about this meeting. I'm not sure we want to go down that path.

jeffnaus:
In my experience the bands are different then EO in that they are magnitude, phase, or SLC (single look complex, which is the response time and strength combined).  The wavelength information usually is set for the sensor and this is then combined with beam mode (like wide or fine...) to determine spatial and spectral resolution.  Multi-look data is different from this also and I don't have direct experience with those datasets.  I'll do more research around this, I may make an initial spec proposal for just single look derived products.

m-mohr:
As we need it for openEO soon, we'll write a proposal in the next weeks and post it for discussions.

m-mohr:
I made a first proposal for the SAR extension, please review #347. Thanks!

m-mohr:
It has been merged to dev. There were still two comments by @mrepse ([see comment](https://github.com/radiantearth/stac-spec/pull/347#issuecomment-442357747)) and @AlenaDostalova ([see comment](https://github.com/radiantearth/stac-spec/pull/347#issuecomment-442383383)) with open points for discussion. It would be great if you could propose your changes as PR for discussion.

Other than that, I still need to add a JSON Schema.

m-mohr:
Is complete now, I think.

PR #347, opened 2018-11-12
[WIP] SAR extension (#155)

m-mohr:
This is a first draft for the SAR extension (see issue #155), which allows to describe synthetic-aperture radar data that represents a snapshot of the earth for a single date and time.

I am by no means an expert in SAR imagery so expect some problems and misunderstandings in the current proposal. I just wanted to make a "kick-off" to get started and I happily await your suggestions for improvements. For example, I was not quite sure whether several fields fit better in the Band properties or in the Item properties so I started with them as Band properties for maximum flexibility. 

Please keep in mind that STAC focuses on *searching* and based on that we should decide whether fields should be added or not. If there are fields that need to be described for processing/ingestion/... tasks, please also report them. Similar to the EO and EO for processing (EOP) extensions with fields like `offset`, `scale` and `nodata`, I plan to also add another SAR for processing (SARP) extension, which is focused on properties for processing and ingesting data.

Files to be added once the draft is reviewed and more stable:
* More examples?
* JSON Schema

Members of the following organizations contributed their knowledge to this extension:
* WWU Münster
* TU Vienna
* JRC
* Terradue
* Sinergise
* University of Washington


AlenaDostalova:
I would propose following changes:

- moving from band properties to item properties: instrument_mode (for SAR, the band is connected to polarization, different instrument modes have different resolution, image extents, ...), center_frequency, center_wavelength (these typically do not change for given instrument), resolution, pixel_spacing and looks.

instrument_mode, sar:type: there are more possibilities than listed here (for example GM and AP instrument modes in Envisat ASAR sensor)

I would add also processing_level to the item_properties (level 0 being raw type, level 1 being GRD or SLC, level 2 being OCN in case on Sentinel-1) as this is more general description for other sensors as well.

I can provide Envisat ASAR as an example, if it would fit here as well.

m-mohr:
Got some more feedback:

* "I guess you also want to consider a geocoded product, which is different from GRD, but what most people are looking for. That further complicates stuff, because you would need to describe additiional processing steps (DEM used, spacing, angle correction yes/no)."

* "Anything placeholder planned for interferometric outputs as well (phase, coherence)? Would need to refer to the "parent" SLC pair (and processing specs)."

* I removed FWHM from the bands, which was available in the EO extension as I was told it called the Pulse Repetition Frequency for SAR, but was claimed to be a low priority search field in any case.

p3dr0:
Hi Mathias 
Many thanks for this work
I propose the following fields to your consideration  

- swath : String : Swath identifier (e.g. Envisat ASAR has 7 distinct swaths (I1,I2,I3...I7) that correspond to precise incidence angles for the sensor). 
- track (or path or wrsLongitudeGrid) : string :  A string providing the coordinate of the product on the across track (east/west) axis of the reference system used. 
- frame (or row or wrsLatitudeGrid) : string : A string providing the coordinate of product along track (north/south) axis of the reference system used. 

The last two should probably not be defined here but directly on the eo extension (the same also to the proposed "processing_level" ). I'll create an issue for this.

mrepse:
I would suggest having certain fields be obligatory based on the satellite type. I'm really only familiar with Sentinel-1 so can't comment on the rest, but for this one the following would be useful:
- sar:type to be obligatory (SLC is very different from GRD)
- sar:pixel_spacing to be obligatory for GRD products (they come in various resolutions)

It could also be useful to add a third optional parameter to the sar:looks array which would be the equivalent number of looks value (but this is a float value then).

I'm also rather confused by the band designation; what is the definition of one band? To me one band equals one polarization value, so sources containing two or more polarizations would therefore contain as many band objects. Now it seems its possible to have all the polarizations in one "band"?
I'm similarly confused by the phase property. As far as I know phase should be a per-pixel value as opposed to a single number. You could even call it a separate band, if you will. I guess it comes down to how you define one band. But at least for S1 SLC the values in one file are written as complex numbers (amplitude, phase). You can separate them, but you don't have to.

scottyhq:
Thanks for getting this going @m-mohr! I'd like to echo the comments above. I'm wondering if it makes sense to have `sar:` and `insar:` as separate specifications? In many cases people are just interested in amplitude images, and `insar` processing results in different metadata and derived products. If the goal is to be general enough for all sensors, NASA Alaska Satellite Facility is a good resource for common metadata: 
https://www.asf.alaska.edu/get-data/learn-by-doing/ 

Often practitioners of InSAR want to find data from a common 'relative orbit' (also called 'path' or 'track') and 'frame' as @p3dr0 mentioned. But tracks and frames aren't standardized across archiving centers or space agencies.

'incidence angle' or 'off nadir angle' is also helpful to know. 

I attempted putting together a toy catalog based on insar processing with stac version 0.5 a while back, and will try to incorporate 0.6 changes and this extension in the near future!
https://dinosar-stac.netlify.com/item/snohomish/A137/int-20170219-20170126

m-mohr:
Thank you very much for all contributions! Please re-review this PR every once in a while to make sure further changes make sense for you so that we get to a useful result for everybody.

@p3dr0 
For track and frame I'd like to wait for the outcomes of #352 and comply with the consensus from that discussion.

> * swath : String : Swath identifier (e.g. Envisat ASAR has 7 distinct swaths (I1,I2,I3...I7) that correspond to precise incidence angles for the sensor).

I'm always feeling bad about free-form string fields where users can add whatever they want. That usually leads to problems as different data gets added and searching doesn't work as expected and the data is of no real value. But if I understand correctly the swath identifier describes the incidence angles. So should we better add a field describing the incidence angles?

@mrepse 
> * sar:type to be obligatory (SLC is very different from GRD)

Changed that.

> * sar:pixel_spacing to be obligatory for GRD products (they come in various resolutions)

Changed it to strongly recommended for GRD.

> It could also be useful to add a third optional parameter to the sar:looks array which would be the equivalent number of looks value (but this is a float value then).

Added as optional third value.

> I'm also rather confused by the band designation

I'm coming from a MSI World and I am not very familiar with SAR so my understanding is probably too much centered around bands as used for MSI. It would be highly appreciated if you could come up with a proposal that removes/replaces/improves the bands the way you think it would make more sense.

> I'm wondering if it makes sense to have `sar:` and `insar:` as separate specifications?

Could be useful, yes. As always it needs people that push things. So feel free to come up with a proposal. ;-)

> https://www.asf.alaska.edu/get-data/learn-by-doing/

That is a VERY useful resource, thanks. I'll check it out and improve the spec accordingly.

> Often practitioners of InSAR want to find data from a common 'relative orbit' (also called 'path' or 'track') and 'frame' as @p3dr0 mentioned. But tracks and frames aren't standardized across archiving centers or space agencies.

So I'll start with absolute orbit in the spec and then we need to check how we can get a useful field for the relative orbit etc. Proposals are always welcome.

> 'incidence angle' or 'off nadir angle' is also helpful to know.

Added off nadir angle, incidence angle could be added, too (see above).

> I attempted putting together a toy catalog based on insar processing with stac version 0.5 a while back, and will try to incorporate 0.6 changes and this extension in the near future!
> https://dinosar-stac.netlify.com/item/snohomish/A137/int-20170219-20170126

Awesome! Would be great to have an actual implementation of this extension to check whether the extension works as expected!

m-mohr:
I tried to incorporate as much feedback as possible. Please re-review the current version, especially the bands changed a lot. There's also a new Envisat ASAR example made by @AlenaDostalova. As soon as I get some agreement on the current proposal, I'll start adding a JSON schema. Nevertheless, feel free to give additional feedback and propose changes if you think they are required. Thanks!

cc @mrepse @AlenaDostalova @p3dr0 @scottyhq @jeffnaus @matthewhanson @hgs-msmith

mrepse:
This does look better, except I would still suggest some changes to Band Objects. Again I should point out that this is coming from a Sentinel-1 perspective:
I'm curious in what situation would the polarization be null? Will this support level 2 products? At least for S1 level 1, all data is associated with a polarization. So if this is for level 1 only then I would suggest making the polarization field obligatory.
I would also suggest adding an obligatory field which hold the essential band information: has it been detected or not? (i.e. does it contain just amplitude = detected, or also phase information = not detected). GRD is detected, SLC is not. This is somewhat of a superset of the data_type idea but I would suggest changing that as well. The data_type and units can be merged as the data_type is essentially a list of units. This would (optionally) give extra information to the obligatory field above. GRD would then have amplitude here (intensity is the square of the amplitude so its essentially the same thing and no need to have both). Sigma_0, gamma_0, beta_0 are calibrated units and ideally you need more information about them than just the units (what terrain model was used), but at least for S1, you never get these directly. SLC has the in-phase and quadrature components, from which you can derive amplitude. But neither is strictly speaking amplitude so its a bit tricky. So I would propose having the new "units" be a string array and you can add any as necessary.
Examples for dual polarization S1 IW:
GRD: VV and VH detected bands with amplitude units
SLC: VV and VH non-detected bands with in-phase and quadrature units
(Which is why I'm not entirely sure if calling them "units" would make sense, maybe data components?)
And maybe make sar:bands obligatory as well.

AlenaDostalova:
as I suggested some of the changes in band object you ask about, I hope it's OK if I respond to part of your feedback.

"I'm curious in what situation would the polarization be null? Will this support level 2 products? At least for S1 level 1, all data is associated with a polarization. So if this is for level 1 only then I would suggest making the polarization field obligatory." - yes, this should suport also level 2 data, for this reason, there is an option to have polarization 0 (for example for the bands of OCN product or of the angle information (i.e. projected local incidence angle) of geocoded data.

"Sigma_0, gamma_0, beta_0 are calibrated units and ideally you need more information about them than just the units (what terrain model was used), but at least for S1, you never get these directly" - for this reason, there is the description field, where the applied processing (i.e. used DEM, used software, method of geocoding, was thermal noiseremoval appliedor not...) should be written. I know, this is not an ideal solution, but as was mentioned in some of the previous feedback (and I also agree with that) - people will often look for geocoded data, so they should be included as well. Although it makes the description more complicated.

"Which is why I'm not entirely sure if calling them "units" would make sense, maybe data components?" I included 'units' field again for the purpose, that not only level-1 but also geocoded data can be described - then it is good to know, if its in linear or dB domain. There might, of course, be a better solution. 

Cheers,
Alena

cholmes:
What's the status on getting this merged? I'm inclined to get it into dev, and then we can gather more feedback as additional PR's on that core.

m-mohr:
I'd say let's merge it and evolve from there. There are some suggestions in the last two comments, but they are not very concrete (i.e. direct changes to the markdown file). I think it is easier to evolve parts of the spec in more concrete PRs, for example the comments from @mrepse regarding data type, other band information or the discussion regarding units.

cholmes:
Great. I'll merge.

m-mohr:
D'Oh, I forgot to add the JSON Schema, but will add with a separate PR. Will open an issue so that we don't forget.

PR #360, opened 2018-11-21
JSON Schema draft-07

mojodna:
This updates the JSON schemas to draft-07.

From http://json-schema.org/draft-07/json-schema-release-notes.html:

> For the Core and Validation specifications, draft-07 is a relatively minor update. In terms of validation keywords and outcomes, it is fully backwards-compatible with draft-06.

GeoJSON `Feature`s are directly addressable (for validators that resolve external URLs). I _think_ `stac-validator` will handle this, but can't check due to https://github.com/sparkgeo/stac-validator/issues/27.

I've got this working with [ajv](https://github.com/epoberezkin/ajv) (which is where the `id` → `$id` change was recommended) in STAC Browser (it validates behind the scenes and logs warnings to the JS console):

https://github.com/radiantearth/stac-browser/commit/5b42b6bc44bb2baee2e710bf053b1f51ad412ca9

`loadSchema` was what was needed there to resolve remote schema `$ref`s (and why it didn't work using `ajv-cli` on the command line).


mojodna:
@jbants does this work with STAClint?

mojodna:
@brianbancroft looks like this might be up your alley?

jbants:
Hey @mojodna, im out travelling in India until mid January. Brian is going to have a look. Off the top of my head I think the JSON spec library we are using supports 7. But we will test it out to confirm

m-mohr:
This change will let the JavaScript Validator fail for mostly all requests (can't resolve external references), so we should remove the instructions for it with this PR, too.

mojodna:
Good call. I'll take that on soon.

PR #362, opened 2018-11-27
Label Extension

matthewhanson:


extensions/training-data/README.md
```diff
@@ -0,0 +1,55 @@
...
+- **properties.assets**: Unlike other extensions, the TD extension requires a single asset with the key "labels". This is a GeoJSON FeatureCollection asset containing the actual label features. As with the core STAC Item a thumbnail asset is also strongly encouraged.
+
+### New Item properties
+| element         | type info       | name                       | description       | 
+|-----------------|-----------------|----------------------------|--------------------------------------------------------------------------------------------------| 
+| td:title | string | Title | A human readable title of the dataset for display |
+| td:description | string | Description | **REQUIRED.** A description of the training data, how it was created, and what it is recommended for |
+| td:classes | [string] | Classes | **REQUIRED.** a list of keywords representing the nature of the labels. (e.g., tree, building, car, hippo)
```
lossyrob: Have you thought about supporting non-classificaiton tasks? e.g. "estimate count of  buildings in this 500x500 pixel grid". Requiring classes here wouldn't fit with regression problems. An easy work around would be to just give a single "value" class, but you may want to bake a flexibility in the data model instead.

daveluo:
Hi, 

To add more ML use cases and sample data for consideration in this STAC extension proposal, here's a small sample set of ML training inputs (source imagery from UAV & building footprint labels) and outputs (raster and polygons of segmented buildings) for building segmentation in Zanzibar. 

catalog: https://github.com/daveluo/stac4ml-demo
stac browser (temp url): https://zen-turing-2069dc.netlify.com/?t=catalogs

Data provided is for illustrative purposes only and I manually defined much of the metadata so any errors/inconsistencies with the official data sources or the schema are mine.

I organized the source imagery into its own collection with `rel` links within the training and output items pointing to them as `rel : source` and `td:assets : rgb`. Redundantly, I also added the source imagery COGs as assets within each training and output item which nicely renders them as previews in the stac-browser. Any better way to do this?

Also still looking for good ideas to organize ML outputs. Currently I have `job...` items under the `outputs` collection and all the artifacts as assets within the item. This is inspired by the spacenet-stac example with ml_exports: https://spacenet-stac.netlify.com/ml-exports

cholmes:
Awesome demo @daveluo! Cool to see it in action in stac browser.

It'll take some modifications to STAC Browser, but we talked about for training data just referencing a single 'master' COG and then making the bounding box meaningful, where a renderer would just render the portion of the source image that is in the bounding box.

The other cool change to make to STAC Browser is to render the geojson asset in addition to the COG. If I'm tempted to try to make that PR for STAC Browser, but my time is way too tight these days... It's nice you've got the gist link now so that anyone can click and see it. To have stac browser render it it'd probably have to be a link to the 'raw' gist https://gist.githubusercontent.com/daveluo/c743c6b0f99795336636a1b0084786b5/raw/2d28c8638a1018e5f581cb7e390f0859ac538810/znz-example-labels.json or else teach stac browser to be able to get the raw from a gist geojson.

cc @mojodna

extensions/training-data/README.md
matthewhanson: It's a good idea @lossyrob , so we could: 

- have a "value" property in the asset features rather than a "label" property
- td:classes renamed to td:values which could contains the list of possible values. Although for something like 'count' this would be unwieldy.

How would you define the task?  Right now we have this "label_type", which is: "One of 'classification', 'detection', or 'segmentation'", but I agree this is unnecessarily limiting. Going to write some more thoughts below.

matthewhanson:
Thanks for the example @daveluo , great seeing it up and with a STAC Browser deployed.

I think the way that you have organized things is fine If your source imagery was already represented in a STAC catalog somewhere you could skip the "imagery" collection and just point to that Item directly.

Here's another example catalog where we simply point to a record in the Sentinel STAC catalog as our source rather than having a separate catalog:
http://mlhub-earth.s3.amazonaws.com/catalog.json

Although note those links are currently incorrect :-(, I've got to fix that.

It's a good point about having the COGs as assets in multiple Items, this does make it easier to preview them.

What we are doing with the catalog above, and I recently just pushed some changes to this PR to reflect it, is you provide an optional "rendered image" asset in the `Item`, because in many cases the training data was generated not using the original source imagery as is, but rather rendered in some way. For example, for Sentinel there's no RGB true color image available, so we create that and save it as an asset along with our training data `labels` asset and include it as an asset in the Item.

matthewhanson:
@lossyrob brought up a good point about supporting regression, it would be nice if the extension could handle both regression and classification tasks.

So perhaps the following changes:
- change "td:label_type" to "td:type" to include "regression", so options would be ['regression, 'classification', 'detection', or 'segmentation']
- Change `td:label_property` to be `td:value_property` which will be the property name containing the value (numeric, string) based on the type.
- Change `td:classes` to be optional, and is not used with regression tasks
- Add in an optional `td:value_range` field which is a 2 element numeric field containing the range of values for regression tasks

Thoughts?

daveluo:
Thanks for the thoughts!

@matthewhanson
Agreed on having optional "rendered image" asset items to capture intermediary steps that get us to train-able datasets. In addition to your example of rendering RGB or other band combos out of source data, I could see other pre-processing steps as rendered assets as well, e.g. realigning/correcting polygons to base imagery. My initial thought is to keep the spec open/flexible and leave it up to the catalog/item maker to decide what is defined and shared as a rendered asset during preprocessing. Could range from everything (rendered assets + notebooks/scripts demonstrating the pipeline) to nothing (just sharing the finalized rendered inputs for ML).

Also agreed with the idea of generalizing labels beyond classification. I need to think more about how generic this should be w.r.t. different ML tasks. There could be more than one label type for a task, ie. object detection with both regression (of bounding box coordinates) and classification (of object within each bbox) labels. Maybe regression and classification are two primary label types and multiple types are allowed within an item to flexibly suit the particular ML task? I'll try out some examples to see what may work well.

@cholmes, @mojodna 
Would be awesome to render geojson directly in the preview, either as a new layer on top of base COG imagery or separately displayed. Having a bounding box to render portions of a master COG would be cool too, although I've love to see the geojson preview first!

daveluo:
Updated stac4ml demo catalog & browser at https://zen-turing-2069dc.netlify.com with some new things for consideration:
- added a "chips" asset to each item which points to a folder of preprocessed image/mask chips. Another example of an asset rendered from raw source imagery+label. Could also include in description the details of how chipping was done, i.e. "512x512 z19@2x tile jpgs and 3-channel RGB mask pngs"
- this training data presents both segmentation and classification (of building condition) tasks so I've listed both within `td:type`. Right now, the class labels are contained within the geojson for each building polygon but I can make that into another asset, i.e. a csv of building_id,building_condition,WKTstring which also would show a different way of presenting training labels
- modified `geometry` of items & stac-browser leaflet properties to render all the building polygons as one nicely preview-able MultiPolygon. Not the ideal way to do this of course, just for illustrative purposes atm.

nrweir:
@matthewhanson, I'm interested to hear if you think this still warrants the [WIP] tag once you've given it a read through. Thanks for checking it out.

cholmes:
Can someone resolve the conflicts? I'm pretty happy to merge after all the work the group put in last week.

aaronxsu:
@cholmes I am happy to resolve the conflicts.

daveluo:
Hi everyone, I've updated my unofficial collection(s) of labeled data samples for ML over at https://zen-turing-2069dc.netlify.com/?t=catalogs per the latest label extension schema. What's missing are links to sources but everything else should conform to what we envisioned during the sprint.

The zanzibar and spacenet samples demonstrate the count object for summing up numbers of features contained in each label item. For @mojodna @matthewhanson, there are also examples of raster labels (and DSM/DTM layers) in the potsdam and urban3d collections. I'll try to gather and present some samples of object detection (probably from xView dataset) and regression tasks next chance I get to work on this.

If it'll be helpful now or later in a more refined state, I can also deploy a browser to this on the stac.cloud site and list it as an example. As a next step, I can see this being helpful as a browsable gallery of samples representing all major open geospatial datasets labeled for ML that folks can easily check out to see if it's what they need without downloading the whole datasets first.

cholmes:
Thanks for trying @aaronxsu - I just busted out the github web editor to resolve it (and it actually loaded it this time), so we should be good.

Though looks like we need more reviewers now.

matthewhanson:
I have reviewed this, but i am unable to officially add my review since I'm the one who opened the PR, so we'll need someone else to add theirs.

Just a few things:
- The Examples linked to in the Label README need to be fixed. The link for the second one is incorrect, it's using .json rather than .geojson as the extension.
- Not all of the examples in the directory are linked to from the README
- I think the names of the examples can be cleaned up - for instance they don't need to have "sprint4" in the titles, they should be shorter and simpler
- I'm not sure llabel:version belongs here. Not because labels don't have a version, but because that's not specific to Labels. I think we need to talk about versions in Items (or assets) as a separate issue, until then versions should be included in the Item ID if you want to track versions.

cholmes:
Discussed on call: Make issue for versioning and remove out of here. Ideally fix Matt's suggestions, but we are also ok to merge.

cholmes:
@daveluo @aaronxsu and @nrweir - are any of you able to fix the things Matt raised? 

And please raise the 'version' issue, as it's something we should propose at the Item level, as it should be consistent across extensions. Unless you have a reason that it is special to labeling.

nrweir:
Yes, I'm happy to do the fixes, will work on them now - sorry about the delay, I've been tied up at CVPR.

nrweir:
@matthewhanson I created a PR against this branch to incorporate the discussed changes.

extensions/label/README.md
```diff
@@ -0,0 +1,163 @@
...
+| counts          | [Count Object]  | Counts                     | An object containing counts for categorical data. |
+| statistics      | [Stats Object]  | Statistics                 | An object containing statistics for regression/continuous numeric value data. |
+
+`label:overview ` generally won't have both counts and statistics, but one of the two.
+
+#### Count Object
+
+| Field Name      | Type            | name                       | description       |
+|-----------------|-----------------|----------------------------|--------------------------------------------------------------------------------------------------|
+| class_name      | string          | Class Name                    | The different possible classes within the property `name`. |
```
m-mohr: Shorten the name from `class_name` to `name` or `id`? (see #499)

extensions/label/README.md
```diff
@@ -0,0 +1,163 @@
...
+    ]
+  }
+
+```
+
+#### Stats Object
+
+| Field Name      | Type       | name                       | description       |
+|-----------------|------------|----------------------------|--------------------------------------------------------------------------------------------------|
+| stat_name       | string     | Stat Name                  | The name of the statistic being reported. |
```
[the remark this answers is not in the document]
nrweir: I'm good with that. I think id would be confusing to data scientists...what’s an id with relation to a stat? Would prefer name.

extensions/label/README.md
nrweir: Same with above, I think `name` would probably make more sense, but no problem changing it from `class_name`.

I'll make the updates.

nrweir:
@matthewhanson @m-mohr I changed those two labels to `"name"` and also merged in `dev` to address the branch being out of date with base, then created another PR into this branch: #504.

m-mohr:
Should be ready for the merge. Can you confirm, @matthewhanson (and others)?

matthewhanson:
Looks good to me, and it's got approvals so I'm merging.

PR #370, opened 2019-01-02
Minor typo in the Single Item Extension header

m-mohr:

PR #371, opened 2019-01-07
Ensure uniformity of jsonspec standards

brianbancroft:
Brings the geojson stac spec schema reference up to draft-06, to the same version of the other stac schemas. Used to minimize hangups with third-party software. Based off of a conversation in gitter with @cholmes 

The resulting JSON standard file was checked against https://www.jsonschemavalidator.net/ to ensure validity with json-spec draft-06.

### DIFF
--- a/item-spec/json-schema/geojson.json
+++ b/item-spec/json-schema/geojson.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "http://json-schema.org/draft-04/schema#",
+  "$schema": "http://json-schema.org/draft-06/schema#",
   "id": "geojson.json#",
   "title": "GeoJSON Object",
   "type": "object",


brianbancroft:
The aim is for this to be a hotfix to version 0.6.0. Wouldn't #360 make for a minor version change in itself?

cholmes:
@m-mohr - my thought with this one was to do it against master and issues a 0.6.1 release. And #360 would be against dev, as it's arguably a bit bigger change. Like this one is just to make things consistent on master - feels like an oversight. #370 is to get us to the next version.

Curious what others think about the approach cc @matthewhanson @hgs-msmith @mojodna If people are for it I can merge and cut a 0.6.1 release, hopefully within a week.

m-mohr:
@cholmes Fine for me, but if we do a 0.6.1 anyway, I'd include the other fixes as well: #346, #349, #364, #370 and all the updates on implementations.

hgs-msmith:
I also am fine with a 0.6.1 release, and I agree we should tie up the loose ends cited by @m-mohr.

m-mohr:
Changed this PR to go towards the 0.6.1 branch, which will contain all changes for a 0.6.1 release. We don't need to port this PR to dev as #360 will make the changes in this PR obsolete anyway.

PR #372, opened 2019-01-08
Explicitly using 'intersect' for search

fredliporace:
Based on our last conference.


api-spec/README.md
```diff
@@ -33,8 +33,7 @@ pageable links to iterate through any results past that limit.
 ## Dynamic Catalog API
 
 The other endpoint that is included as an option in STAC API is `/stac/`, which implements the [STAC Catalog Spec](../catalog-spec/README.md).
-See the [Dynamic Catalog](https://github.com/radiantearth/stac-spec/blob/structure_and_cleanup/catalog-spec/catalog-spec.md#dynamic-catalogs)
-section of that spec for more information. Implementing this enables tools like
```
fredliporace: Link is broken, not sure if there is need to reference a specific section in catalog-spec/README.md

item-spec/item-spec.md
```diff
@@ -41,7 +41,7 @@ inherited from GeoJSON.
...
+| bbox       | [number]                                                                   | **REQUIRED.** Bounding Box of the asset represented by this item. Specified in Longitude/Latitude based on EPSG:4326 - first two numbers are longitude and latitude of lower left corner, followed by longitude and latitude of upper left corner. This field enables more naive clients to easily index and search geospatially. Most software can easily generate them for footprints. STAC compliant APIs are required to compute intersection operations with the item's geometry field, not it's bbox.                 |
 | properties | Properties Object                                                          | **REQUIRED.** A dictionary of additional metadata for the item.                                                                                                                                                                                                                                                                    |
```
fredliporace: This is where I saw a reference to the naive implementation that led me to believe that a simple search based on bounding boxes intersection was allowed for STAC APIs implementations. After reading again it is clear that this refers to general clients.
Not sure about the lower left - upper right corner convention, suggesting the same convention defined for the search API.

Issue #373, opened 2019-01-20
absolute/relative references


fredliporace:
See #160 and #56 for related discussion.

Definitions with examples:

* absolute url
  * `https://cbers-stac-0 6.s3.amazonaws.com/CBERS4/MUX/066/096/CBERS_4_MUX_20181024_066_096_L2.json`
* absolute path reference: `/CBERS4/MUX/066/096/CBERS_4_MUX_20181024_066_096_L2.json`
  * starts with / character
  * No use of ../
* relative path reference: `066/096/CBERS_4_MUX_20181024_066_096_L2.json`, `../catalog.json`
  * does not start with / character

Pros/Cons:

* absolute url
  * PRO: Possibility to directly get web address from STAC file. 
  * PRO: No parse needed to obtain absolute link.
  * CON: Requires data to be online. 
  * CON: Requires definition of __site and protocol__ when creating STAC file.
  * CON: Requires __global__ definition of the files organization structure.
  * CON: All STAC files need to be updated if __site or organization structure__ is changed.
    * That includes the case where the STAC set or subset is copied to other location, if we assume that it is desired for the references to link to data in the same 'location'.
* absolute path
  * Simple parse needed to obtain absolute link when browsing (prefix with site)
  * PRO: Does not require data to be online.
  * PRO: Does not require definition of __site and protocol__ when creating STAC file.
  * CON: Requires __global__ definition of the files organization structure.
  * CON: All STAC files need to be updated if __organization structure__ is changed.
    * That includes the case where a STAC subset is copied to other location,  if we assume that it is desired for the references to link to data in the same 'location'.
  * CON: Given only the STAC file its web address need to be obtained from other mechanism.
* relative path
  * CON: Parse needed to obtain absolute link when browsing.
    * Deal with multiple ../, it's possible that distinct relative paths point to the same file.
  * PRO: Does not require data to be online.
  * PRO: Does not require definition of __site and protocol__ when creating STAC file.
  * PRO: Requires __local__ definition of the files organization structure.
  * PRO: Only some STAC files need to be updated if __local organization structure__ is changed.
  * CON: Given only the STAC file its web address need to be obtained from other mechanism.
  * PRO: Supports set/subset copies, breaking only the links to elements not copied.

I'd say that if I had to choose one model I'd go with absolute path, with a self reference optionally including an absolute url (permalink) if this information is available when creating the STAC files.

__Update 02/04/2019__

[...]

PR #382, opened 2019-02-04
[WIP] Absolute/relative references

fredliporace:
Related to #373


item-spec/item-spec.md
```diff
@@ -106,13 +106,20 @@ A more complete list of possible 'rel' types can be seen at the [IANA page of Li
...
+implementors to provide relative references such as `/foo/bar.json`.
+
+In general, Catalog API's should aim for absolute URIs
 whenever possible. But Static Catalogs are potentially more portable if they can be implemented with
-relative links, so that every link doesn't need to be rewritten when the data is copied. The `self`
-link is required to be absolute.
+relative references, so that every link doesn't need to be rewritten when the data is copied to
+a different server, for instance.
+
+It is however required for all relative references to use absolute paths, that is, the references
```
m-mohr: They are still relative to something, either the root directory of the file system or the root catalog folder. It needs to be clarified what the root directory is.

fredliporace: Maybe [this](https://tools.ietf.org/html/rfc3986#section-4.2)?

item-spec/examples/CBERS_4_MUX_20181029_177_106_L4.json
```diff
@@ -49,15 +49,15 @@
   "links": [
     {
       "rel": "self",
-      "href": "https://cbers-stac-0-6.s3.amazonaws.com/CBERS4/MUX/177/106/CBERS_4_MUX_20181029_177_106_L4.json"
```
whistler: Maybe the example should show the self link as a absolute URL since thats whats recommended in the description below.

item-spec/item-spec.md
fredliporace: Yes, you could still navigate through the links by generating 'relative' links comparing self and the other references, but I agree that this is cumbersome and not standard.

Note however that absolute paths are allowed as part of URIs, so if we say nothing I think they would be allowed.. it's ironic but maybe the conclusion of this exercise will be to __forbid__ absolute paths at all, except for self when data is offline or the provider is not confident to consider an absolute URL as a permalink.

item-spec/examples/CBERS_4_MUX_20181029_177_106_L4.json
fredliporace: I agree, I chose to use absolute path as an example of a possible way to write the item but since the same item has absolute URLs it contradicts the recommendation. It would be better to use absolute paths for all references, representing the case where the data is not online,

cholmes:
Closing this, as we got a cleaner, smaller one in for this.

PR #387, opened 2019-02-04
Replaced id with $id in schemas

### LINKED ISSUE #386: Replace id with $id in specs
The STAC specs are using JSON Schema draft-06 and the old `id` keyword instead of `$id`.
https://json-schema.org/draft-06/json-schema-release-notes.html

jbants:
fixes #386


m-mohr:
Looks good, may also be a thing to be added to a 0.6.2 release. cc @matthewhanson

PR #390, opened 2019-02-05
Update dev to include changes for 0.6.1

m-mohr:
We forgot to merge several changes for 0.6.1 into dev, which means dev is outdated. This fixes the issue.


matthewhanson:
Ah ok, this explains why the diff between dev and master was larger than I expected.

PR #398, opened 2019-02-07
Enforce the actual version number for stac_version.

m-mohr:
Enforce the actual version number for stac_version. Otherwise you could add anything, but we actually want the specific version number in there. This also helps to enforce non-mixed catalogs, which is a requirement in the spec.

### DIFF
--- a/catalog-spec/json-schema/catalog.json
+++ b/catalog-spec/json-schema/catalog.json
@@ -39,7 +39,8 @@
       "properties": {
         "stac_version": {
           "title": "STAC version",
-          "type": "string"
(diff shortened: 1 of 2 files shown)


m-mohr:
Would like to get approval from @jbants or @brianbancroft that this does work well with the validator before merging. May even help with https://github.com/sparkgeo/stac-validator/issues/34

jbants:
Thanks @m-mohr. We'll have a look.

cholmes:
Any word @jbants ? I'd love to get this merged, and even to 0.6.2

PR #380, opened 2019-01-28
Fix typo in Item Specification bbox description

whistler:
A bounding box cannot be defined by four numbers without specifying the right most coordinate. Fixing typo to use the top right corner instead of top left. This is consistent with how GeoJSON defines bbox, lowest coordinate values first followed by highest coordinate values.

PR #400, opened 2019-02-20
Typo on bbox field in item-spec

youngpm:
bbox should ready lower left, upper right vs lower left, upper left


m-mohr:
Thank you for the contribution. This has already been fixed in #380 and will be released in STAC v0.6.2 soon. Closing as duplicate of #380.

youngpm:
ah thanks, sorry for missing it, guess that's why i didn't see it in the open issues!

PR #407, opened 2019-02-26
Document 0.6.1 ISERV catalog

mojodna:


m-mohr:
Not sure how useful it is to link to implementations of old versions if there is a newer one available.

mojodna:
Agreed. Historical context is mildly interesting and CBERS had both versions listed. For ISERV, Raster Foundry expects the old version (or so I understand). We can clean up at some point.

PR #409, opened 2019-02-28
Data cube extension: Better JSON Schema and vertical dimensions better described

m-mohr:
I had some back and forth with myself in the last days how to best describe the vertical dimensions. The previous draft was not very strict in enforcing several aspects such as the axis only used for spatial dimensions. Now it is clearly defined and strictly checked in the JSON schema. In the meantime the other PR was already merged so we need another iteration, sorry for the inconvenience. Should have communicated that in the other PR.


cholmes:
No problem. Idea was to have PR's so we can more granularly see what's happening, so this is in line with that.

PR #411, opened 2019-02-28
Update implementations.md

cholmes:
Added links to the various stac.cloud STAC Browsers, for easy visualization.


matthewhanson:
I'm going to add the Sentinel catalog to this as well, then will add review.

PR #414, opened 2019-03-05
Related to Issue #373, allowing absolute references for self

fredliporace:


item-spec/item-spec.md
```diff
@@ -94,7 +93,7 @@ The following types are commonly used as `rel` types in the Link Object of an It
 
 | Type         | Description                                                  |
 | ------------ | ------------------------------------------------------------ |
-| self         | **REQUIRED.** _Absolute_ URL to the item file itself. This is required, to represent the location that the file can be found online. This is particularly useful when in a download package that includes metadata, so that the downstream user can know where the data has come from. |
+| self         | **REQUIRED.** _Absolute_ URL to the Item if it is publicly online, or an _absolute reference_ to the item in the form ```/path/to/item``` otherwise. This is particularly useful when in a download package that includes metadata, so that the downstream user can know where the data has come from. |
```
m-mohr: Shouldn't that be called "absolute path" instead of `absolute reference`?

collection-spec/collection-spec.md
```diff
@@ -97,7 +97,106 @@ The following types are commonly used as `rel` types in the Link Object of a Col
 
 | Type    | Description                                                  |
 | ------- | ------------------------------------------------------------ |
-| self    | **REQUIRED.** *Absolute* URL to the collection. This is required, to represent the location that the file can be found online. This is particularly useful when in a download package that includes metadata, so that the downstream user can know where the data has come from. |
+| self    | # STAC Collection Specification
+
+The STAC Collection Specification defines a set of common fields to describe a group of Items that share properties and metadata. The
```
m-mohr: Here's something wrong, text got duplicated.

fredliporace: Yes, noticed that also and fixed right after.

catalog-spec/catalog-spec.md
```diff
@@ -215,7 +215,7 @@ The following types are commonly used as `rel` types in the Link Object of a STA
 
 | Type    | Description |
 | ------- | ----------- |
-| self    | **REQUIRED.** _Absolute_ URL to the catalog file itself. This is required, to represent the location that the file can be found online. This is particularly useful when in a download package that includes metadata, so that the downstream user can know where the data has come from. |
+| self    | STRONGLY RECOMMENDED. _Absolute_ URL to the location that the catalog file can be found online, if available. This is particularly useful when in a download package that includes metadata, so that the downstream user can know where the data has come from. |
 | root    | URL to the root STAC Catalog or [Collection](../collection-spec/README.md). Catalogs should include a link to their root, even if it's the root and points to itself. |
```
mojodna: Can/should we strongly recommend this one too?

item-spec/item-spec.md
```diff
@@ -77,27 +77,34 @@ datetime fields.
...
 | ---------- | ------ | ----------------------------------------------------------------------------------------------------------------------------------- |
-| href       | string | **REQUIRED.** The actual link in the format of an URL. Relative and absolute links are both allowed.                                |
+| href       | string | **REQUIRED.** The actual link in the format of an URL. Relative and absolute links are both allowed unless otherwise noted in the ```rel``` type description.                    |
 | rel        | string | **REQUIRED.** Relationship between the current document and the linked document. See chapter "Relation types" for more information. |
 | type       | string | Media type of the referenced entity. 																								|
 | title      | string | A human readable title to be used in rendered displays of the link. |
 
+#### Relative vs Absolute links
+
+Currently the JSON schema for links does not require a URI formatting, to give the option for
```
mojodna:
```suggestion
Currently, the JSON schema for links does not require them to be formatted as URIs, to allow
```

item-spec/item-spec.md
```diff
@@ -77,27 +77,34 @@ datetime fields.
...
+implementors to provide relative links. In general, Catalog API's should aim for absolute links
```
mojodna:
```suggestion
implementors to provide relative links. In general, Catalog APIs should aim to provide absolute links
```

item-spec/item-spec.md
```diff
@@ -77,27 +77,34 @@ datetime fields.
...
+whenever possible. But Static Catalogs are potentially more portable if they can be implemented with
```
mojodna:
```suggestion
whenever possible. Static Catalogs are potentially more portable if they incorporate only
```

item-spec/item-spec.md
```diff
@@ -77,27 +77,34 @@ datetime fields.
...
+recommendations for particular ```rel``` types are given in the ```rel``` type description.
+
 #### Relation types
 
 The following types are commonly used as `rel` types in the Link Object of an Item:
 
 | Type         | Description                                                  |
 | ------------ | ------------------------------------------------------------ |
-| self         | **REQUIRED.** _Absolute_ URL to the item file itself. This is required, to represent the location that the file can be found online. This is particularly useful when in a download package that includes metadata, so that the downstream user can know where the data has come from. |
+| self         | STRONGLY RECOMMENDED. _Absolute_ URL to the Item if it is publicly online. This is particularly useful when in a download package that includes metadata, so that the downstream user can know where the data has come from. |
```
mojodna:
```suggestion
| self         | STRONGLY RECOMMENDED. _Absolute_ URL to the Item if it is available at a public URL. This is particularly useful when in a download package that includes metadata, so that the downstream user can know where the data has come from. |
```

catalog-spec/catalog-spec.md
cholmes: Probably. I can tackle that in my PR on best practices, as I'll likely hit the core a bit more too. I think 'root' is a bit vague in the current writing, so can flesh it out some more.

item-spec/item-spec.md
cholmes: It sounds like we should be able to use another JSON construct (uri-ref or something? @m-mohr was talking about this) so the schema can validate relative and absolute links.

catalog-spec/catalog-spec.md
fredliporace: Will do.

fredliporace:
@mojodna I think that all reviews are addressed, let me know if I missed anything.

Issue #383, opened 2019-02-04
Searching by ID / Collection ID


matthewhanson:
Because `id` is a top level field, it is not queryable with the `query` parameter. However there is also no other parameter in the search body for ID (only bbox, time, and intersects).

http://sat-utils.github.io/sat-api/#tocssearchbody

This means that there's no way to query by ID, you can only form the complete self link which requires the Collection ID, e.g., http://landsat-stac.s3.amazonaws.com/landsat-8-l1/151/018/2015-09-11/LC81510182015254LGN00.json

Which does seem a little strange that you can't search by Item ID and can only look it up if you know the collection.

I'd propose a new search field, `ids`, which allows a list of `ids` to be provided and a FeatureCollection returned of those Items. The one possible drawback here is that the Item ID's, while unique across a collection, may not be unique across a catalog. But I think that's fine, this would return a FeatureCollection so if there is more than one ID match then they would all be returned (note that in sat-api ids *are* required to be unique, but this could be different in other implementations).

m-mohr:
In the actual OpenAPI files there are fragments for featureId and collectionId, but they are not referenced at all. So it seems somebody had the idea, but at some point it got lost (or I am missing something here).

joshfix:
Back when we still had the `/items` endpoint, we also had an `/items/{id}` endpoint.  This seems to have gotten lost, maybe when we moved to the `/search/stac` endpoint. 

https://github.com/radiantearth/stac-spec/blob/e014c8eef7069ed24d3647eed572bfb5d2e4c053/api-spec/spec.yaml#L107

I actually did not realize this was dropped and still support it in my implementation but under `/search/stac/{id}`:

https://stac.boundlessgeo.io/search/stac/LC08_L1TP_058017_20180721_20180721_01_RT

cholmes:
Definitely +1 on adding an id search. Curious about the Harris implementation - cc @hgs-msmith and @hgs-truthe01 - if there was something in there for id search. 

I'm happy with whatever @matthewhanson @joshfix and the Harris folk agree on, since I think those are our three main api implementations.

I think /items/{id} is part of the WFS 3 specification, so I think it'd be useful to still try to use that convention. Which sat-api does, like https://sat-api.developmentseed.org/collections/landsat-8-l1/items/LC81810982019043LGN00 But I agree the search endpoint should have the ability to respond to a set of ID's. 

Happy for a PR and to get this into 0.7.0

joshfix:
[...]

PR #415, opened 2019-03-05
Redefined fields extension to allow filtering on any property in document

joshfix:
The fields extension provides the ability to specify individual fields that should be included or excluded but only applied to properties, and also defined a boolean to specify whether `geometry` should be included as well.
The new definition allows you to specify any field in the JSON document (including root-level fields), but requires that you provide the full JSON path.  Eg, to include only id, bbox, cloud cover, and datetime, you would populate the `include` array with `id,bbox,properties.eo:cloud_cover,properties.datetime`.


api-spec/extensions/fields.fragment.yaml
```diff
@@ -42,3 +41,9 @@ components:
           type: array
           items:
             type: string
+      example:
+        include:
+          - id,properties.eo:cloud_cover
+        exclude:
+          - geometry,properties.datetime
+
```
fredliporace: Is it implicit that we will allow wildcards? If yes then I'd include one as example.

api-spec/extensions/fields.fragment.yaml
```diff
@@ -26,14 +26,13 @@ components:
...
       description: |
-        The geometry member determines whether the geometry is populated or is null. The
-        include and exclude members specify an array of property names that are either 
-        included or excluded from the result, respectively. If both include and exclude 
-        are specified, include takes precedence.
+        The include and exclude members specify an array of
+        property names that are either included or excluded
+        from the result, respectively. If both include and
+        exclude are specified, include takes precedence.
+        Values should include the full JSON path of the property.
```
fredliporace: Are we supporting the 'dot' notation only?

cholmes:
@matthewhanson and @hgs-msmith + @hgs-truthe01 - could you all review this? Michael + Tim - we talked about this on our last call, out of the discussion on #383. Matt and Josh both felt this worked well for them, but we definitely wanted your input on it.

api-spec/extensions/fields.fragment.yaml
hgs-truthe01: I wasn't sure I wanted to force wildcard in, but if others feel strongly, I'm not against the idea. I worry that once we start using "smart" fields we will need something like regex. I was hoping that explicitly naming the fields would be enough to start.

api-spec/extensions/fields.fragment.yaml
hgs-truthe01: That was true before this change for nested fields. It might be worth mentioning though.

api-spec/STAC-query.yaml
```diff
@@ -631,6 +624,11 @@ components:
           type: array
           items:
             type: string
+      example:
+        include:
+          - 'id,properties.eo:cloud_cover'
+        exclude:
+          - 'geometry,properties.datetime'
```
hgs-truthe01: Missed these on first pass. Same as below on yaml array declaration.

api-spec/README.md
```diff
@@ -54,9 +54,9 @@ The [`definitions` folder](./definitions/) contains the YAML files that will imp
 
 See [yaml-files](https://www.npmjs.com/package/yaml-files) for the syntax to import YAML fragments.
 
-### Fragments
+### Extensions
```
hgs-truthe01: Was the change from fragments to extensions intended to be in this PR as well?

hgs-truthe01:
@cholmes I can see the benefit of including the full path here. I do worry that it is not in sync with the query extension now. Perhaps we should concede that one as well and use properties.{name} there as well.

api-spec/README.md
m-mohr: I guess that was changed as the folder is called extensions, not fragments.

matthewhanson:
@hgs-truthe01 We did talk about that (not being in sync with the query extension). I'm ok with them being different.  For fields, it really is best to be able to specify any field, property or not, for inclusion or exclusion, so the `property.` is necessary.

However, for the query extension the properties are the fields that users will want to query on. There are some exceptions at the top level (id, geometry, maybe assets), but these are typically objects not strings or numbers, so the standard querying isn't defined. So instead for those special fields we have additional search terms (e.g., intersects for querying geometry), and don't need the query search term to include them.

I'm good with this PR, any other opinions?

cholmes:
@matthewhanson - you want to merge this when you're ready? Just want to be sure it doesn't conflict with your other PR. Sorry for not getting this merged sooner.

matthewhanson:
Thanks, merging and will resolve conflicts with other PR

Issue #413, opened 2019-03-05
List non-common fields in Collection (Summaries)


m-mohr:
In openEO we need to list also the non-common fields in Collections, i.e. all properties that are available in the Items but have different values. We don't necessarily need to list the actual values (or extents), but we need to inform the user what he can query against. This is probably similar to what we plan to add for the assets as "asset definition" or "asset schema". So a collection with an "item schema" could something as simple as:

```json
{
    "id": "s2",
    "title": "Sentinel-2",
    "description": "...",
    "stac_version": "0.6.2",
    "extent": {
        "spatial": [-180,-90,180,90],
        "temporal": ["2013-06-01",null]
    },
    "properties": {
        "eo:gsd": 10,
        "eo:platform": "sentinel-2",
        "eo:instrument": "MSI",
        "eo:bands": []
    },
    "varying:properties": [
        "eo:epsg",
        "eo:cloud_cover",
        "eo:off_nadir"
    ],
    "links": []
}
```

I need to define this anyway for openEO in the next days, so the question I have is whether that could be something that is useful also for other users so we can combine efforts and standardize. Otherwise I'd probably start of as a proprietary extension.

This could be something that may also be useful for STAC API users that don't want to look into Items to find out what they could query against.

matthewhanson:
I think this is useful. I've always seen a Collection as a definition for the Items that are in it. So a user should be able to look at a Collection and get all the info they need to know about the Items without having to get Items to look at the structure.

I think a name like "available_properties" is better than "varying" though, and don't think it needs a prefix.

m-mohr:
Regarding the name: I understand "available_properties" as "all properties available", but we only list the properties that are not actually listed with fixed values in "properties", so that's why I chose varying. But I'm really open for a new name, because varying was just a synonym I googled for uncommon, which felt even worse. (Non-native speaker issues, I guess)... Happy to remove the prefix.

jbants:
How about `extended_properties`  or `additional_properties`?

cholmes:
Just spent a good 15 minutes thinking about this...

It seems to me like the ideal answer might be to have just a 'properties' object, that can contain:

* 'locked' properties (the common properties), just one value
* 'range' properties, that give the range you can query, or an enum.
* 'open' properties, that just have the property name and take any value.

So you could just have:

```
"properties": {
  "eo:gsd": 10,
  "eo:platform": {
      "values": [
        "Sentinel-2A",
        "Sentinel-2B"
      ]
[...]

PR #416, opened 2019-03-05
Property Summary Extension (#413)

m-mohr:
In openEO we need to list also the non-common fields in Collections, i.e. all properties that are available in the Items but have different values. We don't necessarily need to list the actual values (or extents), but we need to inform the user what he can query against. This is probably similar to what we plan to add for the assets as "asset definition" or "asset schema". See #413 for more information.

Here's the extension as PR.

Does somebody have a better name? Summary Extension? ...?


extensions/non-common/README.md
```diff
@@ -0,0 +1,29 @@
+# Non-Common Properties Extension Specification (`-`)
+
+**Extension [Maturity Classification](../README.md#extension-maturity): Proposal**
+
+This extension lists all properties, which don't have common values across the items. It's the counterpart to the [Commons extension](../commons/), which lists properties with common values across all items.
+
+This extension allows specifying the potential values or extents so that user's don't need to walk through STAC Items to find potential values to query for.
```
mojodna:
```suggestion
This extension allows specifying the potential values or extents so that users don't need to walk through STAC Items to find potential values to query for. 
```

extensions/non-common/README.md
```diff
@@ -0,0 +1,29 @@
...
+This extension allows specifying the potential values or extents so that user's don't need to walk through STAC Items to find potential values to query for. 
+
+- [Example](example.json)
+- [JSON Schema](schema.json)
+
+## Collection fields
+
+| Field Name       | Type                        | Description                                                  |
+| ---------------- | --------------------------- | ------------------------------------------------------------ |
+| other_properties | Map<string, Summary Object> | **REQUIRED.** A map of all properties used in STAC Items, which don't have common values across all items. The keys have to be the property names and the values allow to specify a summary of the values in the items as extent or set of values. |
```
mojodna:
```suggestion
| other_properties | Map<string, Summary Object> | **REQUIRED.** A map of all properties used in STAC Items which don't have common values across all items. The keys must be the property names and the values are allowed to specify a summary of the values in the items as extents or sets of values. |
```

mojodna:
I'm also having trouble with the name...

"Property Summary Extension" feels good to me, as it's summarizing the potential values that properties may have.

extensions/non-common/README.md
```diff
@@ -0,0 +1,29 @@
+# Non-Common Properties Extension Specification (`-`)
```
mojodna: Rename?

m-mohr: Yes, but there are still a lot more things to discuss, see #413.

m-mohr:
Will make a new PR to introduce the summary extension.

PR #418, opened 2019-03-07
[STAC-417] Added href to the asset called index

### LINKED ISSUE #417: EO extension example not validating properly
I think the index object is missing an `href`, otherwise it's not valid with item definition for `assets`

EO example:
https://github.com/radiantearth/stac-spec/blob/612ee4fb4f979c049d2416e7f5374a5301b8a277/extensions/eo/example-landsat8.json#L238-L241

Item spec:
https://github.com/radiantearth/stac-spec/blob/612ee4fb4f979c049d2416e7f5374a5301b8a277/item-spec/json-schema/item.json#L102-L121

jbants:
Fixes #417

### DIFF
--- a/extensions/eo/example-landsat8.json
+++ b/extensions/eo/example-landsat8.json
@@ -236,6 +236,7 @@
             "type": "image/jpeg"
         },
         "index": {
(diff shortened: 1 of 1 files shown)


cholmes:
Just made a 'backport to stable' tag to help us track the changes that should go on minor releases if we do them.

Issue #423, opened 2019-03-13
sar:absolute_orbit and sar:off_nadir more suitable for collections than items?


m-mohr:
After reading a comment from @walterggit in https://github.com/radiantearth/stac-spec/issues/405#issuecomment-470232131, I realized that I may was a bit too much influenced with my collection-level view when defining `sar:absolute_orbit` and `sar:off_nadir`. Currently, it allows quite complex ranges of numbers, but isn't the absolute_orbit and the off_nadir value exactly one value per item?

cholmes:
Yes. Certainly for off_nadir. I just dug in a bit to absolute_orbit, as I don't understand that from my Planet satellite perspective. Looking at ALOS metadata it does look like yes, there's just one per record. Looking at sentinel 1 it seems like they use: Orbit number (start) and Orbit number (stop), but in the two records I saw they were the same value, so maybe it ends up at just one item per?

cholmes:
Hrm, though looking at what @walterggit wrote about incidence angle I wonder if off_nadir in SAR can be more of a range.

walterggit:
For orbit number, I suppose the two numbers could potentially differ for products that cross the boundary between one orbit and the next (start could be in orbit N and end in orbit N+1).  From page 29 of https://www.eorc.jaxa.jp/ALOS-2/en/doc/fdata/PALSAR-2_xx_Format_CEOS_E.pdf (if that is a current spec), it looks like the orbit number at scene center is used for ALOS.  I think that one orbit number parameter is probably enough, so long as it is consistently derived for a given sensor (either start or stop or center, but not a mix).

I haven't used off-nadir for SAR, but for incidence angle it can be useful to have a two-value range even for a given scene, as it can vary quite a lot from near range to far range- e.g. for a single Sentinel-1 IW scene, it goes from 30.86 degrees at the start of IW1 to 46.00 degrees at the end of IW3 (see https://sentinel.esa.int/web/sentinel/user-guides/sentinel-1-sar/acquisition-modes/interferometric-wide-swath ).  The off-nadir has a similarly large variation.  Radarsat-2 can have a large or narrow range over a given scene depending on the beam mode, but the product metadata supplies both the near and far range incidence angles regardless.

cholmes:
Cool, makes sense on the orbit number. We hit a similar thing with time, where we started with a range, but went to a single time, that implementors should pick in a consistent way.

[...]

Issue #405, opened 2019-02-25
Add incidence_angle to SAR extension


walterggit:
I would recommend including an “incidence_angle” option in the list of optional SAR parameters.  The incidence angle is often specified instead of the off-nadir (elevation) angle in SAR product metadata.  The two angles are related, but not the same owing to the earth's curvature.

cholmes:
Thanks for the contribution @walterggit! Are you working on a SAR dataset in STAC that needs this? Like does sentinel-1? We still don't have any solid reference implementations for SAR extension, but I'd love to see one in soon. 

Or I'm happy to add this if more SAR experts sound in, but we have no one in our core STAC community who _really_ knows SAR. Or if you'd like to join and become that expert then you could make a Pull Request explaining this option in the spec.

walterggit:
For a definition for incidence angle, you could use "The incidence angle is the angle defined by the incident radar beam and the vertical (normal) to the intercepting surface.  Measured in degrees (0-90)."  It would be good if this could be specified as a either a single value (assumed to be at scene center) or a two-value array, to allow storage of near-range and far-range incidence angles.

cholmes:
Cool. If it needs to potentially be a two-value array then I think the way to do it is define it as an array, that can have one or two values.

The alternative would be to define two values - incidence_angle_center and incidence_angle_range or something like that. Are there cases where people would define both / be interested in both? Or is it better to just have one field?

walterggit:
I've used both in different contexts, but it would probably be more consistent with the off-nadir angle spec to keep it as a single parameter.  You could support specification of up to all of them by allowing:
[center_inc_angle], or
[near_range_inc_angle,far_range_inc_angle], or
[near_range_inc_angle, center_inc_angle, far_range_inc_angle]

cholmes:
@walterggit - I think we may be moving in the SAR extension to less use of the ranges. See #422 

And I think we're also wondering if off-nadir angle could just be a single value, or does it also need the ability to specify near and far ranges? I think it may make sense to have off-nadir and incidence angle both be more explicit, like have off_nadir_center and off_nadir_range.

Also I think the off-nadir example may just be an error? I too would expect it to have the first value be in between the 24.5 and 30. 

@walterggit - happy to take a recommendation on how to best specify these, like what has worked best to cover the range of sar metadata you've seen. We're still in 'proposal' stage to garner feedback from experts like you, so can change things around.

walterggit:
[...]

PR #426, opened 2019-03-20
SAR: incidence_angle, orbits and off_nadir (#405, #423)

m-mohr:
Ad discussed in #405 and #423:
* Added incidence_angle (#405)
* replaced absolute_orbit with relative_orbit (#423)
* redefined off_nadir to suit incidence_angle (#423)

Please review carefully. cc @walterggit @claxn


walterggit:
@m-mohr @cholmes @claxn : Regarding off-nadir angle, I have never used it for SAR search, and have more commonly seen the incidence angle referenced in the literature that I've read through.  Unless others need it, I think off-nadir angle could be dropped.

For the relative orbit number, the wording is a bit confusing.  The definition of orbit number under https://sentinel.esa.int/web/sentinel/user-guides/sentinel-3-altimetry/definitions/notations matches how I think of relative and absolute orbit numbers, though the reference point for the absolute first orbit and for the first orbit in an orbit cycle may vary with satellite.  The relative orbit number can be useful for finding images with potentially the same viewing geometry. 

Also, I haven't worked with UAVSAR, but looking around a bit, it seems this is an airborne SAR.  Sharing UAVSAR's flightID with satellite SAR orbit numbers in one parameter seems like it is combining two quite different things.  Of the two numbers, I'd say that absolute orbit number is more analogous to flight ID than relative orbit number is, but even that is a stretch.  It would make more sense to me to have flight_id as a separate optional parameter applicable to airborne SAR's.  Is there a plan to differentiate between parameters that are only applicable to airborne versus satellite SARs, or would that be noted in the parameter description?

The observation_direction parameter description looks good to me.

claxn:
The only convenient thing about the off nadir angle is that it is independent from latitude, thus the range of values is fixed for a given observation mode. However, I agree with @walterggit to drop it for now. I am also more used to incidence angles. 

Regarding the orbit numbers I would suggest to keep both. In case these entries should also work with airborne SAR as mentioned by @walterggit we should make it more general. One parameter could be `flight_id` or `track_id` (=absolute orbit number for spaceborne platforms) and an optional parameter `rel_flight_id` or `rel_track_id`.

m-mohr:
Thanks, highly appreciate your feedback @walterggit @claxn .

* I removed off_nadir.
* I re-added absolute_orbit and changed the description of both properties. Still using orbit in the name until we have a consensus to rename them.

Could you please re-review? Thanks!

walterggit:
@m-mohr @claxn : The incidence_angle_center looks good.  

For absolute and relative orbit numbers, I am wondering if it would be better to represent these each with a single value, as with incidence angle.  You don't really gain much by specifying both the start and end orbit numbers, especially since they are almost always identical.  Sentinel-1 defines these values at acquisition start and end; ALOS I think defines a single orbit number at scene center.  It would be a nuisance to extract the metadata if forced to make this anchor point consistent between sensors.  How about something like:

|  sar:relative_orbit  |  number   | A relative orbit number for the acquisition. The relative orbit number is a count of orbits from 1 to the number of orbits contained in a repeat cycle, where relative orbit 1 starts from a defined reference for a given satellite. |

|  sar:orbit_anchor  |  string   | Point in the acquisition that relative and/or absolute orbit numbers are specified for, either 'start', 'center', or 'end'. |

| sar:absolute_orbit    | number | An absolute orbit number for the acquisition. Usually corresponds to the number of orbits elapsed since satellite launch (e.g. ALOS, ERS-1/2, JERS-1, and RADARSAT-1, Sentinel-1). For airborne SAR such as UAVSAR it can be the [Flight ID](http://uavsar.jpl.nasa.gov/cgi-bin/data.pl) or a similar concept. |

Or maybe the anchor could be made part of the relative and absolute orbit definitions themselves (represent with a number and a string instead of just a number).

claxn:
@walterggit: I agree that it is only necessary to have a single value for each orbit number.  I would approve your proposal having those three metadata entries. In my opinion an anchor per orbit number is not necessary, since the numbering of both absolute and relative should follow the same convention.

m-mohr:
@claxn @walterggit I agree with all of your points, except that I don't like the point that a property (orbit_anchor) influences other properties (relative_orbit and absolute_orbit). This makes the latter two not so easily comparable across their range. I'm thinking to follow the thoughts behind the incidence angle and only allow the center. Does that make sense or do you see problems with this approach? I guess, the anchor usually shouldn't change across a collection, but if it does you run into problems when querying as for some results it's then based on start or center or end. That doesn't feel clean and correct. But I'm not sure whether the center can always be computed or retrieved.

walterggit:
@m-mohr @claxn : Sentinel-1 doesn't specify a center orbit number, you'd need to figure it out based on the logic they used to generate the start and end orbit numbers, which is not the job of a metadata extractor.  I do agree that having one parameter influence another could be error prone and unintuitive.  How about dropping the orbit_anchor, but allowing the orbit number to be either the center or start depending on what is available - just call it "orbit number associated with the acquisition" and add a note that this should be the center orbit number if readily available, otherwise the orbit number at the start of the flight?

m-mohr:
Thanks for the quick reply. I think that would work for me, yes. What do you think, @claxn ?

claxn:
@m-mohr @walterggit : Sorry for my late reply. Yes, I agree, this will make things easier! Then we will only have two attributes referring to orbit numbers, which will lead to less confusion at the user side.

m-mohr:
Thanks, I updated the PR accordingly. Feel encouraged to comment if there's still something to improve. Otherwise, I guess we can merge the PR.

m-mohr:
Resolved merge conflicts, we still need a review from a core contributor to be able to merge. @mojodna @matthewhanson ?

Issue #408, opened 2019-02-26
Best practices for static catalog filesystem layout


mojodna:
This is a bit more general than #403, but helps that effort substantially.

1. Root documents (catalogs / collections) should be at the root of a directory tree containing a static catalog (e.g. `/home/seth/static-catalogs/sample/catalog.json`)
2. Catalogs should be named `catalog.json` (cf. `index.html`)
3. Collections that are distinct from catalogs should be named `collection.json`
4. Items should be named `<id>.json`
5. Sub-catalogs _and items_ should be stored in subdirectories of their parent (and only 1 subdirectory deeper than a document's parent) (e.g. `.../sample/sub1/catalog.json`) -- this means that each item and its assets are contained in a unique subdirectory
6. ~Items should be stored in the same directory as their parent catalog (e.g. `.../sample/sub1/item1.json` when a part of the `sub1` catalog)~
7. [Reverse relative links](https://github.com/radiantearth/stac-spec/issues/373#issuecomment-466884051) (name, implementation, necessity TBD) should not start with `../` ((5) should satisfy this)
8. Limit the number of items in a catalog or sub-catalog, grouping / partitioning as relevant to the dataset

m-mohr:
I like that.

A note regarding (4): This could lead sometimes to name conflicts, I guess. If the asset is a GeoJSON called abc.json and the id for it is abc, then the item file would also be named abc.json and that wouldn't work as you can't have files with the same name in a folder. We'd need a prefix or suffix for our items, maybe abc.item.json or so. @cholmes

cholmes:
Sounds great, I like the idea of recommendations. @mojodna - if these are followed will that enable stac browser to make nice URL's with the performance of the encoded stuff now? 

I'm comfortable with us 'taking' <id>.json. Especially since this seems like a 'best practice' / recommendation thing. Spec does not say you have to name it that, and so if you have a good reason to not name it that it's ok. 

I also don't love the notion of using STAC to just refer to a JSON document. Imho you should be using a WFS, and refer to that service. Like a set of JSON features should be in a collection, not in an item. There is cases where I think it's ok, like the 'training data extension', where the 'asset' is the geotiff _plus_ the geojson. But in that case I think the training geojson doesn't need to 'take' the <id>.json - it can easily rename.

matthewhanson:
Great ideas @mojodna 

[...]

PR #428, opened 2019-03-20
Catalog best practices

cholmes:
First go at capturing some of the best practices for catalogs, related to the rel vs absolute stuff that I detailed in https://docs.google.com/document/d/1-P5YLNWz70OSOqK_rdEOwN6qU6Bl1KIQbkOmp-UkpCM/edit?ouid=105927010353605742300&usp=docs_home&ths=true&dl=CgYKBAgBEAE.

I tweaked the language a bit on the best practices, so that in the next paragraph I could more easily recommend to map the concepts to dynamic catalogs as well, so we get good structures out of them too.

@mojodna - I cut out the href:from stuff for now. I think we should put the core idea in an extension, and then I'm happy to add it back in here as a best practice. 

@fredliporace - review on all is appreciated, but do check out the recommendations on going static <-> dynamic in particular. I wasn't sure if you were going to stick with the direct references to the static catalog in the API responses - I remembered you said there were some issues, but not sure where you landed.


m-mohr:
We should introduce links in the README files, otherwise people can't find it easily.

cholmes:
Ah, yes, I meant to put in a number of links to this document. Changed title to WIP, and will work to get those done.

catalog-spec/catalog-best-practices.md
```diff
@@ -0,0 +1,119 @@
...
+and assets, and how to name them. The specification leaves this totally open, and you can link things as you want. We
+encourage people to explore new structures. But the following are what a number of implementors ended up doing. Following
+these recommendations makes for more legible catalogs.
+
+1. Root documents (catalogs / collections) should be at the root of a directory tree containing the static catalog.
+2. Catalogs should be named `catalog.json` (cf. `index.html`).
+3. Collections that are distinct from catalogs should be named `collection.json`.
+4. Items should be named `<id>.json`
+5. Sub-catalogs should be stored in subdirectories of their parent (and only 1 subdirectory deeper than a document's parent) (e.g. `.../sample/sub1/catalog.json`).
+6. Items should be stored in subdirectories of their parent catalog (e.g. `.../sample/sub1/3425/3425.json`).
```
fredliporace: I'd change the example to remove the directory, ```../sample/sub1/3425.json```, just to make clear that it is not required for an item to have a unique subdir with the same name, see https://github.com/radiantearth/stac-spec/issues/408#issuecomment-468242880. I assume that recommendation (7) below was created based on https://github.com/radiantearth/stac-spec/issues/408#issuecomment-468369974, and the idea there is that if the number of itens in a subdir is 'reasonable' then it is OK to have more than one item in a single subdir.

catalog-spec/catalog-best-practices.md
```diff
@@ -0,0 +1,119 @@
...
+* **Use Asset `href` links consistently**: The links to the actual assets are allowed to be either relative or absolute. There
+are two types of 'self-contained catalogs'. The first is just the metadata, and use absolute href links to refer to the 
+online locations of the assets. The second uses relative href links for the assets, and includes them in the folder structure.
+This enables offline use of a catalog, by including all the actual data.
+
+Self-contained catalogs tend to be used more as static catalogs, where they can be easily passed around. But often they will
+be generated by a more dynamic STAC service, enabling a subset of a catalog or a portion of a search criteria to be downloaded
+and used in other contexts. That catalog could be used offline, or even published in another location.
+
+Self-contained catalogs are not just for offline use, however - they designed to be able to be published online and to live
```
fredliporace: typo, 'they *are* designed

catalog-spec/catalog-best-practices.md
```diff
@@ -0,0 +1,119 @@
...
+
+Many implementors are using static catalogs to be the reliable core of their dynamic services, or layering their STAC API
+on top of any static catalog that is published. These are some recommendations on how to handle this:
+
+#### Ingestion and links
+
+Implementors have found that it's best to 'ingest' a static STAC into an internal datastore (often elasticsearch, but a 
+traditional database could work fine too) and then generate the full STAC API responses from that internal representation.
+There were attempts to have the API refer directly to the static STAC Items, but maintaining the link structure in a 
+coherent way was difficult. So the recommendation is to create new absolute links, at least for the STAC search/ endpoint
```
fredliporace: That is the approach I'm using for CBERS, the items returned by the dynamic API are exactly the static items. Using absolute links for everything (absolute published catalog) makes that easy. Elasticsearch in my case index only the searchable fields and keeps a pointer to the item file location.

fredliporace:
> @fredliporace - review on all is appreciated, but do check out the recommendations on going static <-> dynamic in particular. I wasn't sure if you were going to stick with the direct references to the static catalog in the API responses - I remembered you said there were some issues, but not sure where you landed.

I ended using direct references to the static catalog, and using absolute URLs for everything in the static catalog. I found this more convenient if you already have a static catalog in place and then develop a dynamic API to allow searching. The main advantages are:
 - less space needed for datastore, only indexed properties are stored.
 - faster processing for ingesting new data. This is specially relevant when I need to re-index the complete catalog.

The only minor (IMO) drawback is that the self, parent, etc links point to a distinct URI from the URI that was used to access the dynamic catalog.

catalog-spec/catalog-best-practices.md
mojodna: Agreed, this needs more clarification, though `.../sample/sub1/3425.json` differs from `.../sample/sub1/3425/3425.json` in the depth.

Probably the former if assets are limited or external, the latter if there are many assets.

catalog-spec/catalog-best-practices.md
```diff
@@ -0,0 +1,119 @@
...
+A 'self-contained catalog' is one that is designed for portability. Users may want to download a catalog from online and be
+able to use it on their local computer, so all links need to be relative. Or a tool that creates catalogs may need to work 
+without knowing the final location that it will live at online, so it isn't possible to set absolute 'self' URL's. These use
+cases should utilize a catalog that follows the listed principles:
+
+* **Only relative href's in `links`**: The full catalog structure of links down to sub-catalogs and items, and their 
+links back to their parents and roots, should be done with relative URL's. This enables the full catalog to be downloaded or
+copy to another location and to still be valid. This also implies no `self` link, as that link must be absolute.
+
+* **Use Asset `href` links consistently**: The links to the actual assets are allowed to be either relative or absolute. There
```
mojodna: :+1:

catalog-spec/catalog-best-practices.md
```diff
@@ -0,0 +1,119 @@
...
+A 'published catalog' is one that lives online in a stable location, and uses `self` links to establish its location and 
+enable easy provenance tracking. There are two types of published catalogs:
+
+* **Absolute Published Catalog** is a catalog that uses absolute links for everything, both in the `links` objects and in the
+`asset` hrefs. It includes `self` links for every item. Generally these are implemented by dynamic catalogs, as it is quite
+easy for them to generate the proper links dynamically. But a static catalog that knows its published location could easily
+implement it.
+* **Relative Published Catalog** is a catalog that uses relative links for everything, but includes an absolute `self` link at
+the root catalog, to identify its online location. This is designed so that a self-contained catalog can be 'published' online
+by just adding one field (the self link) to its root catalog. All the other links should remain relative.
```
mojodna: Maybe a note about "resolution of item and sub-catalog `self` links may be done by traversing `parent` and `root` links, but requires reading multiple sources to achieve this."

catalog-spec/catalog-best-practices.md
```diff
@@ -0,0 +1,119 @@
...
+item should be treated as the canonical location, as the generated API is more likely to move or be temporarily down. The
+spec provides the `derived_from` rel attribute, which fits well enough, but `canonical` is likely the more appropriate one
+as everything but the links should be the same.
+
+#### Keeping static and dynamic catalogs in sync with cloud notification and queue services
+
+There is a set of emerging practices to use services like Amazon's Simple Queue Service (SQS) and Simple Notification Service
+(SNS) to keep catalogs in sync. There is a great [blog post on the CBERS STAC implementation on AWS](https://aws.amazon.com/blogs/publicsector/keeping-a-spatiotemporal-asset-catalog-stac-up-to-date-with-sns-sqs/). The core 
+idea is that a static catalog should emit a notification whenever it changes. The recommendation for SNS is to use message 
+body as the STAC item JSON itself, with some attributes such as a scene’s datetime and geographic bounding box that allows
```
mojodna: is to use the STAC Item JSON as the message body

mojodna:
> I cut out the href:from stuff for now. I think we should put the core idea in an extension, and then I'm happy to add it back in here as a best practice.

Since there's no prior art, I'm still on the fence about how best to do it, so 💯on removing it from this doc.

catalog-spec/catalog-best-practices.md
cholmes: Cool, will just remove the example.

catalog-spec/catalog-best-practices.md
cholmes: will fix

catalog-spec/catalog-best-practices.md
cholmes: will do

catalog-spec/catalog-best-practices.md
cholmes: cool, will adjust the language to show it's a valid route to go - wasn't sure where you ended up.

catalog-spec/catalog-best-practices.md
cholmes: thanks

cholmes:
Ok, I think this is ready for final review.

* Put links into the readme and the catalog spec as @m-mohr suggested
* fixed everything @mojodna and @fredliporace brought up (I think)
* refactored 'catalog types' into best practices, so the core spec is tighter, as the content fit more there.

catalog-spec/catalog-spec.md
```diff
@@ -207,7 +178,9 @@ with links.
 
 A more complete list of possible 'rel' types can be seen at the [IANA page of Link Relation Types](https://www.iana.org/assignments/link-relations/link-relations.xhtml).
 
-Please see the chapter 'relative vs absolute links' in the [Item spec](../item-spec/item-spec.md#relative-vs-absolute-links) for a discussion on that topic. 
+Please see the chapter 'relative vs absolute links' in the [Item spec](../item-spec/item-spec.md#relative-vs-absolute-links)
+ for a discussion on that topic, as well as the (use of links)[catalog-best-practices.md#use-of-links] section of the
```
kbgg: It looks like this link is broken

cholmes: fixed! Thanks...

cholmes:
I think everything is ready to go now - conflicts resolved and link fixed. Final review from @fredliporace or @m-mohr or @matthewhanson ?

Issue #384, opened 2019-02-04
Add `collection` property field


matthewhanson:
In the `commons` extension:
https://github.com/radiantearth/stac-spec/tree/master/extensions/commons

it mentions having a `collections` property containing the ID of the collection the `Item` belongs to, but this is not part of core, but probably should be.

In any catalog that has multiple collections, perhaps one of the most useful properties to search on would be the collection, because full featured searching is not available at the Collection's Items endpoint: /collections/<cid>/items/, and it would only appear under the Items links which is awkward and not available for searching.

Since Items are not strictly required to be part of a collection it should be optional, but if it is in a collection this field should be provided.

m-mohr:
I don't quite get why this needs to be moved to core and can't reside in the extension?

matthewhanson:
The reason is that the ability to search Items within a Collection has nothing to do with whether or not the `Commons` extension is used.  Collections are core, but being able to search for items within a collection with any of the `query` terms is currently not possible. You could be using no extensions at all, but you might have 50 Collections and no way to do a complex query on Items within one of them.

m-mohr:
Thanks. 

By the way, is a collection field (which is common across all items of a collection) put into the collection properties? Or does it reside in Items, which would somehow be a special behavior?

matthewhanson:
Well for sat-api there is a collection property in both the Collection and the Item, so I guess it is special behavior because it applies to all Items, but stays in the Items.  Based on implementation it might not have to be that way, but if it was not in the Item I think it might make search implementation harder.

On a related note, we recently changed the behavior of sat-api, but will open up new ticket about that.

mojodna:
When figuring out how to display collection information, I was asking myself whether the `collection` property duplicates the link with `rel=collection` or if it has an additional purpose...

The conclusion I came to was that it duplicates the link (with less information), so I was inclined to use the link to populate the UI (especially since it came with a URL to point to for navigation purposes) and ignore the `collection` property.

>  it would only appear under the Items links which is awkward and not available for searching.

This is when items are indexed as-is, correct? If some mild post-processing occurred (surfacing the link), this would be less problematic?

matthewhanson:
@mojodna There are two main problems:

1 - API spec currently doesn't allow for searching of non properties, so there's no way to search for items by collection if there's no collection property
2 - The collection ID doesn't appear in the link info, you have to follow the link to get the true ID of the collection, so this seems it would make searching by collection ID even harder.

[...]

PR #429, opened 2019-03-20
Add collection property to STAC Items (#384)

m-mohr:
As discussed in #384, this is the PR to add a `collection` property to STAC Items at the root-level. 

Having #415 in mind, do we specifically need to add a collection filter field to the API or does it just work out of the box?


matthewhanson:
I'm good with this change, but yes we do want to add a new search term for querying against collection. Want to add it this PR @m-mohr ?

matthewhanson:
Maybe add the change to make `commons` extension core this PR as well @m-mohr ?

m-mohr:
> yes we do want to add a new search term for querying against collection. Want to add it this PR @m-mohr ?

@matthewhanson Yes, should be included here. Directly belongs together, I think. I'm not so familiar with the search API parts yet. Could you help with it? I'm currently not sure where to put it.

> Maybe add the change to make `commons` extension core this PR as well @m-mohr ?

These are related changes, but can be done in separate PRs IMHO. I currently don't have the time to make the commons extension core, maybe will only do during the June sprint. So let's split them. It's also easier to review them separately, I guess.

item-spec/item-spec.md
```diff
@@ -36,15 +36,16 @@ incorporated.
...
+| properties | Properties Object                                                          | **REQUIRED.** A dictionary of additional metadata for the item. |
+| links      | [Link Object]                                                              | **REQUIRED.** List of link objects to resources and related URLs. A link with the `rel` set to `self` is required. |
+| assets     | Map<string, Asset Object>                                                  | **REQUIRED.** Dictionary of asset objects that can be downloaded, each with a unique key. Some pre-defined keys are listed in the chapter 'Asset types'. |
+| collection | string                                                                     | The `id` of the STAC Collection this Item references to (see `collection` relation type below). This field is *required* if such a relation type is present. This field provides an easy way for a user to search for any Items that belong in a specified Collection. |
```
kbgg: Here it says that this is a required field if the collection relation is present yet in the readme it states that this field is strongly recommended. These seem to be contradicting.

m-mohr: Thanks for spotting this, I'll check what we prefer.

matthewhanson:
I'm adding my review to this as I think it's fine to merge, but for 0.7.0 since this change goes in we will also need to update query to include a collections parameter.

@cholmes where we left it (IIRC) was that were would add new `id` and `collection` query terms and leave the `query` term for querying properties only so you wouldn't have to include "properties." for every term.

The `fields` term on the other hand was going to require "properties." for including any properties.

cholmes:
Ah, that's right @matthewhanson  I think there was general head nods on doing it for query too, but I think it's cool to hold off on that till next sprint.

Do we have `fields` term updated for that change? Or do we need an issue / PR for it too?

Will merge this PR.

item-spec/item-spec.md
m-mohr: If I get you correct, than you are referring to the collection rel type mentioned in the Item spec (strongly recommended) and to the collection field/property for items. It makes sense the way it is: It is strongly recommended to link to a collection. If a link to a collection is in the Item, it is required to add a collection field/property.

kbgg: Ah, you're right! Sorry about that.

PR #432, opened 2019-04-20
Cleanup API spec

matthewhanson:
Initial draft, still need to clean up and expand the documentation in README.

This PR:

- consolidates documentation into README.md  and search.md. There was a bunch of redundant or outdated information
- Improve process and documentation on creating OpenAPI from existing definitions, fragments, extensions

PR #440, opened 2019-04-26
Add page parameter

matthewhanson:
This adds the `page` parameter to the core STAC spec, alongside `limit` which was already there.

This PR already has the changes from #432 so that PR should be merged first.


m-mohr:
Waiting for #432 to be merged until I start reviewing this. The changes are just too many to find the page related ones.

cholmes:
@joshfix - would be great for you to review this one too, make sure it'll work with your implementation. Though this seems quite straightforward, so also fine if someone wants to merge it in.

joshfix:
Sorry for the delay; I see this is already merged but this looks good to me.  The very first spec used `limit` and a `next` string intended to hold a token that was some sort of stateful representation of the next page of results from a snapshot of results obtained at the time of the original search.  I've always used that `next` string field to hold what basically comes down to the `page` integer, so this change works perfectly for me.

PR #457, opened 2019-05-02
Updated UML diagram

hgs-msmith:
Updated UML diagram to reflect version 0.7.0


hgs-msmith:
This diagram is currently created using Visio, which is proprietary. The artifact is a VSD file.
I propose to migrate the diagram to draw.io, so that we can freely collaborate on its contents.
In the meantime, I can add the VSD file to the repo - do we want it at the root level, beside the PDF?

cholmes:
Sounds good on migrating to draw.io. I'd say let's just prioritize moving to that, and just stick with the pdf for now - no need to put in the .vsd file.

hgs-msmith:
I went ahead and converted the diagram to draw.io. However, I had to redraw it from scratch, so it will need a careful review for errors. Chris, do you still want to wait on this new version, or shall I merge with the current Visio one?

PR #464, opened 2019-05-05
heads-up: using reserved characters in IDs

fredliporace:
Issue originally raised [here](https://github.com/fredliporace/cbers-2-stac/issues/20)

Marked as WIP since this problem may also affect Item's ids and some other chars may also be forbidden.

One proposal to define the complete set of allowed chars would be an intersection of  [RFC 3986](https://tools.ietf.org/html/rfc3986#section-2) and allowd json characters, minus ``/``.

**PR Checklist:**


fredliporace:
@m-mohr my concern is the same raised in the [first comment](https://github.com/opengeospatial/WFS_FES/issues/156#issue-354007270) of the referenced issue.

I'm OK with escaping but I suggest at least mentioning it as a recommendation in the core spec since some implementations may start with a static catalog and later extend to support API.

I'll test with CBERS stac 0.7 implementation, will name one of the collections ``CBERS4%2FMUX`` and check if some tool I'm using breaks on the way.

fredliporace:
OK, just changed the PR to mention the requirement for percent encoding. Currently the PR mentions it in two places, one of them will be removed depending on the decision,

collection-spec/collection-spec.md
```diff
@@ -24,7 +24,7 @@ Implementations are encouraged, however, as good effort will be made to not chan
 | Element      | Type              | Description                                                  |
 | ------------ | ----------------- | ------------------------------------------------------------ |
 | stac_version | string            | **REQUIRED.** The STAC version the collection implements.    |
-| id           | string            | **REQUIRED.** Identifier for the collection that is unique across the provider. The ``/`` character is forbidden since the id value is used in some STAC API endpoints. |
+| id           | string            | **REQUIRED.** Identifier for the collection that is unique across the provider. Must be [percented encoded](https://tools.ietf.org/html/rfc3986#section-2) if a reserved character is used. |
 | title        | string            | A short descriptive one-line title for the collection.       |
```
fredliporace: This may be removed if the percent encoding is mentioned in api-spec.

m-mohr: We should put this somewhere differently. Percent encoding the ID in JSON doesn't make any sense. Before making any action, let's better figure out in which scenario this makes problems. 

For sure, you need to encode it in the generated links. So putting a description or link there would make sense. Similarly it could be added to the API. Are there other places where the links are not taken from the links section, but generated by clients / manually (i.e. not by the generating servers in the links section)?

fredliporace: Yes, I kept it there for discussion because that's what I understood @matthewhanson suggested on our last call. The PR currently has it also mentioned in the api-spec definition and one of them should go depending on the final decision.

I agree that mentioning it in the API spec is cleaner if we not restrict the allowed charset for ids. The only advantage I see in percent encoding the id is to warn the static stac implementer that some ids may generate awkward endpoints for dynamic APIs built over its content. 

That was my case, I would not choose CBERS4/MUX in the first place if I was aware of that point,
CBERS4-MUX works just fine.

Maybe we could just give a heads-up in the IDs definition that some chars would be escaped in some API endpoints.

collection-spec/collection-spec.md
```diff
@@ -24,7 +24,7 @@ Implementations are encouraged, however, as good effort will be made to not chan
 | Element      | Type              | Description                                                  |
 | ------------ | ----------------- | ------------------------------------------------------------ |
 | stac_version | string            | **REQUIRED.** The STAC version the collection implements.    |
-| id           | string            | **REQUIRED.** Identifier for the collection that is unique across the provider. |
+| id           | string            | **REQUIRED.** Identifier for the collection that is unique across the provider. The use of URI reserved characters such as `:` or `/` is discouraged since this will result in [percented encoded](https://tools.ietf.org/html/rfc3986#section-2) STAC endpoints. |
```
m-mohr: This should also be added to the catalog specification.

Additionally, in all three specs I'd write `STAC API` or `STAC API endpoints` instead of just `STAC endpoints`.

fredliporace: Done.

cholmes:
Discussed on call 8/5. On the call we also discussed the idea of evolving 'catalog best practices' to 'STAC best practices'. And that this goes into there.

We also want to be sure to tell clients to ingest, and add to API spec for how to encode id's. And then use recommendations for special characters.

m-mohr:
Will be tackled with PR #548, I think.

fredliporace:
> Will be tackled with PR #548, I think. Could you have a look there, @fredliporace?

*very* late reply just to confirm that PR #548 captures what I wrote here.

PR #467, opened 2019-05-25
minor: fixing query extension path in example

fredliporace:
I also changed the npm module call to work with locally installed modules, guess that will also work for globally installed modules and will make life easier for npm noobs like me.


api-spec/README.md
```diff
@@ -43,14 +43,14 @@ $ npm run generate-all
...
-!!files_merge_append ["STAC.yaml", "extensions/query.fragment.yaml"]
+!!files_merge_append ["STAC.yaml", "extensions/query/query.fragment.yaml"]
 ```
 
 Then, run the [yaml-files](https://www.npmjs.com/package/yaml-files) command line tool:
 
 ```
 $ npm install
-$ yaml-files myapi.merge.yaml myapi.yaml
+$ $(npm bin)/yaml-files myapi.merge.yaml myapi.yaml
```
m-mohr: I think the better way would be 
```
npm install -g
yaml-files myapi.merge.yaml myapi.yaml
```

fredliporace: -g option requires root level access which may be a hard constraint in some cases. There is also possible conflicts between already installed package versions.

m-mohr: But `$(npm bin)` is not working on Windows, which seems to also be a hard constraint... 
What I usually see in the JS community is -g, not $(npm bin), so I'd still favor the -g option or both as a compromise.

matthewhanson: Agree with @m-mohr .

@fredliporace can you update to leave in the -g option but then add text about having to use npm_bin if you don't have admin access?

fredliporace: @matthewhanson @m-mohr done.

PR #470, opened 2019-06-05
WIP: Add an extension for visualization parameters.

emsal1863:
The primary purpose of this extension is to allow collections to
specify the default way they are to be visualized or overlayed onto a
map.

**PR Checklist:**

### DIFF
--- a/extensions/visualization/README.md
+++ b/extensions/visualization/README.md
@@ -2,13 +2,14 @@
 
 ## Item Fields
 
-| Field Name                 | Type       | Description                                                  |
-|----------------------------|------------|--------------------------------------------------------------|
-| visualization:display_name | string     | Name of the displayed visualization.                         |
-| visualization:vis          | \*         | Parameters for visualizing the dataset as an image.          |
-| visualization:lookat       | LatLonZoom | Coordinate on which the visualization is initially centered. |
-
-<!-- We might want to add a filter field, which is a structure that translates into a query into the dataset of interest. -->
+| Field Name                    | Type       | Description                                                                                     |
+|-------------------------------|------------|-------------------------------------------------------------------------------------------------|
+| vis:display_name              | string     | Name of the displayed visualization.                                                            |
+| vis:vis                       | \*         | Parameters for visualizing the dataset as an image.                                             |
+| vis:lookat                    | LatLonZoom | Coordinate on which the visualization is initially centered.                                    |
+| vis:filter                    | Filter     | Optional structure translating to a query into the dataset.                                     |
+| vis:preview_image_script_name | string     | Optional URL linking to the JavaScript script used to create the sample/thumbnail image.        |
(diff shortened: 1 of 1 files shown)


extensions/visualization/README.md
```diff
@@ -0,0 +1,76 @@
...
+| property_name | string  | The property to visualize.                               |
+| property_vis  | BandVis | Parameters for the visualization of the resulting image. |
+
+BandVis is defined as follows:
+
+## Band Visualization
+
+| Field Name | Type     | Description                                                                                    |
+|------------|----------|------------------------------------------------------------------------------------------------|
+| min        | [double] | Minimum value of each band. If only one value then it is applied as the minimum for all bands. |
```
jisantuc: Is there some way to express optionality? For example, in Raster Foundry land, we consider these [optional](https://github.com/raster-foundry/raster-foundry/blob/37754fc633fc3efdf8e8b6d477b75313cd812c7a/app-backend/backsplash-core/src/main/scala/com/rasterfoundry/backsplash/color/ColorCorrect.scala#L26-L27), since we assume for color correction that we have statistical information about the image, but that a user might want to override it. Does that match the intention here?

extensions/visualization/README.md
```diff
@@ -0,0 +1,76 @@
...
+| vis:earth_engine_script_name  | string     | Optional URL linking to the JavaScript script file used for the Earth Engine site code snippet. |
+
+\* = One of the Image Visualization, Table Visualization, or Polygon Visualization, defined below:
+
+### Image Visualization
+
+| Field Name | Type    | Description                                            |
+|------------|---------|--------------------------------------------------------|
+| band_vis   | BandVis | Vis args for the given band_names.                     |
+| global_vis | BandVis | If vis_args are present here, they apply to all bands. |
```
jisantuc: It looks like there are two places where you can specify global viz arguments -- in `band_vis`, if you provide only one value in the array parameters, or here -- am I reading this correctly that those could conflict? If so, does `global_vis` take precedence? Under what circumstances should someone provide both `global_vis` and `band_vis`?

extensions/visualization/README.md
m-mohr: Everything that is not required explicitly is optional.

jisantuc: I mean optional in the "value or `null`" sense on the individual array items. For example, if I had some bands where I want to set a minimum, but not others, I don't believe I can express that here (unless I've gotten too used to types, and `[0, null, 0]` would be a valid value here)

extensions/visualization/README.md
emsal1863: In Earth Engine it's handled like this:
In `ImageVisualization` a number of `band_names` are specified. For each of these bands, the BandVis in `band_vis` is applied. But for ALL bands of the Image/ImageCollection, the visualization parameters in `global_vis` are applied.

matthewhanson:
Stale PR - will be taken up for consideration after 1.0 if there is still interest

PR #474, opened 2019-06-05
Move Single Item Extension to core

mojodna:
Per in-person discussion.

Resolution rules: use the `license` and/or `providers` "closest" to the definition of an item being resolved. I.e. if one is in the Item's `properties`, use that. If not, check Commons properties in the parent Collection's `properties`. If not, use the parent Collection's top-level `license` and/or `providers`.

This also adds `title` to Item's JSON schema (it was missing).


item-spec/examples/sample-full.json
```diff
@@ -16,6 +16,18 @@
   },
   "properties": {
     "datetime": "2016-05-03T13:22:30.040Z",
+    "title": "A CS3 item",
+    "license": "proprietary",
```
m-mohr: If license is proprietary there should be a link to the license...

item-spec/item-spec.md
```diff
@@ -65,6 +65,8 @@ custom fields.
 | Field Name | Type   | Description                                                  |
 | ---------- | ------ | ------------------------------------------------------------ |
 | datetime   | string | **REQUIRED.** The searchable date and time of the assets, in UTC. It is formatted according to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). |
+| license    | string | Item's license(s) as a SPDX [License identifier](https://spdx.org/licenses/) or [expression](https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60) or `proprietary` if the license is not on the SPDX license list. Proprietary licensed data SHOULD add a link to the license text, see the `license` relation type. |
```
m-mohr: We agreed that we want to add text here that it is preferable to have this information in the collection and not in an item.

item-spec/item-spec.md
```diff
@@ -74,6 +76,32 @@ data, so use whatever single date and time is most useful for a user to search f
...
+| url         | string    | Homepage on which the provider describes the dataset and publishes contact information. |
+
+**roles**: The provider's role(s) can be one or more of the following elements:
+
+* *licensor*: The organization that is licensing the dataset under the license specified in the collection's `license` field.
+* *producer*: The producer of the data is the provider that initially captured and processed the source data, e.g. ESA for Sentinel-2 data.
+* *processor*: A processor is any provider who processed data to a derived product.
+* *host*: The host is the actual provider offering the data on their storage. There should be no more than one host, specified as last element of the list. 
+
+## Relation types
```
m-mohr: This chapter should be merged with the other relation types chapter in this file.

item-spec/json-schema/item.json
```diff
@@ -67,6 +67,48 @@
...
+                "title": {
+                  "title": "Item Title",
+                  "description": "A human-readable title describing the item.",
+                  "type": "string"
+                },
+                "license": {
+                  "title": "Item Licenses",
+                  "type": "string"
+                },
+                "providers": {
```
m-mohr: Would it make sense to import these definitions from a central place (e.g. collection JSON schema?)

item-spec/examples/sample-full.json
mojodna: Good call. I'll change it to something known then.

item-spec/json-schema/item.json
mojodna: Yes... I'll see what I can do about this outside this PR.

PR #476, opened 2019-06-05
Redefine intersects parameter to accept GeoJSON geometry values (473)

joshfix:


api-spec/api-spec.md
```diff
@@ -29,7 +29,7 @@ Items in the collection should match all filters to be returned when querying. T
 | ------------ | ------------- | ---------------------- |
 | bbox | [number]       | Requested bounding box [west, south, east, north] |
 | time | string | Single date, date+time, or a range ('/' seperator), formatted to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6) |
-| intersects | GeoJSON Feature | Searches items by performing intersection between their geometry and provided GeoJSON Feature |
+| intersects | GeoJSON Geometry | Searches items by performing intersection between their geometry and provided GeoJSON GeoJSON |
```
philvarner: typo: "GeoJSON GeoJSON" -> "GeoJSON Geometry"

philvarner: Also, I think we should state here explicitly which Geometry types **must** be supported.  Or, maybe that gets into capabilities? We discussed that `GeometryCollection` should not be supported, but I'm inclined to want to support that, as Elasticsearch does.  Of the others ("Point", "MultiPoint", "LineString", "MultiLineString", "Polygon", "MultiPolygon"), what if there's a case where a backend doesn't support one of those? Does that then mean that you cannot be api compliant?  Maybe we say it is **required** to support Polygon, and **recommended** to support the other 5, and GeometryCollection must return an error that it is not supported.

m-mohr: Well, if you can handle GeometryCollection, not sure why it should throw an error. Other than that, fine with your proposal.

matthewhanson: I'm fine with making it required to support all geometry types, including GeometryCollection. I think any geo backend should and will be able to support any type of geometry, more or less out of the box.

Otherwise, we have to make different types of geometry as extensions, which I'd rather not do at this point.

If we reach a point where someone is having a difficult time supporting all geometries and wants to be fully compliant we can talk about relaxing the requirements.

philvarner: all Geometry types sounds great to me.

cholmes:
Merging in, thanks for everyone's work!

PR #477, opened 2019-06-06
WIP: add native properties

philvarner:


m-mohr:
Why's this in the EO extension? That's nothing specific to electro-optical data and could also be useful for other type of data, right? So wondering whether that should be a separate extension?

philvarner:
> Why's this in the EO extension? That's nothing specific to electro-optical data and could also be useful for other type of data, right? So wondering whether that should be a separate extension?

That's a good point. I think I just put it there because I look at the EO extension most frequently, so I was most familiar with it. I'll gladly put this in another extension -- maybe along the lines of "source metadata"? Matt Eldridge was also keen on getting other metadata like the header information needed to do range reads into the catalog, so tools can operate on them without needing to crack the file open first. I've also had the need for the gdalinfo output from the assets - while that's not a standard, it's very useful in a lot of situations

m-mohr:
+1 in having this in a separate extension.

I have not enough information or experience with the other points you brought up, so neutral on that.

PR #481, opened 2019-06-06
Add suggestions to training data/label extension

aaronxsu:
This PR suggests some changes based on the training data PR. It also adds some linked examples under the implementations section as a result of the proposed changes.

This PR is a result of discussions among @daveluo, @nrweir, @philvarner, @mteldridge, and yours truly during STAC Sprint 4 in San Francisco, CA,  June 4th to 6th, 2019.


matthewhanson:
Briefly looked at this, but going to go ahead and merge it into the #362 where we can discuss from there.
I'll give it a more thorough review this upcoming week.

PR #482, opened 2019-06-06
STAC API Search Metadata Extension

philvarner:


api-spec/api-spec.md
```diff
@@ -27,22 +27,24 @@ Items in the collection should match all filters to be returned when querying. T
 
 | Parameter      | Type          | Description        |
 | ------------ | ------------- | ---------------------- |
-| bbox | [number]       | Requested bounding box [west, south, east, north] |
-| time | string | Single date, date+time, or a range ('/' seperator), formatted to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6) |
+| ids | [string] | Array of Item ids to return. All other filter parameters that further restrict the number of search results (except `next` and `limit`) are ignored |
```
philvarner: just reorganizing these to the top

api-spec/api-spec.md
```diff
@@ -27,22 +27,24 @@ Items in the collection should match all filters to be returned when querying. T
...
+| collections | [string] | Array of Collection IDs to include in the search for items. Only Items in one of the provided Collections will be searched |
 | intersects | GeoJSON Feature | Searches items by performing intersection between their geometry and provided GeoJSON Feature |
-| page | number | The page number of results. Defaults to 1 |
+| bbox | [number]       | Requested bounding box [west, south, east, north] |
```
philvarner: now bbox and intersects are adjacent

api-spec/api-spec.md
```diff
@@ -27,22 +27,24 @@ Items in the collection should match all filters to be returned when querying. T
...
+| time | string | Single date, date+time, or a range ('/' separator), formatted to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6) |
+| next | string | The token to retrieve the next set of results, e.g., offset, page, continuation token|
```
philvarner: page -> next

extensions/search/schema.json
```diff
@@ -0,0 +1,45 @@
+{
```
philvarner: I honestly have no idea if this is the right way to express this in JSON Schema

extensions/README.md
```diff
@@ -53,6 +53,7 @@ An extension can add new fields to STAC entities (content extension), or can add
 | [EO](eo/README.md) (`eo`)                                    | Item             | Covers electro-optical data that represents a snapshot of the earth for a single date and time. It could consist of multiple spectral bands, for example visible bands, infrared bands, red edge bands and panchromatic bands. The extension provides common fields like bands, cloud cover, off nadir, sun angle + elevation, gsd and more. | *Pilot* |
 | [Point Cloud](pointcloud/README.md) (`pc`)                   | Item             | Provides a way to describe point cloud datasets. The point clouds can come from either active or passive sensors, and data is frequently acquired using tools such as LiDAR or coincidence-matched imagery. | *Proposal* |
 | [SAR](sar/README.md) (`sar`)                                 | Item             | Covers synthetic-aperture radar data that represents a snapshot of the earth for a single date and time. | *Proposal* |
+| [Search](search/README.md) | ItemCollection | Adds search-related metadata to ItemCollection. | *Proposal* |
```
m-mohr: What is an ItemCollection? Is this extending the Item or the Collection spec or both?

extensions/search/schema.json
```diff
@@ -0,0 +1,45 @@
...
+          "required": [
+            "next",
+            "returned"
+          ],
+          "properties": {
+            "next": {
+              "type": ["string", "null"]
+            },
+            "limit": {
+              "type": ["number", "null"]
```
m-mohr: I guess number should be integer, or are we allowing floats? ;-)

extensions/README.md
m-mohr: Or is this an API extension? In this case this is not the correct list of extensions...

api-spec/api-spec.md
```diff
@@ -27,22 +27,24 @@ Items in the collection should match all filters to be returned when querying. T
...
 
 STAC provides some additional endpoints for the root Catalog itself, as well as the capability to search the Catalog. Note that a STAC API does not need to implement WFS 3, in which case it would only support the endpoints given below. See the [OpenAPI specification document](definitions/STAC-standalone.yaml).
 
 | Endpoint      | Returns          | Description        |
-| ------------ | ------------- | ---------------------- |
-| /stac | Catalog        | Root catalog |
-| /stac/search | Items | GeoJSON FeatureCollection of Items found |
+| ------------- | ------------- | ---------------------- |
+| /stac         | Catalog        | Root catalog |
+| /stac/search  | ItemCollection   | GeoJSON FeatureCollection, probably containing additional STAC metadata, including search metadata from the `search` extension |
```
m-mohr: Why is it called ItemCollection if it returns a GeoJSON FeatureCollection?

extensions/README.md
joshfix: @m-mohr going back to the very first sprint, we originally defined the basic Item structure as a geojson "compatible" structure, where it was not a "Feature" but an "Item".  And all of the items belonged to an "ItemCollection".  However, we wanted to be friendly to clients, so we stuck to using "Feature" and "FeatureCollection" as the type.  If you look at `[/api-spec/STAC.yaml](https://github.com/radiantearth/stac-spec/blob/master/api-spec/STAC.yaml)', you'll see on line 808 that we define an `ItemCollection` with the same definition as a `FeatureCollection`:

```
    itemCollection:
      type: object
      required:
        - features
        - type
      properties:
        type:
          type: string
          enum:
            - FeatureCollection
```

Basically the exact same way we have `items` in our catalogs/collections that all have `"type": "Feature"`.  Kind of an odd nuance, but those are the historical semantics.

extensions/search/schema.json
joshfix: +1 should be `integer`

api-spec/api-spec.md
joshfix: See my comment above.  STAC defines "Item" and "ItemCollection" that are typed "Feature" and "FeatureCollection" to provide closer alignment with GeoJSON, however we are not technically 100% GeoJSON.

extensions/README.md
m-mohr: Fair enough. It's just confusing for new users as this is not defined outside of schemas and there's no ItemCollection spec. I think this should be improved, either reporting it as GeoJSON FeatureCollection or have an ItemCollection spec somewhere and link to it.

api-spec/api-spec.md
m-mohr: I'd say we are a superset of GeoJSON. 100% compatible, with additions.

extensions/README.md
joshfix: Totally agree.  It's confusing enough that I'm pretty sure I was the only person to even remember that this is how we defined it.

api-spec/api-spec.md
philvarner: I found a clear answer in the GeoJSON spec (hint: I actually read it) [https://tools.ietf.org/html/rfc7946#section-6] The term is "foreign members".  I think the correct description is "an Item is a GeoJSON Feature augmented with foreign members that contain the various values relevant to a STAC entity" -- likewise for ItemCollection/FeatureCollection.

philvarner: Looks like the only place we talk about ItemCollection is in the STAC openapi definition.  We should probably either add it into the existing Item spec, or add another spec on par with Item as we've done with Catalog and Collection -- I lean towards the latter. I think we need to be *very* clear when describing these to avoid confusing between Collection and ItemCollection.

extensions/README.md
philvarner: Filed https://github.com/radiantearth/stac-spec/issues/484 -- I'd gladly take this on, but I don't have permissions to assign it to myself.

extensions/search/schema.json
philvarner: yes, blame my ignorance with json schema, even though my IDE has autocomplete for these

collection-spec/README.md
```diff
@@ -1,4 +1,4 @@
-# STAC Collections
+# STAC Collection Specification
```
philvarner: renamed title to be consistent with other specs

api-spec/api-spec.md
m-mohr: As said in #484: I'm not sure whether I would add this on the same level as Item / Collection / Catalog as those are all used in both static and dynamic catalogs, but ItemCollection is only used in API. Therefore, I'd be in favor of giving it a prominent place in the API spec - or is ItemCollection also getting an equivalent in static catalogs?

I like the statement on foreign members. Make sure to link "foreign members" to the GeoJSON spec!

collection-spec/README.md
m-mohr: That's not correct. It's "Stac Items", "Stac Collections" and "STAC Catalogs" in RADME files. Only the ...-spec files and extensions have the "Specification" suffix. So I'd be in favour to revert this.

philvarner: > That's not correct. It's "Stac Items", "Stac Collections" and "STAC Catalogs" in README files. Only the ...-spec files and extensions have the "Specification" suffix. So I'd be in favour to revert this.

You're right, my mistake. reverted.

api-spec/api-spec.md
philvarner: Added a description of ItemCollection and foreign members language to both Item and ItemCollection in my latest commit.

api-spec/STAC.yaml
```diff
@@ -928,14 +928,10 @@ components:
...
-        An array of links. Can be used for pagination, e.g. by providing a link
-        with the `next` relation type.
+        An array of links to resources related to this collection.
       items:
         $ref: '#/components/schemas/link'
-      example:
-        - rel: next
-          href: >-
-            http://api.cool-sat.com/query/gasd312fsaeg/ANsXtp9mrqN0yrKWhf-y2PUpHRLQb1GT-mtxNcXou8TwkXhi1Jbk
+
```
philvarner: removed this, as it is now a confusing example with the `next` parameter and search:meta field.  We should replace it at some point with a good example.

matthewhanson:
So this is really just the "search:meta" fields that are the extension right?

In which case the "itemCollection" references will be resolved by completing #484 and then can be referenced here.

So we need a new PR for #484 first.

api-spec/STAC.yaml
m-mohr: Can you come up with a good example? I'm always concerned that we forget to add it in again if we just remove it without creating an issue or so...

api-spec/api-spec.md
```diff
@@ -1,12 +1,47 @@
...
 
 In WFS 3 Collections are the sets of data that can be queried ([7.11](https://rawgit.com/opengeospatial/WFS_FES/master/docs/17-069.html#_feature_collections_metadata)), and each describes basic information about the geospatial dataset, like its name and description, as well as the spatial and temporal extents of all the data contained. [STAC collections](../collections-spec/README.md) contain this same information, along with other STAC specific fields and thus are compliant with both WFS Collections and STAC Collections and is returned from the `/collections/{collection_id}` endpoint.
 
 In WFS 3 Features are the individual records within a Collection and are provided in GeoJSON format. [STAC Items](../item-spec/README.md) are analagous to WFS 3 Features, are in GeoJSON, and are returned from the `/collections/{collection_id}/items/{item_id}` endpoint.
 
-### WFS3 Endpoints
+## STAC ItemCollection Specification
```
m-mohr: Would it make more sense to put it at the end of the page where it is used?
At this place it feels a bit lost without prior reference.

item-spec/item-spec.md
```diff
@@ -1,8 +1,9 @@
 # STAC Item Specification
 
-This document explains the structure and content of a SpatioTemporal Asset Catalog (STAC) Item. Each
-is a [GeoJSON](http://geojson.org/) [feature](https://tools.ietf.org/html/rfc7946#section-3.2), plus
-a few required fields that identify the time range and assets of the item. An Item is the core
+This document explains the structure and content of a SpatioTemporal Asset Catalog (STAC) Item. An **Item** is a
```
m-mohr: I like this clarification a lot!

api-spec/STAC.yaml
joshfix: I think something like `http://api.cool-sat.com/stac/search?next=ANsXtp9mrqN0yrKWhf-y2PUpHRLQb1GT-mtxNcXou8TwkXhi1Jbk` should work.

philvarner: added.

api-spec/api-spec.md
philvarner: I've removed "## STAC ItemCollection Specification" from the latest revision -- they're in another PR now.

m-mohr: Unfortunate...

extensions/search/schema.json
```diff
@@ -0,0 +1,45 @@
...
+            },
+            "found": {
+              "type": "number"
+            },
+            "returned": {
```
m-mohr: What is returned useful for? Isn't that just a `Array.size(features)` (or however array length is determined in your programming language of choice)?

extensions/search/schema.json
```diff
@@ -0,0 +1,45 @@
...
+              "type": "number"
+            },
+            "properties": {
```
m-mohr: It feels like there are too many `properties` properties nested here. Maybe rename it to `additional` or so?

extensions/search/schema.json
philvarner: Correct.  The general consensus during discussions at Sprint 4 was to include this, even though it's something that can be computed.

extensions/search/schema.json
philvarner: i just removed it all together, since you throw whatever you want into the object anyway.

extensions/search/schema.json
m-mohr: I can take over the JSON Schema and fix it if you like me to do that.

extensions/search/schema.json
m-mohr: Okay, fine with that. It's just an extension, just trying to not clutter the spec too much. Do you remember the reason for it?

extensions/search/schema.json
philvarner: I added you as a collaborator, or you can just send me the corrected schema.

m-mohr: Thanks, updated.

extensions/search/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+| Element      | Type          | Description                                                  |
+| ------------ | ------------- | ------------------------------------------------------------ |
+| `search:metadata` | SearchMetadata Object | **REQUIRED.** The search-related metadata for the [ItemCollection](../itemcollection-spec/README.md). |
+
+### SearchMetadata Object
+
+| Element      | Type          | Description                                                  |
+| ------------ | ------------- | ------------------------------------------------------------ |
+| next         | string\|null  | **REQUIRED.** The value to set for the `next` query parameter in order to get the next page of results. |
```
m-mohr: It is not documented what `null` is about.

extensions/search/README.md
```diff
@@ -0,0 +1,26 @@
...
+| returned     | integer       | **REQUIRED.** Count of results returned by this response. Equal to the cardinality of the features array. |
+| limit        | integer\|null | Max number of results requested explicitly or the default limit used by the server if no parameter was provided. `null` if no limit should be placed on the results, but servers may still limit the results nevertheless. |
```
m-mohr: I tried to clarify this a bit as it seemed not very well described yet. Realized that when doing the JSON schemas. As it is not required, what is the default value? Is it set by the server implementation? Also added the note that servers may still limit the results for `null`.

extensions/search/README.md
m-mohr: Also, should we also allow integers here?

extensions/README.md
```diff
@@ -53,6 +53,7 @@ An extension can add new fields to STAC entities (content extension), or can add
 | [EO](eo/README.md) (`eo`)                                    | Item             | Covers electro-optical data that represents a snapshot of the earth for a single date and time. It could consist of multiple spectral bands, for example visible bands, infrared bands, red edge bands and panchromatic bands. The extension provides common fields like bands, cloud cover, off nadir, sun angle + elevation, gsd and more. | *Pilot* |
 | [Point Cloud](pointcloud/README.md) (`pc`)                   | Item             | Provides a way to describe point cloud datasets. The point clouds can come from either active or passive sensors, and data is frequently acquired using tools such as LiDAR or coincidence-matched imagery. | *Proposal* |
 | [SAR](sar/README.md) (`sar`)                                 | Item             | Covers synthetic-aperture radar data that represents a snapshot of the earth for a single date and time. | *Proposal* |
+| [Search](search/README.md) | [ItemCollection](../itemcollection-spec/README.md) | Adds search-related metadata to [ItemCollection](../itemcollection-spec/README.md). | *Proposal* |
```
m-mohr: Shouldn't this extension be an API extension in the corresponding folder (api-spec/extensions)?

extensions/search/README.md
philvarner: I'm going to change the language a bit.  This is for reporting back to the user what the `limit` value was if it existed (and null if there was no limit at all).  A few cases are:
* No value in the API call: limit = server default
* value in the API call larger than max limit:  limit = server max limit
* server has the (poor) behavior of no limit: limit = null

extensions/README.md
philvarner: Perhaps it should? This is mostly my ignorance of how the spec is organized.  I can move it.

extensions/search/README.md
philvarner: I dislike multi-typed attributes.  In practice, this is representing a query string parameter, so it's a string, even if it's a string of digits like a `page`/`offset` value would be.

extensions/README.md
m-mohr: Ich think it should. @matthewhanson?

extensions/search/README.md
m-mohr: Oh, my bad. I didn't realize that this is the response when writing the comment. Sorry.

extensions/search/README.md
m-mohr: True. It's the response...

extensions/search/README.md
m-mohr: But if it's the response, a limit could always be delivered so why not make it required? The server should always know what its limit is and if not: null.

joshfix: I tend to agree.  I would even favor returning something like -1 if there were some circumstance where the server was incapable of knowing what limit was used on the back-end.  Returning null values just feels icky.

m-mohr: Hah, I always feel bad when -1 is used and favor null for it. ;-)

joshfix: Lol... agree to disagree!  :)  I'm not going to fight for -1 vs null though.  If we want it to be null, then I guess we can all just do null checks!

philvarner: I put it this way because that's what it seemed other people wanted. I actually prefer the semantics where you just don't include the field if it would be null, so if next is not there, there are no more pages, and if limit is not there, there was no limit.

m-mohr: For the openEO API we realized that it seems to be best to allow both: null and leaving it out. Both should have the same semantics. Non-compliant implementations often leave fields out and compliant back-ends often send null as it's sometimes easier to just set to null instead of removing a field again (at least where you have a relational database in the background, not so much in this case). I'd try to avoid different semantics for null and not set.

extensions/README.md
philvarner: I just moved it into the API spec directory, as it's entirely related to the API.

philvarner:
I believe all concerns around this PR have been resolved. Only remaining question is whether we merge #484 first so that the links to Item Collection are correct.

api-spec/extensions/README.md
```diff
@@ -11,8 +11,9 @@ API Extensions given follow the same guidelines for Extension Maturity as given
 
 | Extension Name | Description | Maturity |
 | -------------  | ----------- | -------- |
-| [Fields](fields/README.md) | Adds parameter to constrol which fields are returned in the response. | *Pilot* |
+| [Fields](fields/README.md) | Adds parameter to control which fields are returned in the response. | *Pilot* |
 | [Query](query/README.md) | Adds parameter to search Item and Collection properties. | *Pilot* |
+| [Search](search/README.md) | [ItemCollection](../item-spec/README.md) | Adds search-related metadata to [ItemCollection](../item-spec/README.md). | *Proposal* |
```
m-mohr: Could you link directly to the ItemCollection section?

api-spec/api-spec.md
```diff
@@ -27,22 +27,24 @@ Items in the collection should match all filters to be returned when querying. T
...
 
 STAC provides some additional endpoints for the root Catalog itself, as well as the capability to search the Catalog. Note that a STAC API does not need to implement WFS 3, in which case it would only support the endpoints given below. See the [OpenAPI specification document](definitions/STAC-standalone.yaml).
 
 | Endpoint      | Returns          | Description        |
-| ------------ | ------------- | ---------------------- |
-| /stac | Catalog        | Root catalog |
-| /stac/search | Items | GeoJSON FeatureCollection of Items found |
+| ------------- | ------------- | ---------------------- |
+| /stac         | Catalog        | Root catalog entry point |
+| /stac/search  | [ItemCollection](../query_type/README.md) | Retrieves a group of Items matching the provided search predicates, probably containing search metadata from the `search` extension |
```
m-mohr: Where does this point to? Should it be the item spec?

philvarner: This seems to be a mis-paste. fixed.

api-spec/extensions/README.md
philvarner: fixed the link and removed the extra column that came from the table in the core extensions that i copied this from.

philvarner:
Fixed issues raised by @m-mohr

api-spec/api-spec.md
```diff
@@ -27,22 +27,24 @@ Items in the collection should match all filters to be returned when querying. T
...
 | intersects | GeoJSON Geometry | Searches items by performing intersection between their geometry and provided GeoJSON geometry.  All GeoJSON geometry types must be supported. |
-| page | number | The page number of results. Defaults to 1 |
+| bbox | [number]       | Requested bounding box [west, south, east, north] |
+| time | string | Single date, date+time, or a range ('/' separator), formatted to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6) |
+| next | string | The token to retrieve the next set of results, e.g., offset, page, continuation token|
 | limit | number | The maximum number of results to return (page size). Defaults to 10 |
-| ids | [string] | Array of Item ids to return. All other filter parameters that further restrict the number of search results (except `page` and `limit`) are ignored |
-| collections | [string] | Array of Collection IDs to include in the search for items. Only Items in one of the provided Collections will be searched |
+
+In general, only one of **intersects** or **bbox** should be specified.  If both are specified, results should match both.
```
metasim: Would it be better to make them mutually exclusive? The **bbox** could be added to the **intersects** as geometry, allowing the query engine uniformly handle the geospatial relation in one pass.

api-spec/extensions/search/README.md
```diff
@@ -0,0 +1,45 @@
...
+| limit        | integer \| null | The maximum number of results to which the result was limited |
+| found        | integer         | The count of total number of results for this query, possibly estimated, particularly in the context of NoSQL data stores |
+
+**next** - The value to set for the `next` query parameter in order to get the next page of results.  This will typically be something like the "page" parameter in many APIs, "offset" parameter in a SQL query, or "searchAfter" parameter in an Elasticsearch query.  The value `null` indicates that there are no more results for which to query. 
+
+The ability to implement meaningful semantics for the `next` query parameter and `next` SearchMeta field is related to the use of a stable sort over query result items.  The default sort of query results should be stable, but may not be depending on the data store's sorting performance.  It is recommended that the [Sort API Extension](../sort/README.md) be implemented in conjunction with this extension and that fields conducive to stable sorting have sorting enabled over them.
```
metasim: Should stable order (even an arbitrary one) be a _requirement_? It could be difficult to implement and test unstable ordering and paging at the same time.

philvarner: I think in practice most apis will implement stable sorting, but I've been reluctant to force users implement specific behaviors other than those that would break interoperability.

api-spec/api-spec.md
philvarner: I think that's a good point.  This change was actually part of the API documentation PR that i picked into this PR, but I was the one who wrote.  In my own implementation, I preclude users from specifying both, and I think that's what the behavior should be here.  I'll create another PR for that.

PR #483, opened 2019-06-07
Rework API examples page

joshfix:
The API examples page primarily focused on dynamic queries against the `/collections/{collection_id}/items` endpoint, which is limited to only the `limit`, `time`, and `bbox` parameters.  Because of the limitations, we should be driving STAC users who wish to perform dynamic queries to the `/stac/search` endpoint.

In this PR, the examples have been reworked to show examples against the `/stac/search` endpoint.  A single example has been left to show how to query the collections endpoint and outline that the collections endpoints does not support HTTP POST.

### DIFF
--- a/api-spec/examples.md
+++ b/api-spec/examples.md
@@ -8,49 +8,49 @@ parameters (such as intersect)
 
 ## Core examples
 
-Requests all the data in the collection that is in New Zealand:
+Requests all the items in the collection that is in New Zealand:
 
 ```
 GET /stac/search?bbox=160.6,-55.95,-170,-25.89
 ```
 
-Request all the data in `mycollection` that is in New Zealand:
+Request all the items in `mycollection` that is in New Zealand:
 
 ```
 GET /stac/search/items?bbox=160.6,-55.95,-170,-25.89&collections=mycollection
 ```
 
-Request 100 results in `mycollection` from New Zealand:
+Request 100 items in `mycollection` from New Zealand:
 
 ```
 GET /stac/search?bbox=160.6,-55.95,-170,-25.89&limit=100&collections=mycollection
 ```
 
-Request all the data in `mycollection` that is in New Zealand from January 1st, 2019:
+Request all the items in `mycollection` that is in New Zealand from January 1st, 2019:
 
 ```
 GET /stac/search?bbox=160.6,-55.95,-170,-25.89&time=2019-01-01&collections=mycollection
 ```
 
-Request 100 results in `mycollection` or `yourcollection` that is in New Zealand from January 1st, 2019:
+Request 100 items in `mycollection` or `yourcollection` that is in New Zealand from January 1st, 2019:
 
 ```
 GET /stac/search?bbox=160.6,-55.95,-170,-25.89&time=2019-01-01&limit=100&collections=mycollection,yourcollection
 ```
 
-Request all the data in any collection that is in New Zealand from January 1st, 2019:
+Request all the items in any collection that is in New Zealand from January 1st, 2019:
 
 ```
 GET /stac/search?bbox=160.6,-55.95,-170,-25.89&time=2019-01-01
 ```
 
-Request all the data in any collection that is in New Zealand and has the id `1` or `2`:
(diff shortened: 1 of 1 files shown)


api-spec/examples.md
```diff
@@ -3,53 +3,84 @@
 A STAC API supports both GET and POST requests. It is best to use POST when using the intersects query for two reasons:
 
 1. the size limit for a GET request is less than that of POST, so if a complex geometry is used GET may fail.
-2. The parameters for a GET request must be escaped properly, making it more difficult to construct when using JSON parameters (such as intersect)
+2. The parameters for a GET request must be escaped properly, making it more difficult to construct when using JSON 
+parameters (such as intersect)
 
 ## Core examples
 
+Requests all the data in the collection that is in New Zealand:
```
philvarner: maybe call this explicitly `Items` rather than data -- since technically all STAC entities are metadata

api-spec/examples.md
```diff
@@ -3,53 +3,84 @@
...
 ```
-GET /collections/mycollection/items?time=2019-01-01/2019-04-01
+GET /stac/search?bbox=160.6,-55.95,-170,-25.89&time=2019-01-01&limit=100&collections=mycollection,yourcollection&ids=1,2
+```
+
+```
+POST /stac/search
+```
+
+Body:
```
philvarner: I'd prefer if we required this to be `multipart/form-data` instead of a JSON query body.

api-spec/examples.md
```diff
@@ -91,11 +122,30 @@ If time is a period without a start or end date, the end date is assigned to "no
 
 `P1Y2M10DT2H30M` is the same as `"P1Y2M10DT2H30M/" + new Date().toISOString()`
 
+## Querying the collections endpoint
+
+The collections endpoint does support the following query parameters to remain in alignment with WFS3:  `time`, `bbox`, 
+and `limit`.  Due to the limited parameter support, it is recommended to use the `/stac/search` endpoint for dynamic 
+STAC queries.  Some STAC implementations may choose to provide support for the full superset of STAC query parameters
```
philvarner: In fewer words:
```Implementations may **optionally** provide support for the full superset of query parameters 
to the `/collections/{collectionId}` endpoint, where the collection ID in the path is equivalent to providing that single value in the `collections` query parameter.```

api-spec/examples.md
m-mohr: Why?

philvarner: I prefer that because all of the values passed to the parameters are exactly the same in the GET and POST cases.  Otherwise, you have to define a separate query payload JSON schema, with similar (but not exact) mappings between values, for example, that `ids` is an array of strings rather than a comma-separated list of strings. With a JSON payload, we also get into the tricky situation when using the `query` extension where the value of one of the  attributes in the JSON is a string representing another query language, e.g., CQL or GraphQL, which can expose all sorts of bugs in JSON implementations (similarly to how our use of `:` in extension prefixes do)

I see more APIs that use this convention. for example, https://developer.twitter.com/en/docs/tweets/search/api-reference/enterprise-search#Methods

joshfix: That would need to be filed as a separate issue.  From the beginning we have used raw application/json.  This examples page is examples of the current spec.

api-spec/examples.md
joshfix: Sounds good

api-spec/examples.md
philvarner: Ah right, will do that.

matthewhanson:
This looks good to me.  Like the idea of @philvarner about multipart, but agree that should be different PR.

I won't merge quite yet, in case @joshfix wants to tweak the verbiage somewhat based on the suggestions by @philvarner

api-spec/examples.md
philvarner: Created a PR for some explicit language around supporting POST. #489

joshfix:
I updated the verbiage to say "items" instead of "data".  Should be good to go.

Issue #356, opened 2018-11-16
Satellite extension / similar fields in extensions


m-mohr:
Now that we are soon getting more similar extensions (e.g. EO, SAR) that share a certain number of fields (platform, constellation, instrument, orbit, ...) we should think about extracting them into a separate extension or so. Could be Satellite (`sat`) for example, but platform / Instrument could also apply for drones or generally non-satellites, so we need to carefully decide this in the future.

Not necessarily required for 0.7, but sometime in the future. Opening this now so we don't forget it.

cholmes:
Relevent discussion in #357

m-mohr:
Another issue that we are facing is that people are confused by the EO extension. They assume EO means earth observation and are not sure whether SAR has to be used in addition or is a separate thing. Having EO as Electro-Optical does not seem intuitive. The question is whether we want to combine common fields in EO and SAR into an EO extension and move purely Electro-Optical fields into someting like an Opical extension. @mojodna

geospatial-jeff:
>  people are confused by the EO extension

I'm one of these people.  I think it is trying to do too much.  It has information about the projection (epsg) and sensor (sensor/instrument/gsd, bands) and scene-level metadata (off_nadir/sun_azimuth/etc.).  Rather than a one size fits all solution, I like the idea of strategically crafting extensions around the raster data model.

This is why I really like the proj extension; it does a good job of describing the coordinate system of the image, an important part of the raster data model.  It provides a single, standardized, and independent way of searching the projection of raster data.  It's more intuitive to use than remembering that epsg is hiding in the EO extension and will maintain separation of concerns because the extension does what it is designed to do and nothing else.

I like the idea of a `sat` extension (`aerial` and `drone` extensions as well) for the same reason.  The extended raster data model acknowledges that raster data doesn't exist on its own but comes from various sensors (SAR/electro-optical/LIDAR etc.) attached to a certain platform (usually drone, aerial, or satellite) which are sensing the target (in this case the Earth) under various conditions (cloud cover etc.).

A `sat` extension nicely abstracts an important component of the raster data model and allows some properties to be offloaded from the EO extension.  Personally I think that gsd, platform, constellation, instrument, and bands would all be more appropriate in the `sat` extension.  I would argue these fields are intrinsic properties of the specific platform/sensor whereas extensions such as SAR/EO which are describing sensor-specific, item-level metadata should be exclusively concerned with extrinsic properties such as cloud cover, nadir angle, etc.

[...]

PR #485, opened 2019-06-08
Projection Extension

philvarner:
**PR Checklist:**


extensions/eo/README.md
```diff
@@ -25,12 +25,11 @@ The exact metadata that would appear in a STAC Collection record will vary depen
...
 | ---------------- | ------------------------ | ----------- |
-| eo:gsd           | number                   | **REQUIRED.** Ground Sample distance. The nominal distance between pixel centers available, in meters. |
+| eo:gsd           | number                   | **REQUIRED.** Ground Sample Distance. The nominal distance between pixel centers available, in meters. |
 | eo:platform      | string                   | **REQUIRED.** Unique name of the specific platform the instrument is attached to. For satellites this would be the name of the satellite (e.g., landsat-8, sentinel-2A), whereas for drones this would be a unique name for the drone. |
 | eo:constellation | string                   | Name of the constellation that the platform belongs to. See below for details. |
 | eo:instrument    | string                   | **REQUIRED.** Name of instrument or sensor used (e.g., MODIS, ASTER, OLI, Canon F-1). |
 | eo:bands         | [Band Object]            | **REQUIRED.** This is a list of the available bands where each item is a Band Object. |
-| eo:epsg          | integer\|null            | [EPSG code](http://www.epsg-registry.org/) of the datasource, `null` if no EPSG code. |
```
philvarner: the epsg id isn't specific to EO, so I feel a better place for it would be the `proj` extension.  Completely open to not removing this.

extensions/projection/schema.json
```diff
@@ -0,0 +1,52 @@
...
+            "type": "number"
+          }
+        },
+        "proj:centroid": {
+          "title": "Centroid",
+          "type": "array",
+          "minItems": 2,
+          "maxItems": 2,
+          "items": {
+            "type": "number"
```
philvarner: restrict these to -180 to 180 to make it clear it's lat lon

m-mohr:
Some unsorted thoughts:

Having it in a separate extension is better, thanks.

I'm not sure whether I'd remove the epsg code from EO. 

We decided some versions ago to keep epsg and drop WKT/PROJ as STAC is focused on search. You can search for EPSG codes quite well, but nobody would ever search for a PROJ/WKT string, we argued. Is this still what we think/want?

I think this extension should have a required field so that validation works.

Should we also allow WKT?

A GeoJSON geometry that is not in EPSG:4326 is not GeoJSON, therefore we shouldn't name it GeoJSON...

What is the centroid useful for?

philvarner:
>> I'm not sure whether I'd remove the epsg code from EO.

I will re-add it.

philvarner:
>> I think this extension should have a required field so that validation works.

I will add one

philvarner:
> GeoJSON geometry that is not in EPSG:4326 is not GeoJSON, therefore we shouldn't name it GeoJSON...

Good point, will update words.

philvarner:
> What is the centroid useful for?

There are a lot of tools, for example Kibana Maps, which can generate heatmaps based on point data. This would allow tools to do this without needing to do the (relatively expensive) calculation at render-time based on the geometry or bbox.

philvarner:
> We decided some versions ago to keep epsg and drop WKT/PROJ as STAC is focused on search. You > can search for EPSG codes quite well, but nobody would ever search for a PROJ/WKT string, we 
> argued. Is this still what we think/want?
> Should we also allow WKT?

As a general point, I'd like STAC to add more features around usage of data, beyond the existing exploration/query focus.  Ideally, we'd like to do as much of our compute-time spatial filtering, etc, based on the metadata in the STAC entities without needing to retrieve any  of the asset bits until we actually need the pixels.  For example, the COG headers to do the offset retrieval without needing to fetch the header for each request. 

This is one of those options. We're using PROJ4 internally instead of just EPSG code, though I need to ask my coworkers why we chose to do that.  I also wanted to be as general as possible, in case there are projections that don't have EPSG codes (even though I can't think of any and don't think I'll have any use of that.

m-mohr:
> I will re-add it.

It was just my opinion. We should discuss it in a broader group, at least with @matthewhanson. Best on the next bi-weekly call.

> I will add one

It will also require a JSON Schema, of course ;-)

> There are a lot of tools, for example Kibana Maps, which can generate heatmaps based on point data. This would allow tools to do this without needing to do the (relatively expensive) calculation at render-time based on the geometry or bbox.

Oh, I see. At one hand I'd like to keep the spec and fields always relatively small, on the other hand I see that this can be useful. Hmm...

> As a general point, I'd like STAC to add more features around usage of data, beyond the existing exploration/query focus. Ideally, we'd like to do as much of our compute-time spatial filtering, etc, based on the metadata in the STAC entities without needing to retrieve any of the asset bits until we actually need the pixels. For example, the COG headers to do the offset retrieval without needing to fetch the header for each request.

Very similar for us. We also add processing information (e.g. offset, scale, no data values) for processing as we need it within openEO. As I said above, it's hard to keep things simple and useable at the same time. We don't want to be the next ISO 19115, right? ;-)

> This is one of those options. We're using PROJ4 internally instead of just EPSG code, though I need to ask my coworkers why we chose to do that. I also wanted to be as general as possible, in case there are projections that don't have EPSG codes (even though I can't think of any and don't think I'll have any use of that.

There are indeed some projections that have no EPSG code. I think there were some in the context of MODIS data. I also remember this discussion when we removed PROJ (and WKT) from the EO extension.
Should we rename proj:crs to proj:proj (well, interesting combination though) so that we could also add proj:wkt  at some point? Or whould we allow both in proj:crs? I'm always unsure which to allow in addition to EPSG. In openEO we started with proj, added WKT, but overall it feels like a mess.

geospatial-jeff:
> I'm always unsure which to allow in addition to EPSG. In openEO we started with proj, added WKT, but overall it feels like a mess.

From my experience, WKT is the messiest with the least standardization between vendors (see [https://github.com/cf-convention/cf-conventions/wiki/OGC-WKT-Coordinate-System-Issues](here)).  As mentioned above, EPSG is the most searchable and is more standardized than WKT but still has some issues.  My biggest issue with EPSG is its inability to describe custom coordinate systems, only those which reside in the EPSG database.  PROJ4 is the best for explicitly describing a spatial reference and provides a very flexible representation but is harder to search.

> Should we rename proj:crs to proj:proj (well, interesting combination though) so that we could also add proj:wkt at some point?

I think a `proj:type` field which indicates the type of spatial reference expressed in `proj:crs` would be helpful.  These two fields could then be used to define the spatial reference of an item in cases where there is no valid EPSG value (where `proj:epsg` is null) while giving the user flexibility in their choice of representation.

> You can search for EPSG codes quite well, but nobody would ever search for a PROJ/WKT string, we argued

This got me thinking about searches on PROJ strings.  A simple implementation could decompose the PROJ string into a JSON of its parts, potentially exposing more advanced searches on the item's spatial reference than possible with EPSG.  Consider the PROJ string for WGS 84 UTM Zone 11 (EPSG:32611):

`+proj=utm +zone=11 +ellps=WGS84 +datum=WGS84 +units=m +no_defs`

```json
{
  "proj:type": "proj4",
  "proj:crs": {
    "proj:proj": "utm",
    "proj:zone": "11",
    "proj:ellps": "WGS84",
    "proj:datum": "WGS84",
    "proj:units": "m"
  }
}
```

This would allow searching across individual components of the spatial reference (such as datum) which is not always possible when searching with EPSG.

matthewhanson:
As we talked about in the sprint, my overall concern here is bloating the individual Items.  I get that these records are small size wise, relative to assets, but if we end up adding lots of extensions for pretty specific purposes, the size of a STAC Item can get very large and contain a lot of data many users will not be concerned about.

For instance, as the creator of the Landsat and sentinel catalogs I would not be inclined to include the information for this extension.

However, one alternative, which we talked briefly about, is to think about adding details usage information in standardizes assets. That is, we could have extensions that don't add this information to the STAC Item, but rather then define a specific asset that the Item should contain. For instance, a projection asset that is an JSON asset with a specific structure.

I could also see this approach being used for other more complex info that is for a specific use, such as the visualization extension (#470).  This way, if you want to use that info after getting the STAC Item, you just need to fetch that JSON asset, but it doesn't force everyone to fetch it.

philvarner:
I appreciate the concerns with bloating the items, but I feel that this is fundamental spatial information about the items and assets, rather than a "pretty specific purpose", even to the point that this should be part of core rather than an extension.

By not including the type of information in the Landsat and Sentinel-2 catalogs, it limits the utility of it for a significant class of users. In my experience, we found *almost* all the data we needed in the catalogs, but not quite, so we had to pull it from the source data or metadata files anyway, which was pretty costly in terms of development and run-time.

m-mohr:
The PR as it is right now would require all users to also provide a PROJ string in addition to the EPSG code. Why should I be required to add both if it could simply be a EPSG code for most(?) use cases? I think I would only require proj:epsg...


> From my experience, WKT is the messiest with the least standardization between vendors (see [here](https://github.com/cf-convention/cf-conventions/wiki/OGC-WKT-Coordinate-System-Issues)). 

Is that solved with WKT2?

geospatial-jeff:
> Is that solved with WKT2?

Yes WKT2 fixes a lot of the standardization issues within the spec itself but there are still lots of vendors who don't use WKT2.  I've always preferred PROJ over WKT but since GDAL recently added support for WKT2 in 3.0 and already supports both OGC and ESRI style WKT strings, it could be a nice alternative.

> I think I would only require proj:epsg...

Agreed.  I see this extension primarily as a way of defining the spatial reference of the item in cases where an EPSG code is not satisfactory, so it makes sense to keep the other fields optional.

philvarner:
Updated so only `proj:epsg` is required.  Also made explicitly that these should be in Properties rather than in the Item root. Since Assets are not defined to have Properties (maybe they should?), I removed that as a target type for the extension attributes. 

 However, I see that we have at least two extensions (sar and checksum) that add properties directly into Asset -- so maybe we should have the convention that extension attributes are only added into Properties, and change those ones?

m-mohr:
> Also made explicitly that these should be in Properties rather than in the Item root.

All other extension only just say "Item fields" and not "Item Properties fields". So either it's clear or we should do it for all extensions.

> Since Assets are not defined to have Properties (maybe they should?)

Feels like too much nesting. We basically only use properties because GeoJSON wants us to. I would also be fine to have everything top level ;-)

> However, I see that we have at least two extensions (sar and checksum) that add properties directly into Asset

EO has that, too.

> so maybe we should have the convention that extension attributes are only added into Properties, and change those ones?

Not sure, see above. For example, we don't have properties in collections. I think extensions could add anywhere, but properties are required for Items due to the GeoJSON spec.

philvarner:
One example of common projection without an EPSG code is Sinusoidal, used by MODIS.

philvarner:
> > Also made explicitly that these should be in Properties rather than in the Item root.
> 
> All other extension only just say "Item fields" and not "Item Properties fields". So either it's clear or we should do it for all extensions.

We should do it for all extensions to make it clear where the the attributes are supposed to be added.  although in practice, most people will just look at the examples.

philvarner:
I believe all of the issues have been addressed.  @m-mohr @matthewhanson , please let me know if that's not the case.

extensions/projection/README.md
```diff
@@ -0,0 +1,124 @@
...
+'projection') used by the data, and can usually be referenced using an [EPSG code](http://epsg.io).
+If the data does not have a CRS, such as in the case of non-rectified imagery with Ground Control
+Points, proj:epsg should be set to null. It should also be set to null if a CRS exists, but for which
+there is no valid EPSG code.
+
+**proj:crs** - A Coordinate Reference System (CRS) is the native reference system (sometimes called a
+'projection') used by the data. This value is a PROJ string.
+If the data does not have a CRS, such as in the case of non-rectified imagery with Ground Control
+Points, proj:crs should be set to null. It should also be set to null if a CRS exists, but for which
+a PROJ string does not exist.
```
metasim: Should it be made explicit that the `proj4` format is required? Some people may assume WKT, which has some issues from what I understand.

Another option would be to support multiple formats. This could be done explicitly with a `type` or `format` specifier, or implicitly, like what's [done in RasterFrames](https://github.com/locationtech/rasterframes/blob/3538bda126d885e3daa8b3913a82b3890a78da23/core/src/main/scala/org/locationtech/rasterframes/model/LazyCRS.scala#L53-L57) whereby a single field is used to specify the CRS regardless of format, since each format has a different prefix.

My preference would be to have a single `proj:crs` field that accepts strings of the form:

* `"epsg: 32614"`
* `"+proj=utm +zone=14 +datum=WGS84 +units=m +no_defs"`
* `"PROJCS["WGS_1984_UTM_Zone_14N",GEOGCS["GCS_WGS_1984",DATUM["D_WGS_1984",SPHEROID["WGS_1984",6378137,298.257223563]],PRIMEM["Greenwich",0],UNIT["Degree",0.017453292519943295]],PROJECTION["Transverse_Mercator"],PARAMETER["latitude_of_origin",0],PARAMETER["central_meridian",-99],PARAMETER["scale_factor",0.9996],PARAMETER["false_easting",500000],PARAMETER["false_northing",0],UNIT["Meter",1]]"`

rather than separate fields for two different types. The downside is that you lose stronger typing on the EPSG code value, and I'm unclear how this would affect searching (just document it?).

extensions/projection/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+**proj:geometry** - A Polygon object representing the footprint of this item, formatted according the Polygon object format specified in [RFC 7946, sections 3.1.6](https://tools.ietf.org/html/rfc7946), except not necessarily in EPSG:4326 as required by RFC7946.  Specified based on the `proj:crs` field (not necessarily EPSG:4326). Ideally, this will be represented by a Polygon with five coordinates, as the item in the native CRS should be a square aligned to the CRS grid.  It is recommended that either or both of `proj:geometry` and `proj:extent` be defined.
+
+**proj:extent** - Bounding box of the asset represented by this item in the native CRS. Specified as four coordinates based on the CRS defined in the `proj:epsg` and `proj:crs` fields.  First two numbers are coordinates of the lower left corner, followed by coordinates of upper right corner, e.g., \[west, south, east, north]. It is recommended that either or both of `proj:geometry` and `proj:extent` be defined.
```
metasim: Does saying 

> e.g., [west, south, east, north] 

imply a certain class of CRSs? Perhaps "[xmin, ymin, xmax, ymax]" instead?

metasim: I'd consider further clarification/disambiguation on what it means to provide both `geometry` *and* `extent`. Does one take precedence? Is one expected to be more precise than another?

Because this spec is for imagery in a rectilinear grid and an extent + CRS precisely defines the coverage, and the approximate 4326 geometry is already covered in the parent spec, I'd suggest removing the `proj:geometry` field, and require the expression to be in extent form.

extensions/projection/schema.json
```diff
@@ -0,0 +1,52 @@
...
+        },
+        "proj:extent": {
+          "title": "Extent",
+          "type": "array",
+          "minItems": 4,
+          "maxItems": 4,
+          "items": {
+            "type": "number",
+            "minimum": -180,
+            "maximum": 180
```
metasim: Remove these `max`/`min` values... they are CRS specific.

extensions/projection/schema.json
```diff
@@ -0,0 +1,52 @@
...
+        },
+        "proj:centroid": {
+          "title": "Centroid",
+          "type": "array",
+          "minItems": 2,
+          "maxItems": 2,
+          "items": {
+            "type": "number",
+            "minimum": -180,
+            "maximum": 180
```
metasim: Remove these `max`/`min` values... they are CRS specific.

extensions/projection/README.md
philvarner: The reasoning was that you'd likely search on the espg field, could search on the PROJ field if that didn't exist, and could use the PROJ field if the epsg field was null. 

I didn't add a WKT option as WKT Original Flavor has some ambiguity and implementation concerns with it (that @geospatialjeff mentions above), and WKT2 isn't yet widely adopted.

philvarner: It's explicit that PROJ is required in the fields table

extensions/projection/README.md
philvarner: I think the issue with min/max terminology is that for antimeridian-crossing boxes, that xmin is actually greater than xmax, which doesn't have the same implication with west/east

extensions/projection/schema.json
philvarner: fixed.

extensions/projection/schema.json
philvarner: fixed

extensions/projection/README.md
metasim: What about a CRS (e.g. [EPSG:22182](http://www.epsg-registry.org/export.htm?wkt=urn:ogc:def:crs:EPSG::22182)) where positive X is "northing"? I'd worry about mistakes being made if coordinates have to be flipped in the expression of extent.

geospatial-jeff: Should this align with the bbox definition in the item spec?  In the spec its defined as [lower_left_corner, upper_right_corner].   From my experience,  [xmin,ymin,xmax,ymax] is a modification of the traditional bounding box expression with corner coordinates.  It is more intuitive and almost always correct (with the exception of antimeridian crossings as Phil mentions) but is less clear than the corner expression in my opinion.

m-mohr: We should strictly follow the Item and Collection spec, which are following the GeoJSON and WFS3 spec. I think we should not diverge from that.

m-mohr: You could also ask this in the item spec, which also has bbox and geometry. 

By the way, why is proj:extent not called proj:bbox to align with the GeoJSON fields in the item spec?

geospatial-jeff: There are also edge-cases where an image can be in a projected coordinate system but the extent and geometry don't match due to rotation in the geotransform, in which case both `proj:geometry` and `proj:extent` would be required to describe the spatial properties of the image.

metasim: But in this extension, `proj:geometry` is defined in the native crs (`proj:crs`), so technically they are redundant. This is distinct from the GeoJSON case where the `bbox` is in EPSG:4326 and and a separate (yet still approximate)  `geometry` is needed due the aforementioned geotransform rotation (which is *not* an edge case in our use cases.)

So, in the Item spec you need at least a `geometry` to get the approximate footprint, and a `bbox` gives you a convenient and fast approximation. In the `proj` extension everything is in the native CRS, so (I'm assuming?) the footprint can be fully expressed as an extent/bbox (unless there's some spherical geometry condition I'm not considering).

metasim: The problem with following the GeoJSON spec (I don't know about WFS3) is that the CRS is assumed to be 4326, so north is positive Y, east is positive X, etc. 

> A GeoJSON object MAY have a member named "bbox" to include  information on the coordinate range for its Geometries, Features, or  FeatureCollections.  The value of the bbox member MUST be an array of  length 2*n where n is the number of dimensions represented in the  contained geometries, with all axes of the most southwesterly point  followed by all axes of the more northeasterly point.  The axes order  of a bbox follows the axes order of geometries.  The "bbox" values define shapes with edges that follow lines of  constant longitude, latitude, and elevation.

At least that's how I'm interpreting the above.... so feel free to correct me.

No such assumption can be made of arbitrary CRSs. You'd have to project back into 4326 (or similar) to fully comply with the spec, completely negating the purpose of this extension spec.

geospatial-jeff: The difference in `geometry` and `bbox` you are referring to (such as seen in Landsat8 images) is not due to geotransform rotation but rather because the Landsat image is resampled onto a square array for ease-of-use.  If you pull the geotransform of any standard Landsat8 it will be 0, indicating no rotation -- even though visually there is rotation.

I believe you are making an assumption that images in projected coordinate systems always line up with the crs axis but this is only true when the image is not tilted.  When no tilt is present, the `bbox` and `geometry` will perfectly coincide (with a jaccard index of 1).  If there is tilt in the geotransform the array will not align with the crs axis, causing disparate `bbox` and `geometry` (with a jaccard index of < 1).  This is similar to the Landsat images mentioned above but its happening for a different reason.  

Because of this I think that the potential for `proj:geometry` and `proj:extent` to differ within a single item is a valid edge case across both geogcs and projcs.

matthewhanson: I think @geospatial-jeff meant that the extension fields should at least match in name to the GeoJSON terms.  In other words, bbox is used at the GeoJSON feature level to describe the bbox, so if the proj extension provides the bounding box in a different CRS it should follow the same naming convention.
It also avoids confusion with the "extent" field which is used at the collection level and has a quite different meaning (the bbox of the convex hull containing all member Items).

metasim: Thanks for explaining!

philvarner: I should have named this `bbox` instead of extent if that's used in collection.  I'm also okay leaving this out if it's too ill-defined outside of 4326.

joshfix:
> > What is the centroid useful for?
> 
> There are a lot of tools, for example Kibana Maps, which can generate heatmaps based on point data. This would allow tools to do this without needing to do the (relatively expensive) calculation at render-time based on the geometry or bbox.

Also useful for elasticsearch geo aggregations as it relies on points.

m-mohr:
What's the state of this PR? @philvarner 
Still WIP or ready to review and merge?

cholmes:
> One example of common projection without an EPSG code is Sinusoidal, used by MODIS.

This came up before... As the person who came up with using EPSG:900913 for the google projection (before epsg adopted 3857), I think we should just define an ad hoc code for the MODIS one, and put it in EPSG 'user space' (I forget what they call it, but anything above a certain number is not defined by EPSG but can be used).

philvarner:
Remove WIP designation.

philvarner:
I think creating a MODIS id is a good idea.  I wasn't able to find any info on what the user-defined space is, but I picking something between the highest currently-used code of EPSG:32766 and the deprecated ones starting with EPSG:61206405 is safe.

m-mohr:
We had a discussion yesterday during the telco on how to handle extensions that are not primarily for search and discovery, but more for analytics and other domains. Not quite sure into which category this extension would fall, but the idea was to only have extensions in this repo that are for search and discovery. As this PR is quite controversial (it seems?), I think we should discuss it in the next telco again and decide for a way to go forward.

philvarner:
After discussion at the July 22 telecon, I believe the consensus was 

1. STAC is primarily for search/query, but metadata-only related extensions (like this one) can also be added.
2. These non-query metadata extensions can go into this repo for now, but will likely be reorganized somehow later.

m-mohr:
>     2. These non-query metadata extensions can go into this repo for now, but will likely be reorganized somehow later.

I'd propose to add these non-query extensions to the third party / vendor extensions table: https://github.com/radiantearth/stac-spec/tree/master/extensions#third-party--vendor-extensions

That makes it clear that they are not core.

As it's easy to add but hard to remove things, I'd say we should move the non-query extensions for 1.0.0 at latest and make an issue for 1.0.0 so that we don't forget.

m-mohr:
> This got me thinking about searches on PROJ strings. A simple implementation could decompose the PROJ string into a JSON of its parts, potentially exposing more advanced searches on the item's spatial reference than possible with EPSG. Consider the PROJ string for WGS 84 UTM Zone 11 (EPSG:32611):
> 
> `+proj=utm +zone=11 +ellps=WGS84 +datum=WGS84 +units=m +no_defs`
> 
> ```json
> {
>   "proj:type": "proj4",
>   "proj:crs": {
>     "proj": "utm",
>     "zone": "11",
>     "ellps": "WGS84",
>     "datum": "WGS84",
>     "units": "m"
>   }
> }
> ```
> 
> This would allow searching across individual components of the spatial reference (such as datum) which is not always possible when searching with EPSG.

There's related work for PROJ by @rouault, see https://github.com/OSGeo/PROJ/pull/1547

philvarner:
> There's no link from the extensions README to this extension (see my related comment).

I added this to the core extensions list.  If this is going to be a non-core extension (e.g., vendor), then it should be in a different repo, and, to be perfectly honest, I probably won't purse this being standardized in any way. 

> Also, no CHANGELOG entry.

Added.

> Question: Do we have eo:epsg AND proj:epsg after this gets merged?

Yes.  I'd initially remove eo:epsg when I created this PR, but then readded when it was pointed out that an implementation may decide not to implement both extensions.

extensions/projection/schema.json
m-mohr: There's no restriction in place?!

extensions/projection/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Item Properties fields
+
+| Field Name       | Type                     | Description |
+| ---------------- | ------------------------ | ----------- |
+| proj:epsg        | integer\|null   | **Required** [EPSG code](http://www.epsg-registry.org/) of the datasource |
+| proj:crs         | string \|null   | Recommended. PROJ string representing the Coordinate Reference System (CRS) that the `proj:geometry` and `proj:bbox` fields represent |
+| proj:geometry    | Polygon Object  | Recommended. Defines the footprint of this item. |
+| proj:bbox        | [number]        | Recommended. Bounding box of the item in the native CRS |
+| proj:centroid    | Centroid Object | Recommended. Coordinates representing the centroid of the item in the native CRS |
```
m-mohr: Wouldn't it make the spec easier to not define a new Centroid object, but just make it a two element array? Would be more inline with the bbox.

m-mohr: Also, it feels odd that all fields are recommended. There should be an indication on why the fields are recommended.

extensions/projection/schema.json
```diff
@@ -0,0 +1,48 @@
+{
```
m-mohr: The schema doesn't "inherit" from the item-spec.

extensions/projection/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+**Extension [Maturity Classification](../README.md#extension-maturity): Proposal**
+
+This document explains the fields of the STAC Projection (`proj`) Extension to a STAC Item. Here `proj` is short
+for "projection", and not a reference to the use of the PROJ format for the CRS attribute.
+
+The field names defined herein should be added as fields in the Item's Properties object. 
+
+When specified on an Item, the values are assumed to apply to all Assets in that Item.  For example, an Item may have several related Assets each representing a band for the Item, all of which are in the same native CRS, e.g., a UTM Zone.  However, there may also be assets intended for display, like a preview image or thumbnail, that have been reprojected to a different CRS, e.g., Web Mercator, to better accommodate that use case.  This case of differing projections per Asset is not currently handled by this extension.
+
```
m-mohr: Other extensions link to the JSON schema and a full example (which I am missing here):

* [Example](examples/example.json)
* [JSON Schema](json-schema/schema.json)

extensions/projection/schema.json
```diff
@@ -0,0 +1,58 @@
...
+              "maxItems":6,
+              "items":{
+                "type":"number"
+              }
+            },
+            "proj:centroid":{
+              "title":"Centroid",
+              "type":"array",
+              "minItems":2,
+              "maxItems":3,
```
m-mohr: Why is maxItems set to 3 and not 2? I don't see an explanation in the spec what the third element should be.

extensions/projection/schema.json
```diff
@@ -0,0 +1,58 @@
...
+          "properties":{
+            "proj:epsg":{
+              "title":"EPSG code",
+              "type":"integer"
+            },
+            "proj:crs":{
+              "title":"Coordinate Reference System in PROJ format",
+              "type":"string"
+            },
+            "proj:geometry":{
```
m-mohr: This seems a bit too basic. Could we derive a JSON Schema from https://geojson.org/schema/Polygon.json.
Maybe we can simply $ref to https://geojson.org/schema/Polygon.json - it doesn't seem to be restricted to a CRS.

extensions/projection/schema.json
```diff
@@ -0,0 +1,58 @@
...
+      "properties":{
+        "properties":{
+          "type":"object",
+          "required":[
+            "proj:epsg"
+          ],
+          "properties":{
+            "proj:epsg":{
+              "title":"EPSG code",
+              "type":"integer"
```
m-mohr: `null` type is missing.

extensions/projection/schema.json
```diff
@@ -0,0 +1,58 @@
...
+            },
+            "proj:crs":{
+              "title":"Coordinate Reference System in PROJ format",
+              "type":"string"
```
m-mohr: `null` type is missing.

extensions/projection/schema.json
```diff
@@ -0,0 +1,68 @@
+{
+  "$schema":"http://json-schema.org/draft-07/schema#",
+  "$id":"schema.json#",
+  "title":"Projection Extension",
+  "description":"STAC Projection Extension to a STAC Item",
+  "allOf":[
```
m-mohr: The schema didn't "inherit" from the item-spec. I updated the schema for you.

extensions/projection/schema.json
```diff
@@ -0,0 +1,68 @@
...
+          "type":"object",
+          "required":[
+            "proj:epsg"
+          ],
+          "properties":{
+            "proj:epsg":{
+              "title":"EPSG code",
+              "type":[
+                "integer",
+                "null"
```
m-mohr: Added the missing `null` data for you.

extensions/projection/schema.json
```diff
@@ -0,0 +1,68 @@
...
+              ]
+            },
+            "proj:crs":{
+              "title":"Coordinate Reference System in PROJ format",
+              "type":[
+                "string",
+                "null"
```
m-mohr: Added the missing `null` data for you.

extensions/projection/README.md
```diff
@@ -0,0 +1,139 @@
...
+Points, proj:crs should be set to null. It should also be set to null if a CRS exists, but for which
+a PROJ string does not exist.
+
+**proj:geometry** - A Polygon object representing the footprint of this item, formatted according the Polygon object format specified in [RFC 7946, sections 3.1.6](https://tools.ietf.org/html/rfc7946), except not necessarily in EPSG:4326 as required by RFC7946.  Specified based on the `proj:crs` field (not necessarily EPSG:4326). Ideally, this will be represented by a Polygon with five coordinates, as the item in the native CRS should be a square aligned to the CRS grid.  It is recommended that either or both of `proj:geometry` and `proj:bbox` be defined.
+
+**proj:bbox** - Bounding box of the asset represented by this item in the native CRS. Specified as four coordinates based on the CRS defined in the `proj:epsg` and `proj:crs` fields.  First two numbers are coordinates of the lower left corner, followed by coordinates of upper right corner, e.g., \[west, south, east, north], \[xmin, ymin, xmax, ymax], \[left, down, right, up]. It is recommended that either or both of `proj:geometry` and `proj:bbox` be defined.
```
m-mohr: The spec doesn't say anything about the 5th and 6th element (JSON Schema has maxItems set to 6).

extensions/projection/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+**proj:centroid** - Coordinates representing the centroid of the item in the native coordinate system.  Coordinates are defined in latitude and longitude, even if the native coordinate system does not use lat/long.
```
m-mohr: The spec doesn't say anything about the 3rd element (JSON Schema has maxItems set to 3).

extensions/projection/README.md
```diff
@@ -0,0 +1,139 @@
...
+## Centroid Object
+
+This object represents the centroid of an asset's geometry.
+
+| Field Name          | Type   | Description                                                  |
+| ------------------- | ------ | ------------------------------------------------------------ |
+| lat                 | number | The latitude of the centroid.  |
+| lon                 | number | The longitude of the centroid. |
+
+## Examples
```
m-mohr: Could we please make this a complete example and add it to the examples folder so that it can be used for validation etc?

extensions/projection/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+When specified on an Item, the values are assumed to apply to all Assets in that Item.  For example, an Item may have several related Assets each representing a band for the Item, all of which are in the same native CRS, e.g., a UTM Zone.  However, there may also be assets intended for display, like a preview image or thumbnail, that have been reprojected to a different CRS, e.g., Web Mercator, to better accommodate that use case.  This case of differing projections per Asset is not currently handled by this extension.
+
+## Item Properties fields
+
+| Field Name       | Type                     | Description |
+| ---------------- | ------------------------ | ----------- |
+| proj:epsg        | integer\|null   | **Required** [EPSG code](http://www.epsg-registry.org/) of the datasource |
+| proj:crs         | string \|null   | Recommended. PROJ string representing the Coordinate Reference System (CRS) that the `proj:geometry` and `proj:bbox` fields represent |
+| proj:geometry    | Polygon Object  | Recommended. Defines the footprint of this item. |
```
m-mohr: Poylgon Object has no corresponding chapter that explains it?! All the other Objects in the spec usually have a corresponding sub header explaining the fields.

extensions/eo/README.md
m-mohr: Hmm, I don't feel it's a good idea to have two places that define such a field. Related discussions in #356

extensions/projection/README.md
```diff
@@ -0,0 +1,139 @@
...
+  "assets": {
+    "B1": {
+      "href": "http://example.com/L8/153/025/LC81530252014153LGN00/LC81530252014153LGN00_B1.TIF",
+      "type": "image/vnd.stac.geotiff",
+      "eo:bands": [0]
+    },
+    "thumbnail": {
+      "href": "http://example.com/L8/153/025/LC81530252014153LGN00/LC81530252014153LGN00_thumbnail.jpg",
+      "type": "image/jpeg",
+      "proj:crs": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0 +k=1.0 +units=m +nadgrids=@null +wktext +no_defs ",
```
m-mohr: Here are projection properties for an asset, but the description of the extension says:

> This case of differing projections per Asset is not currently handled by this extension.

Also, the JSON Schema doesn't have any information how to validate these information in the assets.

So I'd say the proj fields in the asset should be removed?!

geospatial-jeff:
> Also, I am not sure yet how to handle the duplication of proj:epsg and eo:epsg

I've thought about this a lot and I honestly can't come up with a single reason why I would use eo:epsg over proj:epsg.  Unless the EO extension is renamed back to Earth Observation, EPSG doesn't make sense in the extension because EPSG isn't exclusive to electro-optical imagery.  I find it curious that the EO extension was renamed from EarthObservation to ElectroOptical in order to limit the scope of the extension (https://github.com/radiantearth/stac-spec/blame/dev/extensions/eo/README.md#L11-L15) but it still contains very broad fields which are being re-used by non-electro-optical payloads (such as SAR) because eo:epsg is the only way STAC provides of describing the projection of an image.  We could fix this by also implementing a sar:epsg but now we have the problem of fragmented searches on projection across multiple payload types because epsg is split across multiple namespaces.

Even the implementation of eo:epsg within the EO extension is flawed to me since EPSG fails in a lot of use-cases (this is well discussed above) whereas the proj extension provides a much needed backup (proj:crs) for describing the projection when EPSG fails.

To me this extension is a win-win.  It provides an encapsulated and robust (at least more robust than eo:epsg) representation of the projection of the image which may be applied to any STAC Item (regardless of aerial/drone/eo/sar/lidar/whatever) while also narrowing the scope of the EO extension.

m-mohr:
So if me make proj core (which the current proposal does), I'd strongly prefer to remove eo:epsg.
If this is worked out as a vendor/third party extension, I'd *not* remove eo:epsg.

extensions/projection/README.md
hgs-truthe01: I'm wondering if centroid should live next to the top level bbox? Projections seems very specific, and centroid seems more like a quick lookup in a standardized way if you are specifying that is has to be lat long (WGS84 I assume as well?)

extensions/projection/schema.json
hgs-truthe01: I'm wondering if geometry in a specific projection might be better served as an asset as a shapefile?
I'm mostly worried that geometry is our biggest property and duplicating it seems like a big lift. Also, there are lots of ways to specify geometries so a specific format would be needed to put it in the metadata to make it readable.

m-mohr:
I discussed this a little with @cholmes before you joined, @philvarner. We discussed whether we'd split the extension in two parts, first only having proj:epsg and(?) proj:crs. The other one to have centroid, geometry and bbox. We didn't really conclude anything, but wanted to write it down here for discussion.

cholmes:
So we don't really have the concept of a 'core extension', and in general I'm comfortable having this be an extension in the repo.

My overall take on this is that we could consider making just 'proj:epsg' to be 'more core' than the other fields. Not sure exactly what form this should take, perhaps we have an 'extended properties' section as a 'core spec', or a 'common extensions'. On the call @matthewhanson suggested for these properties we'd drop the prefix, which seems ok to me.

m-mohr:
> So we don't really have the concept of a 'core extension', and in general I'm comfortable having this be an extension in the repo.

When talking about core extensions, I am referring to extensions that are maintained here long-term by us and listed at https://github.com/radiantearth/stac-spec/tree/master/extensions#list-of-content-extensions

Other extensions are the Third party / Vendor extensions, which were to be listed here: https://github.com/radiantearth/stac-spec/tree/master/extensions#third-party--vendor-extensions

And we should start to think about which extension is core and which is not before 1.0, I think.

> On the call @matthewhanson suggested for these properties we'd drop the prefix, which seems ok to me.

No extension means putting them in the item spec together with title, datetime etc? Or how do we reduce potential collisions?

cholmes:
Discussed a bit on call on 8/5 - general consensus on getting epsg to be 'core', and this to be an 'extension' that is 'less core', but those terms aren't super well defined. Short term fix could be to make this the first 'vendor extension' and put it in astraea repo to start, and then for 0.9.0 fix up all the things to make this 'better' - like defining 'core (common?) extension fields' (maybe with date time range, and surveying other things that are emerging as common'.

m-mohr:
Just a thought, but now that something like [PROJJSON](https://github.com/OSGeo/PROJ/pull/1547) is developed, could we just link from STAC with a known rel- and/or media-type to the projection information?

cholmes:
Discussed again on call 8/19 - most likely epsg is going to be an unprefixed 'core extended' field (name tbd). But for 0.9.0 we want to move up a number of commonly used properties to be deemed as more core, and to drop extension.

@philvarner - we can get this extension in for 0.9.0 easily, with that change. If you want it in for 0.8.0 you could just remove the required proj:epsg, and rely on like the EO one. Which is a bit weird, so fine to wait for 0.9.0, or happy to have it in so you can start using it.

extensions/projection/README.md
```diff
@@ -0,0 +1,93 @@
...
+The field names defined herein should be added as fields in the Item's Properties object. 
+
+When specified on an Item, the values are assumed to apply to all Assets in that Item.  For example, an Item may have several related Assets each representing a band for the Item, all of which are in the same native CRS, e.g., a UTM Zone.  However, there may also be assets intended for display, like a preview image or thumbnail, that have been reprojected to a different CRS, e.g., Web Mercator, to better accommodate that use case.  This case of differing projections per Asset is not currently handled by this extension.
+
+## Item Properties fields
+
+| Field Name       | Type                     | Description |
+| ---------------- | ------------------------ | ----------- |
+| proj:epsg        | integer\|null   | **Required** [EPSG code](http://www.epsg-registry.org/) of the datasource |
+| proj:crs         | string \|null   | Recommended. PROJ string representing the Coordinate Reference System (CRS) that the `proj:geometry` and `proj:bbox` fields represent |
```
m-mohr: Using PROJ is discouraged: https://twitter.com/EvenRouault/status/1189515652958212096
We should switch this field to use WKT(2) and/or PROJJSON.

philvarner: I don't have a strong opinion on this.  The intention of this field was as a fallback when there's no EPSG code, e.g., MODIS Sinusoidal. I doubt it will be used in practice very frequently, and almost certainly not for search. It's primarily as metadata to avoid needing to retrieve the imagery bits to get the CRS.

philvarner:
taking this out of ready for review, as we need to change the prefixes and likely remove the prefix from epsg

matthewhanson:
The last topic discussed was that we should remove eo:epsg from the EO extension, as it seems to belong here and seems ill-suited to the EO extension. 

I'll create a separate PR to remove from EO

philvarner:
Since the crs isn't required, what about having separate fields `proj:crs_proj4`, `proj:crs_wkt`, `proj:crs_projjson` depending on what your want to put in your catalog? I see this is primarily for metadata purposes rather than search purposes, so it's just to standardize the names rather than make everyone have the same fields.

matthewhanson:
Hey @philvarner, just a few more changes and we are almost there, I know this PR has been open for a while and would be good to get this into 0.9.0.

We think that the prefix, proj, should come back and the names should be simplified:

- proj:epsg
- proj:geometry
- proj:bbox
- proj:centroid

Finally, the use of proj4 strings is discouraged now that WKT2 is the new thing, so I think it should be `proj:wkt2` instead. Do you have a use case for still needing proj now that WKT2 seems to be supported in the major tools?

matthewhanson:
@philvarner to:

- remove eo:epsg field from the EO extension
- add back in proj prefix
- remove "asset_crs" from field names
- add additional fields for different types of CRS specs

extensions/projection/README.md
```diff
@@ -0,0 +1,75 @@
...
+projections per Asset is not currently handled by this extension.
+
+## Item Properties fields
+
+| Field Name       | Type                     | Description |
+| ---------------- | ------------------------ | ----------- |
+| proj:epsg        | integer\|null   | **Required** [EPSG code](http://www.epsg-registry.org/) of the datasource |
+| proj:proj4       | string \|null   | Recommended. PROJ4 string representing the Coordinate Reference System (CRS) that the `proj:geometry` and `proj:bbox` fields represent |
+| proj:wkt2        | string \|null   | Recommended. WKT2 string representing the Coordinate Reference System (CRS) that the `proj:geometry` and `proj:bbox` fields represent |
+| proj:projjson    | PROJJSON Object \|null   | Recommended. PROJJSON object representing the Coordinate Reference System (CRS) that the `proj:geometry` and `proj:bbox` fields represent |
```
m-mohr: Linking the "PROJJSON Object" to the PROJJSON readme would be nice.

extensions/projection/json-schema/schema.json
```diff
@@ -0,0 +1,89 @@
...
+            },
+            "proj:wkt2":{
+              "title":"Coordinate Reference System in WKT2 format",
+              "type":[
+                "string",
+                "null"
+              ]
+            },
+            "proj:projjson":{
+              "title":"Coordinate Reference System in PROJJSON format",
```
m-mohr: You could actually link to the official PROJJSON schema here, right?

extensions/projection/json-schema/schema.json
```diff
@@ -0,0 +1,89 @@
+{
+  "$schema":"http://json-schema.org/draft-07/schema#",
+  "$id":"schema.json#",
+  "title":"Projection Extension",
+  "description":"STAC Projection Extension to a STAC Item",
+  "allOf":[
+    {
+      "$ref": "https://raw.githubusercontent.com/radiantearth/stac-spec/master/item-spec/json-schema/item.json#/definitions/core"
```
m-mohr:
```suggestion
      "$ref": "../../../item-spec/json-schema/item.json"
```

philvarner: fixed

extensions/projection/json-schema/schema.json
philvarner: added

extensions/projection/README.md
philvarner: updated

extensions/projection/json-schema/schema.json
```diff
@@ -45,12 +45,8 @@
                 "null"
               ]
             },
-            "proj:projjson":{
-              "title":"Coordinate Reference System in PROJJSON format",
-              "type":[
-                "object",
-                "null"
-              ]
+            "proj:projjson": {
```
m-mohr: Sorry, my bad. I forgot the "null" case. It should be: 

```json
"proj:projjson": {
  "oneOf": [
    {
      "$ref": "https://proj.org/schemas/v0.1/projjson.schema.json"
    },
    {
      "type": "null"
    }
  ]
}
```

extensions/projection/json-schema/schema.json
m-mohr: Sorry, my bad. I forgot the "null" case. It should be: 

```json
"proj:projjson": {
  "oneOf": [
    {
      "$ref": "https://proj.org/schemas/v0.1/projjson.schema.json"
    },
    {
      "type": "null"
    }
  ]
}
```

m-mohr:
Here are some unresolved comments to consider/discuss:
* https://github.com/radiantearth/stac-spec/pull/485#discussion_r309105017
* https://github.com/radiantearth/stac-spec/pull/485#discussion_r309105563
* https://github.com/radiantearth/stac-spec/pull/485#discussion_r309118191
* https://github.com/radiantearth/stac-spec/pull/485#discussion_r309114595
* https://github.com/radiantearth/stac-spec/pull/485#discussion_r309111580
* https://github.com/radiantearth/stac-spec/pull/485#discussion_r309118304
* https://github.com/radiantearth/stac-spec/pull/485#discussion_r309119340

extensions/projection/schema.json
philvarner: added restriction

extensions/projection/README.md
philvarner: Centroid is this way so that it can be used directly by visualization tools like Kibana w/ Elasticsearch.  

Removed recommended, agree about that.

extensions/projection/README.md
philvarner: added

extensions/projection/README.md
philvarner: this is outdated, as it how has the lat/lon object

extensions/projection/README.md
philvarner: I think this is pretty well-described now.

extensions/projection/README.md
philvarner: they existed, but weren't linked. added links.

extensions/projection/schema.json
philvarner: Updated with a $ref to Polygon.

philvarner:
@m-mohr just addressed all of those open comments you had.

extensions/projection/examples/example-landsat8.json
```diff
@@ -0,0 +1,164 @@
+{
+  "id": "LC81530252014153LGN00",
```
m-mohr: This example is incomplete/outdated. It is missing stac_version and stac_extensions, geometry, bbox and other item related fields?!

philvarner: updated

m-mohr:
Okay, looks good now as far as I can see. But it's hard to see with so many commits and comments ;-)

Issue #484, opened 2019-06-08
Create documentation ItemCollection


philvarner:
Currently, ItemCollection is only documented in the openapi spec, and referenced in the API endpoint descriptions and Search Extension (both changes in PR now). This should have clear documentation on par with catalog-spec and item-spec. 

I think the correct description is "A GeoJSON FeatureCollection augmented with foreign members that contain values relevant to a STAC entity".  Should probably also update the Item spec with this language to make the relationship to GeoJSON clear. 

[https://tools.ietf.org/html/rfc7946#section-6]

philvarner:
Also volunteering to do this, but I don't have permissions to assign it to myself.

m-mohr:
I'm not sure whether I would add this on the same level as Item / Collection / Catalog as those are all used in both static and dynamic catalogs, but ItemCollection is only used in API. Therefore, I'd be in favor of giving it a prominent place in the API spec - or is ItemCollection also getting an equivalent in static catalogs?

I can assign you, if you'd like...

matthewhanson:
While it is used only in the API right now, it would also serve as the basis for a single file STAC catalog, so I could see this getting its own definition at the root level. In fact by keeping it in the core, we keep consistency between the static and dynamic catalogs, it's just that best practices say that static catalogs should not link to 'items'.

philvarner:
I have it written up as part of the search meta work, so I'll split it out and create a new PR to add it tonight or tomorrow.

philvarner:
Created https://github.com/radiantearth/stac-spec/pull/487/files (after an accidental PR against master)

m-mohr:
I'm a bit confused now as there are ItemCollection specifications in #482 (which is the right place to put it in my opinion) and #487.

I still disagree that this should be a core specification for now. It is only used in API for now and will probably get an extension that uses it, but in this case it just belongs in API + extension, but not in core as it is not a core entity used throughout the specs. If you don't use an extension, it's never used in static catalogs, only in dynamic. If it's a core spec we would have to explain it similarly to the other core specs, which I feel is confusing as it doesn't play a very central role. I would have a hard time explaining in presentations what ItemCollection is about and why it's an important core spec.

joshfix:
[...]

Issue #422, opened 2019-03-13
Avoid array fields in the SAR extension


m-mohr:
Yesterday, I tried to come up with a collection-level description for SAR data using the proposed extension for non-common fields, see #416 and #413. It was quite easy to do so until I reached the SAR fields defined as array of numbers. I had to define extents, e.g. range resolution would have been something like 10 to 50 meters and the azimuth resolution would have been something like 12 to 80 meters (just an arbitrary example). Unfortunately, it's not really possible to to so as the field was defined as array of numbers, see the current situation below. So I propose to change the array fields to single value properties as below in the proposal. I guess the proposal is also easier to be used in the API as it's easier to filter and query based on individual values as it is on arrays. Thoughts?

Related issue: https://github.com/radiantearth/stac-spec/issues/405

## Current situation

| Field Name            | Type               | Description                                                  |
| --------------------- | ------------------ | ------------------------------------------------------------ |
| sar:resolution        | [number]           | The maximum ability to distinguish two adjacent targets, in meters (m). The first element of the array is the range resolution, the second element is the azimuth resolution. |
| sar:pixel_spacing     | [number]           | The distance between adjacent pixels, in meters (m). The first element of the array is the range pixel spacing, the second element is the azimuth pixel spacing. Strongly RECOMMENDED to be specified for products of type `GRD`. |
| sar:looks             | [number]           | The number of groups of signal samples (looks). The first element of the array must be the number of range looks, the second element must be the number of azimuth looks, the optional third element is the equivalent number of looks (ENL). |

## Proposal

| Field Name            | Type               | Description                                                  |
| --------------------- | ------------------ | ------------------------------------------------------------ |
| sar:resolution_range        | number           | The range resolution, which is the maximum ability to distinguish two adjacent targets perpendicular to the flight path, in meters (m).  |
| sar:resolution_azimuth        | number           | The azimuth resolution, which is the maximum ability to distinguish two adjacent targets parallel to the flight path, in meters (m).  |
| sar:pixel_spacing_range     | number           | The range azimuth, which is the distance between adjacent pixels perpendicular to the flight path, in meters (m). Strongly RECOMMENDED to be specified for products of type `GRD`. |
| sar:pixel_spacing_azimuth     | number           | The azimuth pixel spacing, which is the distance between adjacent pixels parallel to the flight path, in meters (m). Strongly RECOMMENDED to be specified for products of type `GRD`. |
[...]

PR #487, opened 2019-06-10
Add ItemCollection documentation

philvarner:
#484


philvarner:
> I'm a bit confused now as there's also a ItemCollection specification in #482 (which is the right place to put it in my opinion).

The IC spec in #482 has already been moved here -- I think you were looking at an older revision.

itemcollection-spec/itemcollection-spec.md
```diff
@@ -27,7 +27,18 @@ This object describes a STAC ItemCollection. The fields `type` and `features` ar
 | type       | string | **REQUIRED.** always "FeatureCollection" to provide compatability with GeoJSON  |
 | features   | [Item] | **REQUIRED** a possibly-empty array of Items          |
 | links      | [Link] | an array of Links related to this ItemCollection   |
-          
+| properties      | Properties | an array of Links related to this ItemCollection   |
```
m-mohr: Having properties in a FeatureCollection is conflicting with the GeoJSON spec. We should NOT do the same mistake as the OGC and do things that are clearly disallowed in the GeoJSON spec!

itemcollection-spec/itemcollection-spec.md
```diff
@@ -27,7 +27,18 @@ This object describes a STAC ItemCollection. The fields `type` and `features` ar
...
+
+### Properties Object
+
+The Properties object adds additional metadata to the ItemCollection. Basically, each entry is a
+key-value pair. The values SHOULD not be an array or object to avoid GIS systems mis-rendering them.
```
m-mohr: We are not following this recommendation in Items any longer so I think we can also leave it out here.

itemcollection-spec/itemcollection-spec.md
```diff
@@ -27,7 +27,18 @@ This object describes a STAC ItemCollection. The fields `type` and `features` ar
...
+Metadata that require an object or array SHOULD be placed a level up, directly in the
+ItemCollection object. Additional fields can be introduced through extensions. It is generally allowed to add
+custom fields.
+
+There are no required fields for the core definition of ItemCollection.
+
 ## ItemCollection Example
 
 An ItemCollection is identical to a GeoJSON FeatureCollection.
```
m-mohr: No, it's not. properties are not allowed in GeoJSON.

itemcollection-spec/itemcollection-spec.md
philvarner: Ugh. 
```
  The GeoJSON "geometry" and "properties" members define a Feature
   object.  FeatureCollection and Geometry objects, respectively, MUST
   NOT contain a "geometry" or "properties" member.
```
So this brings up the question as to whether the non-GeoJSON attributes we're putting in the `properties` object should go in their own, e.g., `stacProperties`

itemcollection-spec/itemcollection-spec.md
philvarner: Okay, we should also remove from Items spec (since I copy/pasted from there)

itemcollection-spec/itemcollection-spec.md
m-mohr: What properties do you think could be in there? I don't see a use case yet. It seems nobody missed it yet? (But whatever name it will be: snake_case please ;-) )

itemcollection-spec/itemcollection-spec.md
philvarner: will remove -- I wrote that before I realized there was a `links` attribute.

itemcollection-spec/itemcollection-spec.md
m-mohr: Oh, is it? We can still (should?) have a recommendation that explains why arrays/objects are problematic, but a SHOULD seems contradicting as some of our own specifications and extensions use arrays/objects. (e.g. bands).

itemcollection-spec/itemcollection-spec.md
m-mohr: Wording should be similar to other specs (foreign members, ...).

itemcollection-spec/itemcollection-spec.md
philvarner: snake_case, of course.  Right now, we have the Search extension that will augment the ItemCollection with metadata about search.  One option would be to have the convention that Item/Feature extension attributes should go into `properties` (flat), but ItemCollection/FeatureCollection attributes should go into another object (nested)?

itemcollection-spec/itemcollection-spec.md
philvarner: Maybe a more precise recommendation should be to use a nested data structure only when the data itself is actually nested (as with bands), but not when you just want to group a set of related attributes together.  Related attributes should be flat and to use prefixes on the attributes (as we do with extensions) when you just want to group them.

itemcollection-spec/itemcollection-spec.md
m-mohr: I have a hard time understanding this / coming up with examples. But regarding the search extension, could this just be placed on the root level in a separate object?

```
{
  search:pagination: {
    next: ...
  },
  type: "FeatureCollection",
  features: [...], 
  ...
}
```

Something like that?

itemcollection-spec/itemcollection-spec.md
philvarner: That wording is at the top.  I was trying to point explicitly that a ItemCollection from core with no extensions is a default GeoJSON FeatureCollection, which isn't true because of `links`.

m-mohr: Oh, haven't realized that. Thanks for pointing it out.

itemcollection-spec/itemcollection-spec.md
philvarner: Yes, that's exactly what we'd done with the search extension, before i started down this path of thinking it should be in Properties. I'll just revert to what we had before for search extension.

itemcollection-spec/itemcollection-spec.md
m-mohr: Sounds good for me. That's exactly the ruleset we also just applied to the SAR extension, which used quite a lot of arrays for grouping. See #422.

philvarner:
This PR is ready for final review

api-spec/api-spec.md
```diff
@@ -42,7 +42,7 @@ STAC provides some additional endpoints for the root Catalog itself, as well as
 | Endpoint      | Returns          | Description        |
 | ------------ | ------------- | ---------------------- |
 | /stac | Catalog        | Root catalog |
-| /stac/search | Items | GeoJSON FeatureCollection of Items found |
+| /stac/search  | [ItemCollection](../itemcollection-spec/README.md) | Retrieves a group of Items matching the provided search predicates, probably containing search metadata from the `search` extension |
```
m-mohr: Link is broken

item-spec/examples/itemcollection-sample-minimal.json
```diff
@@ -0,0 +1,4 @@
+{
```
m-mohr: I feel like this is too minimalist. I think this could be removed without loosing any valuable information.

item-spec/item-spec.md
```diff
@@ -1,6 +1,10 @@
-# STAC Item Specification
+# STAC Item and ItemCollection Specification
 
-This document explains the structure and content of a SpatioTemporal Asset Catalog (STAC) Item. Each
+This document explains the structure and content of a SpatioTemporal Asset Catalog (STAC) Item and ItemCollection. 
+
+## Item fields
```
m-mohr: This section should be named just "Item" as it explains more then just the Item fields.

item-spec/item-spec.md
```diff
@@ -21,18 +25,6 @@ required fields is a valid STAC Item.
   - Real world [implementations](../implementations.md) are also available.
 - [JSON Schema](json-schema/item.json)
 
-## WARNING
```
m-mohr: Why has the warning and the Item fields header been removed?

item-spec/item-spec.md
```diff
@@ -210,13 +202,62 @@ Common STAC Item Media Types:
...
 
-## Extensions
+### Item Extensions
 
 There are emerging best practices, which in time will evolve in to specification extensions for
 particular domains or uses.
 
 The [extensions page](../extensions/README.md) gives an overview about relevant extensions for STAC Items.
 
+## ItemCollection fields
```
m-mohr: Wondering whether we should move that into a separate file in the item-spec folder so that ItemCollection and Item have separate files. Basically as they were before and then both are linked from in the README.md in the item-spec. Would make the individual pages more slim and helps concentrating on what people are looking for. Also, makes linking to the specs easier.

item-spec/examples/itemcollection-sample-minimal.json
philvarner: I think it's important to specify the minimal case.  I think it makes clear what is optional.

item-spec/item-spec.md
philvarner: This is only in 3 of the many specs, and is also in the README.  I think it's time to take it out the specs, as a warning like this made more sense when the project was at a less mature stage.  I can re-add it you want.

m-mohr: It's in all core JSON specs (item, collection, catalog). So if you remove it here, remove it in all others, too. But we should leave it open to @cholmes to decide when he thinks we are "stable" (1.0?).

philvarner: re-added it in the right place.

itemcollection-spec/itemcollection-spec.md
philvarner: Updated the language

item-spec/item-spec.md
philvarner: I think it's simpler just to put them in the same file for now, since all of the other specs are one spec per directory.

m-mohr: I would say one spec per file and not mixing specs. Also, the Item spec is quite long now. But just my 2 cents...

philvarner: okay, i split it out.

m-mohr: Thanks. Having an item-spec.md and an itemcollection-spec.md in the same folder, borh linked from the item-spec/README:md makes moste sense to me.

PR #489, opened 2019-06-10
STAC API description of POST implementation

philvarner:


api-spec/README.md
```diff
@@ -1,39 +1,40 @@
...
+The API is a [Web Feature Service 3.0 (WFS 3) API](https://github.com/opengeospatial/WFS_FES), in that WFS 3 defines many of the endpoints that STAC uses. A STAC API should be compatible and usable with WFS3 clients. However, WFS 3 is still under development and while STAC tries to stay in sync with WFS3 developments, there may be discrepancies prior to final versions of both specifications.
 
 ## In this directory
 
 **The Specification:** The main description of the STAC API specification is in the *[api-spec.md](api-spec.md)* file. It includes an overview and in depth explanation of the REST endpoints and parameters.
 
-**Extensions:** API Extensions are given in the *[extensions](extensions/)* folder. YAML fragments are provided for each extension with details provided in the *[README](extensions/README.md)*.
```
philvarner: IntelliJ flags these as references as an error because they should just be a link to the directory.

api-spec/api-spec.md
```diff
@@ -50,3 +52,17 @@ The `/stac/search` endpoint is similar to the `items` endpoint in WFS3 in that i
...
+
+| Parameter    | Type            | APIs       | Description        |
+| -----------  | --------------- | ---------- | ---------------------- |
+| collectionId | [string]        | WFS3       | **Path-only** Single Collection ID to include in the search for items. Only Items in one of the provided Collection will be searched |
+| limit        | number          | WFS3, STAC | The maximum number of results to return (page size). Defaults to 10 |
+| bbox         | [number]        | WFS3, STAC | Requested bounding box [west, south, east, north] |
+| time         | string          | WFS3, STAC | Single date, date+time, or a range ('/' seperator), formatted to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6) |
+| intersects   | GeoJSON Feature | STAC       | Searches items by performing intersection between their geometry and provided GeoJSON Feature |
+| page         | number          | STAC       | The page number of results. Defaults to 1 |
```
philvarner: the change from page -> next will be picked up by a different PR.

api-spec/README.md
m-mohr: Well, depends... on mobile it just shows the directory content, but not the README file, which is more useful to navigate to, I think.

api-spec/api-spec.md
```diff
@@ -1,43 +1,45 @@
-## STAC API Specification
+# STAC API Specification
+
+The STAC API is intended to be a superset of the WFS 3.0 API specification.
```
m-mohr: I think this should be just WFS 3, at least we call it like this in all the other documents.

api-spec/api-spec.md
```diff
@@ -1,43 +1,45 @@
...
+
+In WFS 3 Features are the individual records within a Collection and are provided in GeoJSON format. [STAC Items](../item-spec/README.md) are analogous to WFS 3 Features, are in GeoJSON, and are returned from the `/collections/{collection_id}/items/{item_id}` endpoint.
+
+## HTTP Request Methods and Content Types
 
-In WFS 3 Features are the individual records within a Collection and are provided in GeoJSON format. [STAC Items](../item-spec/README.md) are analagous to WFS 3 Features, are in GeoJSON, and are returned from the `/collections/{collection_id}/items/{item_id}` endpoint.
+For WFS3-compliance, it is only **required** that `GET` be implemented for the WFS3 endpoints.  It is also only **required** that `GET` be implemented for the STAC endpoints.
```
m-mohr: Not a native speaker, but "be implemented" sounds strange to me. Should it be "is implemented"?

api-spec/api-spec.md
```diff
@@ -1,43 +1,45 @@
...
 
-In WFS 3 Collections are the sets of data that can be queried ([7.11](https://rawgit.com/opengeospatial/WFS_FES/master/docs/17-069.html#_feature_collections_metadata)), and each describes basic information about the geospatial dataset, like its name and description, as well as the spatial and temporal extents of all the data contained. [STAC collections](../collections-spec/README.md) contain this same information, along with other STAC specific fields and thus are compliant with both WFS Collections and STAC Collections and is returned from the `/collections/{collection_id}` endpoint.
+In WFS 3 Collections are the sets of data that can be queried ([7.11](https://rawgit.com/opengeospatial/WFS_FES/master/docs/17-069.html#_feature_collections_metadata)), and each describes basic information about the geospatial dataset, like its name and description, as well as the spatial and temporal extents of all the data contained. [STAC collections](../collection-spec/README.md) contain this same information, along with other STAC specific fields and thus are compliant with both WFS Collections and STAC Collections and is returned from the `/collections/{collection_id}` endpoint.
```
m-mohr: Please replace the rawgit link to the most current one WFS uses. rawgit is about to be shut down and I remember there was a discussion in the WFS repo that all references should be replaced ASAP.

api-spec/api-spec.md
```diff
@@ -1,43 +1,45 @@
...
+For WFS3-compliance, it is only **required** that `GET` be implemented for the WFS3 endpoints.  It is also only **required** that `GET` be implemented for the STAC endpoints. 
 
-### WFS3 Endpoints
+Since STAC adds additional filter parameters that may have much larger values, like `intersects`, it is **recommended** to also support `POST` for both the WFS3 and STAC endpoints that
+ accept filter parameters (e.g., `/collections/{collection_id}/items` and `/stac/search`). 
+ 
+It is **recommended** that these `Content-Type` options be supported when using POST:
 
-The core WFS 3 endpoints are shown below, with details provided in an [OpenAPI specification document](definitions/WFS3.yaml).
+1. POST with the `Content-Type` header set to `application/x-www-form-urlencoded` with the appropriate content body format.
```
m-mohr: Is there an easy way to transform between the three? Otherwise this feels maybe a bit much to ask for.

api-spec/api-spec.md
joshfix: I could see how it would sound weird @m-mohr but it is correct.

api-spec/api-spec.md
philvarner: same for my Scala/Akka implementation -- just a few lines of code to handle all three cases.

Python Flask example: https://www.flaskapi.org/api-guide/parsers/

api-spec/api-spec.md
philvarner: I think either is correct, so I changed it to `is` to sound less strange to non-native readers.

api-spec/api-spec.md
philvarner: Fixed.

api-spec/api-spec.md
philvarner: Fixed.

api-spec/README.md
philvarner: Changed it back.

api-spec/api-spec.md
m-mohr: Not yet, sorry. I think the correct link would be http://docs.opengeospatial.org/DRAFTS/17-069r1.html#_feature_collections
The one you linked is the source and has not included the requirements etc.

api-spec/api-spec.md
m-mohr: Okay, great.

joshfix:
The Filter Parameters table shows limit and page typed as `number`.  Should we change those to explicitly be `integer`?

m-mohr:
> The Filter Parameters table shows limit and page typed as `number`. Should we change those to explicitly be `integer`?

Yes, sure!

philvarner:
What's supposed to be the difference between the api-spec/STAC.yaml and api-spec/openapi/STAC.yaml?

m-mohr:
> What's supposed to be the difference between the api-spec/STAC.yaml and api-spec/openapi/STAC.yaml?

STAC.yaml is generatred from openapi/STAC.yaml (and others like WFS). I.e. openapi/STAC.yaml is a fragment, which only includes the core STAC API without extensions and external input such as WFS.

philvarner:
> > What's supposed to be the difference between the api-spec/STAC.yaml and api-spec/openapi/STAC.yaml?
> 
> STAC.yaml is generatred from openapi/STAC.yaml (and others like WFS). I.e. openapi/STAC.yaml is a fragment, which only includes the core STAC API without extensions and external input such as WFS.

Ah, understand that in the documentation now.

philvarner:
This PR is ready for merge pending approval.

api-spec/api-spec.md
```diff
@@ -1,43 +1,45 @@
...
+In WFS 3 Features are the individual records within a Collection and are provided in GeoJSON format. [STAC Items](../item-spec/README.md) are analogous to WFS 3 Features, are in GeoJSON, and are returned from the `/collections/{collection_id}/items/{item_id}` endpoint.
+
+## HTTP Request Methods and Content Types
 
-In WFS 3 Features are the individual records within a Collection and are provided in GeoJSON format. [STAC Items](../item-spec/README.md) are analagous to WFS 3 Features, are in GeoJSON, and are returned from the `/collections/{collection_id}/items/{item_id}` endpoint.
+For WFS3-compliance, it is **required** only that `GET` is implemented for the WFS3 endpoints.  Also, it is **required** only that `GET` is implemented for the STAC endpoints. 
 
-### WFS3 Endpoints
+Since STAC adds additional filter parameters that may have much larger values, like `intersects`, it is **recommended** to also support `POST` for both the WFS3 and STAC endpoints that
+ accept filter parameters (e.g., `/collections/{collection_id}/items` and `/stac/search`).
```
joshfix: I think here we do not want to recommend supporting POST for the collections endpoints as this conflicts with the transactions extension: https://github.com/radiantearth/stac-spec/tree/master/api-spec/extensions/transaction
I believe we want to recommend that for more complex searches that require POST to exclusively use the `/stac/search` endpoint.

m-mohr: Agreed.

philvarner: I agree with that as to not conflict with the Transactions Ext as it is currently defined.  

However, I think we should reconsider whether or not Transactions should extend the behavior of the paths defined in WFS 3.  For example, what if an existing WFS 3 implementation has already implemented the GET and POST form-data operations on `/collections/{collectionID}/items`?  I feel that if we want to provide extensions/behaviors on top of the ones defined by WFS 3 that we should duplicate the endpoints, such that, for example, we have  `/collections/{collectionID}/items` that implements only the WFS 3 semantics, and `/stac/collections/{collectionID}/items` that implements our own semantics (e.g., the transactions POST behavior)

joshfix: So originally, before it was an actual extension, I used PUT, PATCH, and DELETE to handle creating, updating, and deleting items against the dynamic api endpoint.  I think it was during the second sprint it was decided that transactions should be executed via the collections endpoint.  I'm not 100% certain why.  @m-mohr is there a WFS-T3 spec that defines transactions against the collections endpoint?  We also have our own catalog endpoints.  Would that make more sense?

philvarner: I don't think there's as much conflict as I originally thought.  I think we just needed to be explicit about the use of the `application/json` .  See the changes in https://github.com/radiantearth/stac-spec/pull/489/commits/e7dbfaaf486e783b949f1226dc42e9edf7a8e764

m-mohr: @joshfix Not yet, as far as I know, but there will probably be a WFS-T extension in the future that we could adopt.

Fine with the changes by @philvarner. Approved (again).

Issue #490, opened 2019-06-11
Modify fields extension object structure


joshfix:
Using the WFS 2 spec as a reference, my original dynamic STAC implementation allowed users to provide a `propertyName` array of strings as a request parameter.  The returned items would only contain the json fields defined in that array.  Since that time, a `fields` extension was introduced that expanded the capability to allow users to not only specify a list of fields to include, but alternatively a list of fields to exclude.

The extension however uses a seemingly over-complex structure:

```
{
    "fields": [
        {
            "include": ["id", "geometry", "properties.eo:cloud_cover"]
        }
    ]
}
```

The `fields` array can also contain an object with an "exclude" key.  The way the object is designed seems to imply that you could have multiple includes and multiple excludes objects, eg:

```
{ "fields": [
    {"include":  ["id", "geometry", "properties.eo:cloud_cover"]},
    {"include": ["other_stuff"]},
    ...
]}
```
It seems this could be greatly simplified by defining the fields request object to contain a single `include` field and a single `exclude` field, where the values are string arrays:

```
{ 
    "fields": 
        "include": ["id", "geometry"]
}
```
or
```
{ 
    "fields": 
        "exclude": ["links"]
}
```

In addition, I would like to propose a syntax that would make GET requests a bit cleaner by using parameter names with dot notation rather than including a JSON string.  The GET request would look like this:

```
/stac/search?fields.include=id,geometry
```
or
```
/stac/search?fields.exclude=links
```

If the conversation gets overly complex or contested, I can separate this out into a separate issue.

joshfix:
The [OpenAPI fragment](https://github.com/radiantearth/stac-spec/blob/master/api-spec/extensions/fields/fields.fragment.yaml) actually seems to define the structure that I proposed in this issue, so I'm hoping the examples were just constructed incorrectly in the README. 
 @matthewhanson where you the original author of this?  Can you confirm?

joshfix:
Created PR: https://github.com/radiantearth/stac-spec/pull/491

joshfix:
fixed example with https://github.com/radiantearth/stac-spec/pull/493.  will address query api GET parameters separately.

PR #491, opened 2019-06-12
Fix fields extension examples; introduce dot notation parameters for GET (#490)

joshfix:
The `fields` api extension readme had example requests that did not align with the spec yaml.  The examples were updated.  In addition, a new optional syntax was introduced for GET requests.  Instead of having to submit a JSON string as a value, eg:
```
GET /stac/search?fields={"include":["id","geometry","properties.eo:cloud_cover"]}
```
(along with additional encoding), users will be allowed to utilize the dot notation equivalents as request parameters, eg:
```
GET /stac/search?fields.include=id,geometry,properties.eo:cloud_Cover
```

Signed-off-by: Josh Fix <josh@federal.planet.com>

### DIFF
--- a/api-spec/extensions/fields/README.md
+++ b/api-spec/extensions/fields/README.md
@@ -2,20 +2,20 @@
 
 **Extension [Maturity Classification](../../../extensions/README.md#extension-maturity): Pilot**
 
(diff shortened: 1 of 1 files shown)


m-mohr:
You are mixing a "bug fix" with a new feature here. I'd split this. I'm fine with the fix and would accept it. For the new feature (dot notation) I'd like to hear other API devs input first. I'm not implementing it as of now and therefore would not be comfortable with deciding on it.

PR #500, opened 2019-06-17
WIP: Query Extensions framework and stacql naming

philvarner:
This isn't really in a form for merging, it's more a strawman for discussion. A few of the points:
1. GET and POST JSON endpoints will likely implement different extensions, where, in general, GET will use simpler, string-oriented values and POST will implement arbitrarily complex query languages.
2. Language around reserved parameter and attribute names, e.g., "query".  The key point here is that a core API-compliant implementation may add additional behaviors, but those must not conflict with behaviors defined in extensions
3. GraphQL extension language
4. Examples of GET string-based sort, fields, and CQL

**PR Checklist:**


api-spec/api-spec.md
```diff
@@ -50,3 +59,198 @@ The `/stac/search` endpoint is similar to the `items` endpoint in WFS3 in that i
...
+  "query": {
+    "predicates": [
+      {
+        "name": "eo:cloud_cover",
+        "gte": 0,
+        "lte": 10
+      }
+    ]
+  }
+  ...
```
joshfix: So this implies that a predicate key could be either "name" or any one of the operators?  It seems to me, if we wanted to keep a syntax such as this, that the predicate should have 3 keys: `field`, `operator` and `value`.  If you wanted items where a field were in some range (cloud cover between 0 and 10) you'd have two predicates, one for less than and one for greater than.  The one thing I still don't like about the syntax is the implicit AN and no support for OR.

api-spec/api-spec.md
```diff
@@ -50,3 +59,198 @@ The `/stac/search` endpoint is similar to the `items` endpoint in WFS3 in that i
...
+
+Query Extensions are **recommended** to use attribute names qualified from the root of Item, rather than Item Properties.
+
+Example:
+
+```
+{ 
+  ...
+  "query_type": "cql",
+  "query": {
```
joshfix: How would this work for GET?  It seems like you'd still have to provide a json value in the query parameter, eg `?query={\"predicate\":\"xxx\"}` or use the dot notation `?query.predicate=xxx`.

m-mohr:
I generally feel that the STAC spec is getting quite complex in the last weeks. It's not just this PR, but also a general direction, I feel. When I'm speaking with people they always say it's good that we are taking a simple/slim approach, but nowadays it feels we are more and more making it "OGC/ISO-complex". We shouldn't do that otherwise we loose our unique selling point.

Allowing multiple query languages makes it hard to implement clients for searching as you'd need to support all query languages, otherwise you loose the interoperability any clients will only be supporting searching on a subset of servers. I feel we should stick to a single, standardized query language that has a good ecosystem to support implementations. GraphQL, CQL or anything else. It should cover most of the use cases and be our primary tool for searching. As a compromise, I'm fine to allow a simpler query language for GET and having a fully featured one for POST. In the best case they can be easily translated, but that's a bonus, but people could also simply opt to not implement GET. 

The GET proposal (something like `eo:cloud_cover lt 50 AND ids in 1,2,3`) from yesterday seemed quite easy (split at "and" and "or", then split by space into name, comparator and value), but is probably getting a bit harder to parse once you allow strings. Parsing `collectionid eq "my eq collection"` makes things complicated due to keywords and/or spaces in the string. We just need to make sure it's really as easy to implement and use as it seems.

joshfix:
My main drive is to keep things simple, but you have to consider both the simplicity of the user and the simplicity of the implementor.  It does make things very easy if you say that we're not going to support complex GET requests.  The concern I have is that we are making a lot of the request objects unnecessarily complex.  When you don't support any type of query language, sort filters, or field filters for GET requests, I don't think that will make users happy.  Or if you make them write JSON as a URL parameter for something that should otherwise be a simple request.  Writing OGC filter expressions as XML in GET parameters has infuriated countless developers in the past.

I think the simplest approach is to use the same syntax for GET and POST in most cases.  Fields seems easy enough.  We could either use dot notation to access the object fields as parameters (`?fields.include=xxx&fields.exclude=yyy`) or simply use a single fields array with the `-` prefixes (`?fields=xxx,-yyy`). 

The query language is a bit harder if everybody mostly wants a highly structured and easily parsed JSON object as the request.  And if that's the case, I would strongly desire some alternative (potentially optional) for GET.  CQL has worked great for me, but even a subset of CQL where you only need to support the current operators we have defined, but you can write your queries like `?query=eo:cloud_cover < .1`. 

I agree with @m-mohr and @cholmes that we don't want to make clients have to implement multiple query languages to support any given STAC service.  It would be good to get WFS3 and STAC both to choose a default language.  Where it makes sense (the intersects geometry) we should not support GET parameters.  I believe we should make our best effort to define request objects so that they support GET whenever possible.

api-spec/api-spec.md
philvarner: lol, this is definitely not what i meant to write here.  My fault for doing this late night.  Will update it.

philvarner: I feel strongly that we have a base query extension that is very simple to implement, and I think that means only supporting `AND`.  Lowest barrier to entry possible.

philvarner:
> My main drive is to keep things simple, but you have to consider both the simplicity of the user and the simplicity of the implementor

💯 

This is the primary driver behind wanting both a simple, verbose query language (like the existing Query Extension defines) and a fully-expressive, but more complex implementation like CQL.  Allowing only one either puts a significant burden on the implementor (meaning that they're likely to just do something else that's arbitrary and custom) or prevents queries that are expressive enough for many use cases.

philvarner:
> I generally feel that the STAC spec is getting quite complex / over engineered in the last weeks.

I think part of this is that we have actual use cases driving the development, rather than contrived examples.  I completely understand the desire for simplicity, but that can easily be something that's not useful enough.  If it's not useful enough, implementors will chose arbitrary/custom behaviors, and then the standardization is no longer valuable.  With the few PRs that I've submitted in the last week (and whatever becomes of this one), I've satisfied all the behaviors of my existing custom API -- the fact that I'm rewriting my existing service to the STAC API is evidence of that. 

I think another thing that's contributing to the perception of increasing complexity (particularly with the API) is that we're now explicitly describing behavior that was either implicit or ill-defined before, and for which implementers would have made arbitrary and probably differing interpretations. While it's more complex, it's also necessary if we're going to have interoperable implemenations.

philvarner:
Related to GraphQL: I think an extension for that is likely out past 0.9, but I think it's useful to think about how we'd incorporate it.  It feels a little shoe-horned in to the `/stac/search` endpoint, so maybe just putting it at a completely different endpoint (`/stac/graphql`?) would be a better option.

metasim:
Mostly a lurker here, so take my comment with a grain of salt...

> I think the simplest approach is to use the same syntax for GET and POST in most cases. Fields seems easy enough. We could either use dot notation to access the object fields as parameters (?fields.include=xxx&fields.exclude=yyy) or simply use a single fields array with the - prefixes (?fields=xxx,-yyy).

I'd argue for not necessarily tying the syntaxes, considering the following:

1. As has already been raised in this conversation, the `GET` syntax should be super easy to implement, but not aim for being comprehensive; just implement enough to provide "conservative" results (i.e. not as refined as you might want, maybe requiring additional client-side filtering). This would be the required syntax.
2. Allow, via query extensions, `POST` queries with via more complex/standardized query languages, selected/specified by `content-type`.  For example `application/x-cql` or `content-type: application/json; dialect=elasticsearch`
3. The default `POST` implementation would be `content-type=application/x-www-form-urlencoded` with the same syntax as the `GET` form.

cholmes:
A few thoughts on simplicity vs complexity:

* I am leaning more towards 'breaking out' the STAC API in some way, so the core STAC is just our Item, Catalog and Collection. Indeed I think we could go to '1.0' pretty soon with that, while for the API I'd much prefer to help WFS 3 shake out a bit more and get to 1.0.
* I'm ok with a bit more complexity than the core of WFS 3 in STAC API, because we can point people at the core spec. They can implement statically, or also implement catalogs/collections dynamically (but not search), and another STAC API can ingest them. 
* In an ideal world WFS 3 has already got great recommendations for Query and we can just adopt. In lieu of that I'm quite happy to have these discussions here that spiral out the complexity a bit, especially since they are based on real-world implementations. Ultimately I'd like us to get to a single required Query language for STAC (even if WFS3 doesn't require it), that strikes a good balance between ease of implementation and power. More complex query languages should be WFS extensions, and we can recommend another 1-2 ones we recommend / that our community often implements. 
* GET vs POST for geospatial is inherently tricky, since people have large geometries that hit the GET limit, so it's challenging to make that the one required path. Though we could consider like a simple geometry reference or something? I agree that a good simple approach is to keep get + post the same.
* I do like the direction of an intuitive URL GET structure that is easy for a naive person to play with, and it seems like a group overall we like that. 
* So I think it's ok to let this conversation expand out a good bit, in part because WFS 3 has not really done that yet - they're a bit too focused on the core imho. And I think you can 'miss' on the core if you don't actually build out the extensions first, to know that the core can support them. But ultimately I think we should move many of our more 'complex' constructs to be WFS extensions, that live outside 'STAC', and we just have a paragraph in the STAC spec that says 'we like these set of WFS extensions, and if you use them then STAC clients will likely take advantage of them'. Indeed my medium term hope is that we can make the STAC API spec very slim - that it would refer to WFS 3 and list a set of extensions and best practices, along with the appropriate STAC content types.
* But it's a fair point on managing the perception of STAC 'growing more complex'. I can try to do some blog posts addressing it. I do think us starting to make PR's for extensions to WFS will help, instead of just adding them here. But I'm happy for us to have a collaboration space for a bit. Let's just be sure that the 0.8.0 API spec mostly feels like a tighter version than the previous (and I think a lot of the PR's have been that - tightening the http response codes, making implicit behavior explicit, cleaning up examples, etc), and not a 'bigger' thing.

philvarner:
I just updated the description in a way that I think strikes a good balance between overly-nested objects and an easy GET syntax.  Also introduced the name "STACQL" for the custom query language defined in the existing Query extension.

philvarner:
> * I am leaning more towards 'breaking out' the STAC API in some way, so the core STAC is just our Item, Catalog and Collection. Indeed I think we could go to '1.0' pretty soon with that, while for the API I'd much prefer to help WFS 3 shake out a bit more and get to 1.0.

I agree with this -- I was thinking how the API and item definitions aren't really that coupled, and the main driver behind the API is that OGC - Features Part 1 API is (intentionally) so barebones.

matthewhanson:
to be superseded by OAF ongoing work, would need new PR depending on that outcome.

philvarner:
Thanks @matthewhanson -- you literally closed it a minute before I was going to

Issue #501, opened 2019-06-18
Data/label version tag


nrweir:
As we were working on the label extension at Sprint 4, we discussed the idea of creating a `"label:version"` tag to identify label files where multiple versions existed. In our minds we saw two scenarios where this could happen:

1. A set of labels are iteratively updated based on the same source image, for example after validation or due to incomplete labeling during a first pass (a common phenomenon for OSM)
2. Multiple labelers may independently create complete label sets based on the same source data, which are then aggregated somehow into a single validated version.

In both cases, it would be great to have a means to identify different versions of an asset or item which has some intrinsic relationship to other versions of the same object. This was discussed a bit in #362 and it was determined it shouldn't be restricted to the label extension, as it may be appropriate for other item/asset types.

A few questions related to this:
1. Is including something like this a good idea in general?
2. Should it be at the Asset or Item level? I think the main reason to have it at the asset level would be if a single item could potentially include different versions of the same label - should that type of structure be supported?
3. What should it be called? `"version"` could get confusing with `"stac_version"`.

m-mohr:
Adding a version number to the Item seems reasonable. We have it similarly in the collections. There, they are called `version` and `stac_version`, too, and live alongside each other quite well. Currently, there's no `stac_version` in Items anyway.

matthewhanson:
Versioning is something that is definitely reasonable, but it's different than other properties so I don't think it's as simple as just adding a version field.   In fact, I think the collection version field suffers from the same issue.

A version is *not* a separate record.  It is a different version of the same record. 

As a user I typically want the latest version of any specific Item (or Collection).
However, for data provenance I also want to be able to link to any previous version of an Item or Collection. 
But when I search I do *not* want every version, I only want the latest.  

So the issue here is that if you want them to be separate records then the version has to appear in the ID, but then that means they are separate records that will all get returned in a search.

So I think version is quite distinct than other metadata fields and deserves special consideration.

nrweir:
One other potentially relevant use case for labels, though this may be less relevant for non-label formats, would be "parallel" versions. For example, a common part of crowdsourced labeling pipelines is to have multiple people generate "preliminary" labels for a given piece of imagery, which are then merged by an expert validator to generate the final version. In this case you wouldn't _necessarily_ just want the most recent one. I recognize this is probably something of an edge case, though, and not necessarily something we need to structure version tagging around.

m-mohr:
[...]

PR #502, opened 2019-06-18
Cleaning up examples and removing label:version

nrweir:
1. Renamed various directories and files in example to make them a bit more straightforward
2. Updated asset example link in the label extension README.md
3. Added links to the other label examples in README.md
4. Removed `"label:version"` per the discussion in #362 and created an issue to discuss its inclusion in a broader scope: #501


extensions/label/examples/multidataset/catalog.json
```diff
@@ -0,0 +1,25 @@
+{
+    "id": "label_extension_demo",
+    "title": "label extension demo",
+    "description": "Sample ML training data labels in the STAC format",
+    "stac_version": "0.7",
```
m-mohr: Example should be 0.7.0, otherwise the validation would fail.

PR #505, opened 2019-06-19
Update api spec STAC.yaml file via generate-all

philvarner:


api-spec/STAC.yaml
```diff
@@ -738,7 +738,7 @@ components:
           $ref: '#/components/schemas/time'
     intersectsFilter:
       type: object
-      description: Only returns items that intersect with the provided geometry.
+      description: Only returns items that intersect with the provided polygon.
```
philvarner: I just ran `npm run generate-all`, and one recent change hadn't been picked up into this file yet.

PR #506, opened 2019-06-19
Transaction Extension Status Codes are more REST conventional

philvarner:


api-spec/extensions/transaction/transaction.fragment.yaml
```diff
@@ -1,174 +1,211 @@
...
-      description: allows for partial collections of metadata fields
-      additionalProperties: true
-      properties:
-        datetime:
-          $ref: '#/components/schemas/time'
-tags:
-  - name: Extensions
-    description: >-
-      STAC-specific operations to add, remove, and edit items within WFS3
+openapi: 3.0.1
```
philvarner: turn on Hide Whitespace Changes via the gear icon above to get a real diff

api-spec/extensions/transaction/transaction.fragment.yaml
```diff
@@ -1,174 +1,211 @@
...
+        - $ref: '#/components/parameters/collectionId'
+      requestBody:
+        content:
+          application/json:
+            schema:
+                oneOf:
+                 - $ref: '#/components/schemas/item'
+                 - $ref: '#/components/schemas/itemCollection'
+      responses:
+        '201':
```
philvarner: 201 Created is more REST conventional than 200 for creating new resources

api-spec/extensions/transaction/transaction.fragment.yaml
```diff
@@ -1,174 +1,211 @@
...
+          description: Status of the create request.
+          content:
+            application/geo+json:
+              schema:
+                type: string
+            text/html:
+              schema:
+                type: string
+        '400':
```
philvarner: For all of our API endpoints, I would like us to be explicit about what status codes are expected for which behaviors.

api-spec/extensions/transaction/transaction.fragment.yaml
```diff
@@ -1,174 +1,211 @@
...
+          content:
+            text/html:
+              schema:
+                type: string
+            application/json:
+              schema:
+                type: string
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '404':
```
philvarner: Explicitly state 404 for all operations if the resource doesn't exist.

api-spec/extensions/transaction/transaction.fragment.yaml
```diff
@@ -1,174 +1,211 @@
...
+        - Extensions
+      parameters:
+        - $ref: '#/components/parameters/collectionId'
+        - $ref: '#/components/parameters/featureId'
+      responses:
+        '204':
+          description: Status of the delete request.
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '404':
```
philvarner: Explicitly state 404 for delete on a resource that doesn't exist.  Some APIs return a 2xx response if it's already deleted, as the desired non-existence is fulfulled, but that's not the standard convention.

api-spec/extensions/transaction/transaction.fragment.yaml
```diff
@@ -1,174 +1,211 @@
...
+      summary: delete an existing feature by Id
+      description: Use this method to delete an existing feature.
+      operationId: deleteFeature
+      tags:
+        - Extensions
+      parameters:
+        - $ref: '#/components/parameters/collectionId'
+        - $ref: '#/components/parameters/featureId'
+      responses:
+        '204':
```
philvarner: 204 No Content instead of 200 is more REST conventional

api-spec/extensions/transaction/transaction.fragment.yaml
m-mohr: At the same time I would expect a Location header where it was created. Should we add that?

api-spec/extensions/transaction/transaction.fragment.yaml
```diff
@@ -1,174 +1,211 @@
...
+                type: string
+            application/json:
+              schema:
+                type: string
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '404':
+          $ref: '#/components/responses/NotFound'
+        '503':
+          $ref: '#/components/responses/InternalServerError'
```
m-mohr: 503 is Service Unavailable. InternalServerError is 500. Maybe just go with 5XX (yes, that's actually allowed) to allow all 500-599 errors. Many can occur on servers and should send a harmonized error response.

philvarner: ha, my typo.  5XX sounds great -- I mostly wanted to clearly specify what what the "you submitted an invalid request" and "hit a bug/data store down/etc" cases.  Updated.

api-spec/extensions/transaction/transaction.fragment.yaml
philvarner: Added.  I thought about that, but I didn't know how to write it in OpenAPI, so I had just left it out.

PR #508, opened 2019-06-20
WIP: Update API Fields Extension with more precise semantics and behavior

philvarner:
I started working on implementing the API Fields extension, and ran into some concerns around some ambiguous requirements.  

For example, 
```
If both `include` and `exclude` attributes are specified, `include` takes precedence.
```

This could be interpreted to mean that if I include `properties` and exclude `properties.datetime`, that the latter will still be included because the parent `properties` was explicitly included.

There was no mention that excluding certain properties could result in invalid Item/Feature entities. 

There was no mention of what the baseline (no includes or excludes) should look like -- the core default of "everything" is probably not good, as that could be a huge object, but neither is the GraphQL convention of including nothing by default, since that would always result in an invalid Item/Feature. 

Also added explicit language around the difference between the JSON object representation and the query string parameter representation (CSV using `-` for excludes and no prefix for includes)


api-spec/extensions/fields/README.md
```diff
@@ -2,29 +2,96 @@
...
+Return baseline fields.  This **must** return valid STAC Item entities. 
+
+```json
+{
+  "fields": {
+  }
+}
+```
+
+This has a similar effect as an empty object for `fields`, but it is up to the discretion of the implementation
```
m-mohr: The example below would remove fields added to the root level (by extensions, custom fields or later versions).

api-spec/extensions/fields/README.md
```diff
@@ -2,29 +2,96 @@
 
 **Extension [Maturity Classification](../../../extensions/README.md#extension-maturity): Pilot**
```
m-mohr: Is all the semantics below - whenever possible - also modeled in the OpenAPI? Especially the content types?

philvarner: I think they are, though it's very limited.  This behavior follows the description in one of my other PRs of using simple string parameters for GET or POST form, and additional attributes into the JSON body for POST application/json.

api-spec/extensions/fields/README.md
```diff
@@ -2,29 +2,96 @@
...
 
-The STAC search endpoint, `/stac/search`, by default returns entire Items. The fields API extension adds a new parameter, `fields` that allows the user to define fields in the items returned to be returned or excluded. If both include and exclude are specified, include takes precedence.
+By default, the STAC search endpoint `/stac/search` returns all attributes of each Item, as there is no way to specify exactly those attributes that should be returned. The Fields API extension adds a new option that allows the user to explicitly define which fields in an Item should be included or excluded in the response. 
+
+When calling `/stac/search` using POST with`Content-Type: application/json`, this extension adds an attribute `fields` with an object value to the core JSON search request body. The `fields` object contains two attributes with string array values, `include` and `exclude`.
```
philvarner: this originally incorrectly said "GET or POST"

api-spec/extensions/fields/README.md
```diff
@@ -4,9 +4,9 @@
...
 
-When calling `/stac/search` with GET or POST with `Content-Type: application/x-www-form-urlencoded` or `Content-Type: multipart/form-data`, the semantics are the same, except the syntax is a single parameter `fields` with a comma-separated list of attribute names, where `exclude` values are those prefixed by a `-` and `include` values are those with no prefix, e.g., `-geometry`, or `id,properties,-properties.proj:geometry`.
+When calling `/stac/search` using GET or POST with `Content-Type: application/x-www-form-urlencoded` or `Content-Type: multipart/form-data`, the semantics are the same, except the syntax is a single parameter `fields` with a comma-separated list of attribute names, where `exclude` values are those prefixed by a `-` and `include` values are those with no prefix, e.g., `-geometry`, or `id,properties,-properties.proj:geometry`.
```
joshfix: Perfect thanks for the update!

api-spec/extensions/fields/README.md
philvarner: correct.  Any fields added as `required` on Item should be added to this default list (though I doubt we'll add any more required fields at this point)

philvarner: The point here is that, by default, you're returning a valid Item, not necessarily that it has all of the extension fields, etc.

Issue #472, opened 2019-06-05
Update Collections and API to WFS 3 latest


cholmes:
WFS 3 changed their collection definition a bit. We should update to align.

m-mohr:
I'm taking this over, but will wait for an actual implementation into the spec until at least the OGC Hackathon was held.

m-mohr:
Backporting WFS to STAC makes sense only around July as it's when they plan to release the next release candidate. Is that early enough for 0.8 or do we delay it for 0.9? @cholmes

m-mohr:
see also https://github.com/opengeospatial/WFS_FES/issues/243

m-mohr:
Draft 2 has been released 5 days ago: https://github.com/opengeospatial/WFS_FES/releases/tag/1.0.0-draft.2 - will make a PR very soon.

m-mohr:
I saw that there are two new (optional) properties in their collections: 
* `itemType`: indicator about the type of the items in the collection (the default value is 'feature').
* `crs`: the list of coordinate reference systems (OGC URls) supported by the service (there's an additional crs in the spatial extent, too)

Those seem somewhat unrelated to STAC and backporting them to STAC Collections feels wrong. Should we ignore them?

Also, they have new optional properties in the links:
* `hreflang` language of the referenced document
* `length` not sure yet what this is about. length of the referenced document?

Should we also ignore them? I think not providing optional properties would be fine.

Other than that, we only need to change the structure of `extent` in STAC collection. 

There are probably some more changes in the API.

cc @cholmes

m-mohr:
PR: #525 to update the API and backport `extent` to STAC collections.

PR #509, opened 2019-06-21
add WFS 3 /api endpoint to docs, and reformat some tables

philvarner:


api-spec/api-spec.md
```diff
@@ -22,11 +22,12 @@ The WFS 3 and STAC APIs follow a RESTful model.  A core principal of this is the
...
 
-| Endpoint      | Returns          | Description        |
-| ------------ | ------------- | ---------------------- |
-| / | JSON        | Landing page, links to API capabilities |
-| /conformance | JSON | Info about standards the API conforms to       |
-| /collections | Collections | List of Collections contained in the catalog |
+| Endpoint     | Returns       | Description |
+| ------------ | ------------- | ----------- |
+| /            | JSON          | Landing page, links to API capabilities |
+| /api         | JSON          | API definition |
```
philvarner: this endpoint was missing from this doc, but in the WFS 3 spec

philvarner:
I'm looking at https://rawcdn.githack.com/opengeospatial/WFS_FES/3.0.0-draft.1/docs/17-069.html#_api_definition_2 
`The server SHALL support the HTTP GET operation at the path /api.`

conformance was already there, i just cleaned up the formatting

joshfix:
> There's no /api in https://stacspec.org/STAC-api.html and it seems it's not required in WFS, so we may leave it out. Let's wait a little with this until we know more about OGC API Commons. I'll take care to update with #472
> 
> Adding conformance though is fine, so maybe remove /api and we'll update to include /api once I backport WFS and if it's included there.

I went ahead and approved as it seems to be legit.  Back to you @m-mohr

m-mohr:
I just discussed this with Clemens, one of the authors of the WFS spec. 
/api is indeed required by the WFS spec, but is not (and probably won't be in the future) included in the OpenAPI spec of WFS. So, I'm feeling that it is confusing if /api is explained here, but not mentioned anywhere ein the OpenAPI documentation.

That said, now I'd even go further and add explanation here that /api is required by the WFS specification (with link to the requirement) and then explain here that it is not part of the OpenAPI spec, but still needs to be implemented.

joshfix:
That seems super confusing.  So WFS is publishing an OpenAPI spec for WFS, but not including required endpoints?  Do we really need to explain WFS documentation in our STAC documentation, or can we just require the endpoint?

m-mohr:
Yes, the reason was that /api returns the API document itself, which you are looking at anyway. I made the point that requiring an endpoint that is not in the OpenAPI file is confusing, but we didn't get to a consensus. Maybe open an issue in the WFS issue tracker so that they know I'm not the only one that finds this super confusing.

philvarner:
Added language around /api missing.  will file issue with WFS 3.

philvarner:
This is the issue that caused it to be removed https://github.com/opengeospatial/WFS_FES/issues/7

philvarner:
Filed https://github.com/opengeospatial/WFS_FES/issues/236

It as a requirement to implement the /api endpoint according to the table in the Abstract section of https://rawcdn.githack.com/opengeospatial/WFS_FES/3.0.0-draft.1/docs/17-069.html and 7.3. API definition. However, this is endpoint is not present in the openapi.yaml documentation. I see that it was removed as a result of #7 , however, as an implementer, I find this confusing. Just adding a definition of the endpoint and that it returns an entity of type application/vnd.oai.openapi would make this clear.

m-mohr:
Thanks for finding that issue and adding a short sentence about the absence of /api in the OpenAPI spec. Approved and merged.

PR #513, opened 2019-06-24
Sort Extension: simple format for GET

philvarner:
1. added a slightly different, non-JSON format for the GET parameter case.  @joshfix 
2. changed the description so that field from the root of Item rather than properties.  This is so that a user can provide a stable sort over `properties.created ascending, id` (ascending or descending).  Without the id as the secondary sort, items with the same created timestamp would result in a non-stable sort, which may give users incorrect results with trying to page through an entire result set. 

**PR Checklist:**


joshfix:
I like the GET support!  :)  

I have no idea if this is a good idea or bad idea, but what about making the GET values use a similar syntax as the proposed fields values, where a `+` denotes ascending and a `-` denotes descending:

`GET /stac/search?sort=+properties.created,-id&fields=+properties.created,+xxx`

philvarner:
As I wrote up Fields, it's only nothing or `-`, which I think is fine for sort also ==> ascending is the default, with `-` prefix meaning descending.

cholmes:
Moving to 0.9.0 - as there's more thinking to do on this. Sean will sound in with some more thoughts on doing things in headers, and we want to try to align with WFS.

api-spec/extensions/sort/README.md
```diff
@@ -2,7 +2,29 @@
...
+
+Two values for direction are supported: "asc" (ascending) or "desc" (descending). If the direction is not specified, the value is ascending. The `sort` value is an array, so multiple sort fields can be defined which will be used to sort the data in the order provided (e.g., first by `datetime`, then by `eo:cloud_cover`).
+
+## GET or POST Form
+
+When calling `/stac/search` using GET or POST with `Content-Type: application/x-www-form-urlencoded` or `Content-Type: multipart/form-data`, the semantics are the same, except the syntax is a single parameter `sort` with a comma-separated list of "<name>|<direction>" definitions.  The sort order must be specified.
+
+Examples of `sort` parameter:
+
+    GET /stac/search?sort=created|asc,id|desc
```
tomkralidis: What about replacing `|` with `;` ?  Like `GET /stac/search?sort=eo:cloud_cover;desc`.  `|` may be ambiguous (expressing this or that property, say)

api-spec/extensions/sort/README.md
```diff
@@ -15,4 +37,21 @@ The STAC search endpoint, `/stac/search`, by default returns results in descendi
 }
 ```
 
-where <direction> is either "asc" (ascending) or "desc" (descending). The `sort` value is an array, so multiple sort fields can be defined which will be used to sort the data in the order provided (e.g., first by `datetime`, then by `eo:cloud_cover`).
+```json
+{
+    "sort": [
```
tomkralidis: Borrowing from OGC approaches, what about:

```json
    "sort": [
        {
            "property": "created",
            "order": "asc"
        }
```

tomkralidis:
Note relevant conversation in https://github.com/opengeospatial/ogcapi-features/issues/157

api-spec/extensions/sort/README.md
philvarner: @tomkralidis can you point me to the documentation for those OGC approaches?

tomkralidis: @philvarner 

- from CSW 2.0.2: http://portal.opengeospatial.org/files/?artifact_id=20555 (Table 65)
- from CSW 3.0: http://docs.opengeospatial.org/is/12-176r7/12-176r7.html (Table 22)

matthewhanson:
@philvarner   Looks like this might need some updates, thought colon was going to be avoided in favor of semi-colon?

philvarner:
I think I have all the necessary changes in now @joshfix @tomkralidis @matthewhanson

api-spec/extensions/sort/README.md
```diff
@@ -10,18 +10,15 @@ Two values for direction are supported: "asc" (ascending) or "desc" (descending)
 
 ## GET or POST Form
 
-When calling `/stac/search` using GET or POST with `Content-Type: application/x-www-form-urlencoded` or `Content-Type: multipart/form-data`, the semantics are the same, except the syntax is a single parameter `sortby` with a comma-separated list of "<name>:<direction>" definitions.  If no direction is specified, the default is ascending.
-
-The use of colons in extension names is known to be problematic with respect to this format.  However, there is a desire
-to remove the use of colons in extension names, and it is unlikely that a property will be named like `foo:desc`.
+When calling `/stac/search` using GET or POST with `Content-Type: application/x-www-form-urlencoded` or `Content-Type: multipart/form-data`, the semantics are the same, except the syntax is a single parameter `sortby` with a comma-separated list of "<name>|<direction>" definitions.  It is recommended that in implementations, direction be mandatory, and that an error should result from not specifying a direction.
```
m-mohr: `/stac/search` should be `/search`. Not sure why this is outdated.

philvarner: Did you already have this loaded in your browser? it's fixed in the current PR.

m-mohr: Can't remember, fixed now.

m-mohr:
I'm just wondering whether we should use an approach that can be modeled in OpenAPI. We can't properly describe the current behavior except for textual descriptions. I went through the OpenAPI specification and found we could model it properly this way:
```
{
  "name": "sortby",
  "in": "query",
  "schema": {
    "type": "object",
    "additionalProperties": {
      "type": "string",
      "enum": [
        "asc",
        "desc"
      ]
    }
  },
  "style": "deepObject"
}
```

`sortby[properties.datetime]=desc&sortby[id]=asc`

This translates to the following object, which could also be used in POST:
```
{
  "properties.datetime": "desc",
  "id": "asc"
}
```

PR #515, opened 2019-06-25
rename search extension found to matched per WFS3 terms

philvarner:
https://github.com/opengeospatial/WFS_FES/blob/980a445f6bc89cb7a2804feb59ca1505a901ebe1/core/openapi/schemas/featureCollectionGeoJSON.yaml

### DIFF
--- a/api-spec/extensions/search/README.md
+++ b/api-spec/extensions/search/README.md
@@ -21,7 +21,7 @@ search, for example, from calling the `/stac/search` API endpoint.
 | next         | string \| null  | **REQUIRED.** The value to set for the `next` query parameter in order to get the next page of results |
 | returned     | integer         | **REQUIRED*** The count of results returned by this response. equal to the cardinality of features array |
 | limit        | integer \| null | The maximum number of results to which the result was limited |
-| found        | integer         | The count of total number of results for this query, possibly estimated, particularly in the context of NoSQL data stores |
+| matched        | integer         | The count of total number of results that match for this query, possibly estimated, particularly in the context of NoSQL data stores |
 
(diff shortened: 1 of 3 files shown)


joshfix:
So they define those fields at the root level of the item collection and not inside of an object.  Do we want to completely follow suit and remove our `search:metadata` object and put those fields at the root of the item collection?

philvarner:
> So they define those fields at the root level of the item collection and not inside of an object. Do we want to completely follow suit and remove our `search:metadata` object and put those fields at the root of the item collection?

I think that would be a good idea, being that we have a guideline that nested JSON structures should be used for nested data structures rather than grouping related attributes.  

A separate discussion is if we actually name it the same, and if we start using lowerCamelCase names like WFS 3.

joshfix:
Ha, I overlooked the camel case naming there :)  I think the only thing that will make this community break away from WFS3 is camel case!

matthewhanson:
"matched" looks good to me.  I'm not too particularly fond of polluting the top level response with a bunch of fields that should be logically grouped, but this PR is just for renaming found to matched so approving.

We should convince WFS3 to use snake_case.

philvarner:
> We should convince WFS3 to use snake_case.

Or the other way round.  I'd worked exclusively in Java until a couple years ago, so lowerCamelCase was what we always used for JSON, too.  It's only been in working with Python in the last couple years that I'd even used snake_case. 

What was the motivation for using snake_case in this spec?

This issue around using camelCase (though mostly instead of kebab case) has a pretty good summary of all the sources that recommend camelCase:
https://github.com/json-api/json-api/issues/1255

joshfix:
@matthewhanson I'm fine with polluting or not polluting :)  Just curious how conformant we want to be with WFS3.  This structure seems rather important to be aligned with, whether it's us conforming to them or them conforming to us.

@philvarner I always use camelCase for JSON in my projects for the same reasons you've stated.  For STAC, I've simply followed suit and went along with snake case.  I didn't realize until fairly recently how strong of an anti-camelCase and anti-Java sentiment there is amongst non-Java devs.  I think the motivation behind using snake_case is simply that the majority of people come from more of a Python background.

PR #516, opened 2019-06-25
Clarify EO items platform, constellation, and instrument

philvarner:
**PR Checklist:**

- [X] API only: I have run `npm run generate-all`. to update the [generated OpenAPI files](https://github.com/radiantearth/stac-spec/blob/dev/api-spec/README.md#openapi-definitions).


extensions/eo/README.md
```diff
@@ -26,10 +26,10 @@ The exact metadata that would appear in a STAC Collection record will vary depen
 | Field Name       | Type                     | Description |
 | ---------------- | ------------------------ | ----------- |
 | eo:gsd           | number                   | **REQUIRED.** Ground Sample Distance at the sensor. |
-| eo:platform      | string                   | **REQUIRED.** Unique name of the specific platform the instrument is attached to. For satellites this would be the name of the satellite (e.g., landsat-8, sentinel-2A), whereas for drones this would be a unique name for the drone. |
-| eo:constellation | string                   | Name of the constellation that the platform belongs to. See below for details. |
```
philvarner: moved so all REQUIRED are at top

extensions/eo/README.md
```diff
@@ -48,12 +48,28 @@ PlanetScope Ortho Tile Product has an `eo:gsd` of 3.7 (or 4 if rounding), even t
...
 
-**eo:constellation** is the name of the group of satellites that have similar payloads and have their orbits arranged in a way to increase the temporal resolution of acquisitions of data with similar geometric and radiometric characteristics. Examples are the Sentinel-2 [constellation](https://www.esa.int/Our_Activities/Observing_the_Earth/Copernicus/Sentinel-2/Satellite_constellation), which has S2A and S2B and RapidEye. This field allows users to search for Sentinel-2 data, for example, without needing to specify which specific platform the data came from.
-
+**eo:platform** is the unique name of the specific platform the instrument is attached to. For satellites this would 
+be the name of the satellite, whereas for drones this would be a unique name for the drone.  It is recommended that 
+the value be lowercase with no spaces. Examples include `landsat-8` (Landsat-8), `sentinel-2a` and `sentinel-2b` 
+(Sentinel-2), `terra` and `aqua` (part of NASA EOS, carrying the MODIS instruments),
+ `mycorp-uav-034` (hypothetical drone name), and `WORLDVIEW02` (Maxar/DigitalGlobe WorldView-2).
```
m-mohr: Uppercased WORLDVIEW02, which is not following the recommendation.

extensions/eo/README.md
```diff
@@ -48,12 +48,28 @@ PlanetScope Ortho Tile Product has an `eo:gsd` of 3.7 (or 4 if rounding), even t
...
+ 
 **eo:instrument** is the name of the sensor used, although for Items which contain data from
 multiple sensors this could also name multiple sensors. For example, data from the Landsat-8
 platform is collected with the OLI sensor as well as the TIRS sensor, but the data is distributed
-together and commonly referred to as OLI_TIRS.
+together and commonly referred to as OLI_TIRS. Examples include `OLI_TIRS` (Landsat-8), `MSI` (Sentinel-2), 
+`ASTER` (Terra), and `MODIS` (Terra and Aqua).
```
m-mohr: If instrument and platform should be lowercased, why are all instruments uppercased?

philvarner:
@m-mohr I'm changing them to all be lowercase.  I was reluctant to force instrument lowercase since all of them seem to be acronyms, but there should be not difference between that and a capitalized stylization.

extensions/sar/README.md
```diff
@@ -41,13 +41,28 @@ The exact metadata that would appear in a STAC Collection record will vary depen
...
+**sar:constellation** is the name of a logical collection one or more platforms that have similar payloads and have 
+their orbits arranged in a way to increase the temporal resolution of acquisitions of data with similar geometric and 
+radiometric characteristics. This field allows users to search for related data sets without needing to specify from 
+which specific platform the data came. It is recommended that the value be lowercase with no spaces. Most existing 
+SAR systems are single-satellite, so the constellation name is the same as the satellite name, e.g., 
+`sentinel-1a` (Sentinel-1) and `envisat` (Envisat).
+
+**sar:instrument** is the name of the sensor used, although for Items which contain data from
+multiple sensors this could also name multiple sensors.  It is recommended that the value be lowercase with no spaces. 
+For example, data from the Landsat-8
```
m-mohr: Could we add a SAR example here? Landsat-8 is EO... 

Also, the example is separated from the specification by `It is recommended that the value be lowercase with no spaces.` - I think this sentence should be moved after the example.

extensions/sar/README.md
```diff
@@ -41,13 +41,28 @@ The exact metadata that would appear in a STAC Collection record will vary depen
...
+**sar:platform** is the unique name of the specific platform the instrument is attached to. For satellites this would 
+be the name of the satellite, whereas for drones this would be a unique name for the drone.  It is recommended that 
+the value be lowercase with no spaces. Examples include `sentinel-1a` (Sentinel-1) and `envisat` (Envisat).
+ 
+**sar:constellation** is the name of a logical collection one or more platforms that have similar payloads and have 
+their orbits arranged in a way to increase the temporal resolution of acquisitions of data with similar geometric and 
+radiometric characteristics. This field allows users to search for related data sets without needing to specify from 
+which specific platform the data came. It is recommended that the value be lowercase with no spaces. Most existing 
+SAR systems are single-satellite, so the constellation name is the same as the satellite name, e.g., 
+`sentinel-1a` (Sentinel-1) and `envisat` (Envisat).
```
m-mohr: Wait, Sentinel 1 constists of 1A, 1B (at the moment), 1C and 1D (in the future)?!

philvarner:
> Wait, Sentinel 1 constists of 1A, 1B (at the moment), 1C and 1D (in the future)?!

Yeah, I realized that yesterday afternoon after i'd pushed the changes.  Just fixed.

philvarner:
@m-mohr I believe all of your concerns have been addressed now.

item-spec/examples/digitalglobe-sample.json
```diff
@@ -68,16 +68,16 @@
...
     "eo:gsd": 2.047,
     "eo:epsg": null,
     "dg:catalog_id": "103001004B316600",
     "dg:soli": "056823192",
     "dg:part": 2,
-    "dg:platform": "WORLDVIEW02",
+    "dg:platform": "worldview02",
     "dg:product_level": "LV1B",
-    "dg:product": "WORLDVIEW02_LV1B",
+    "dg:product": "worldview02_LV1B",
```
m-mohr: LV1B is upper-cased.

m-mohr:
@philvarner What benefits do you expect from recommending all lower-case and no spaces? I'm not convinced we need to recommend that. It makes the spec longer, but I don't see a big benefit. APIs should still work case-insensitive whenever possible as it is how user expect it IMHO and otherwise would make searching very hard. It would only be really helpful if it is actually required and can be checked with the validation tools.

item-spec/examples/digitalglobe-sample.json
philvarner: fixed.

philvarner:
The recommendation is to reduce the cognitive load by attempting to have consistency among names.  Otherwise, we'll get every variation, and that will make search more difficult.  E.g., "landsat8" vs. "landsat_8", "landsat-8", "Landsat 8", "Landsat-8", "LANDSAT8", etc., or "landsat8" vs. "Sentinel-2 L2A"

m-mohr:
> The recommendation is to reduce the cognitive load by attempting to have consistency among names. Otherwise, we'll get every variation, and that will make search more difficult. E.g., "landsat8" vs. "landsat_8", "landsat-8", "Landsat 8", "Landsat-8", "LANDSAT8", etc., or "landsat8" vs. "Sentinel-2 L2A"

I agree that having an underscore or a space or a dash between name and number is a problem, but that's not solved by case sensitivity. There's still nothing in the spec regarding dash or underscore. Maybe we should even discourage dashes and underscores. Don't get me wrong, I'm fine with most of the changes in this PR, but I don't think case sensitivity is really a problem here, but the "special chars". So I am not sure how much value the recommendation regarding case sensitivity has and would rather not add it. That's just my opinion though.

philvarner:
My view is that people are going to pick something arbitrary anyway, so we should have a lot of recommendations (but few required).  I agree adding something about non-alphanumeric characters should be added.

hgs-truthe01:
Not sure if it is a complete fix, but I think this might be helped by the summary PR #416 
That way it would be up to the catalog to list out valid strings.

philvarner:
@m-mohr this is ready to merge -- your change request is blocking it, so please clear that. Thanks!

m-mohr:
> @m-mohr this is ready to merge -- your change request is blocking it, so please clear that. Thanks!

So what was the conclusion on the issues I brought up, but couldn't discuss until the end?

1. Why do we specify recommendations regarding case sensitivity for just those fields? Shouldn't we better add those to the best practices document? I feel we make the field descriptions to long with all those recommentations.
2. What about special characters, especially the mentioned `-` and `_`? You said that it leads different names such as `Sentinel-2` and `Sentinel_2`. So Why not just recommend one of the characters instead of two?

Honestly, I'm not really happy about the recommendations yet. Either make strict recommendations or don't make any recommendations. Just some parts don't help very much, I think.

cholmes:
Discussed on call 8/5. We want to move the special character / case recommendations to a 'best practices' document (evolve catalog one to be STAC), and put this info here, along with the recommendations in #464

philvarner:
Removed naming recommendations.

PR #518, opened 2019-06-26
fix mismerge with query 'in' operator and 'time' operator

philvarner:


api-spec/extensions/query/README.md
```diff
@@ -17,7 +17,7 @@ The syntax for the `query` filter is:
 }
 ```
 
-Each property to search is an entry in the `query` filter. <operator> can be one of: `eq`, `neq`, `lt`, `lte`, `gt`, `gte`, `startsWith`, `endsWith`, `contains`. 
+Each property to search is an entry in the `query` filter. <operator> can be one of: `eq`, `neq`, `lt`, `lte`, `gt`, `gte`, `startsWith`, `endsWith`, `contains`, `in`.
```
philvarner: i forgot to add `in` to this list when I originally added it in the yaml and example below.

api-spec/extensions/query/query.fragment.yaml
```diff
@@ -79,5 +79,3 @@ components:
               items:
                 type: string
               description: Find items with a property that matches one of the specified strings. A case-insensitive comparison must be performed.
```
philvarner: `time` got inadvertently re-added when the branch with the `in` operator was merged with the commit that removed `time`

PR #520, opened 2019-06-26
update eo:bands example to clarify index array over name

philvarner:


extensions/eo/README.md
```diff
@@ -139,14 +139,14 @@ See [landsat8-merged.json](examples/landsat8-merged.json) for a full example.
   },
   "eo:bands": [
     {
-      "name": "1",
+      "name": "B01",
```
philvarner: rename this so that there's no confusing that the bands array in the asset is referring to the eo:bands array by index rather than by name value

PR #521, opened 2019-06-26
api: update time parameter to datetime per WFS 3 change

philvarner:


api-spec/STAC-extensions.yaml
```diff
@@ -761,35 +769,44 @@ components:
...
         intersects:
           $ref: 'https://geojson.org/schema/Geometry.json'
-    time:
+    datetimeFilter:
+      description: An object representing a datetime-based filter.
+      type: object
+      properties:
+        datetime:
+          $ref: '#/components/schemas/datetimeOrDatetimeInterval'
+    datetimeOrDatetimeInterval:
```
philvarner: we had two uses of this type, but one should only have allowed an RFC 3339 datetime, while the other allowed that or the `datetime/datetime` interval syntax, so split them into two definitions.

api-spec/STAC-extensions.yaml
```diff
@@ -761,35 +769,44 @@ components:
       properties:
         bbox:
           $ref: '#/components/schemas/bbox'
-    timeFilter:
```
philvarner: moved this down so bbox and intersects would be next to one another, and the 3 datetime-related ones would be also.

api-spec/STAC-extensions.yaml
m-mohr: This is how WFS defines it. https://raw.githubusercontent.com/opengeospatial/WFS_FES/master/core/openapi/parameters/datetime.yaml

Should we follow them?

matthewhanson:
@philvarner Does WFS3 allow for specifying incomplete datetimes?  That is, can I specify just the date?  This is of course the main use case, by a wide margin, and I'd like to not have to specify a time.

Being able to specify a single datetime (and not an interval) isn't that useful unless I can specify only the date.  This means that it really is an interval of a single day, but I think the shortcut is valuable.

m-mohr:
> @philvarner Does WFS3 allow for specifying incomplete datetimes?

No.


> This is of course the main use case, by a wide margin, and I'd like to not have to specify a time.

Clients could extend `2019-01-01` to `2019-01-01T00:00:00Z/2019-01-02T00:00:00Z` I guess?!

api-spec/STAC-extensions.yaml
philvarner: > Should we follow them?

I'll update this wrt their definition

m-mohr:
After aligning with WFS3 draft2 in #525, I would propose the following changes:
* Follow WFS3 strictly regarding datetime
* Add another query parameter `date`, which allows exactly the same, but for dates only. This would cover what @matthewhanson critizied in #521. Examples
    * `?date=2017-01-01` to cover the whole 1st of Jan 2017
    * `?date=2017-01-01/2017-03-31` to cover Jan/Feb/Mar

philvarner:
>After aligning with WFS3 draft2 in #525, I would propose the following changes:
> Follow WFS3 strictly regarding datetime (I could include this in #525)

Yes.

> Add another query parameter date in this PR, which allows exactly the same, but for dates only. This > would cover what @matthewhanson critizied in #521. Examples
> ?date=2017-01-01 to cover the whole 1st of Jan 2017
> ?date=2017-01-01/2017-03-31 to cover Jan/Feb/Mar

I don't think this is necessary, as one can do `datetime=2017-01-01T00:00:00Z/2017-01-31T23:59:59.999Z`  (thought a separate clarification is whether the semantics of WFS 3 is that the time is >= to the start and <= end, as it's not explicitly clear in that spec.  I've usually implemented the semantics that it's >= start and < end to facilitate month query)

m-mohr:
> I don't think this is necessary, as one can do `datetime=2017-01-01T00:00:00Z/2017-01-31T23:59:59.999Z` (thought a separate clarification is whether the semantics of WFS 3 is that the time is >= to the start and <= end, as it's not explicitly clear in that spec. I've usually implemented the semantics that it's >= start and < end to facilitate month query)

I've asked Clemens (WFS editor) some days ago and he said they are following ISO8601 and it is >= and <=. I don't like that and prefer >= and < as you do, but I think that has been discussed already in the WFS group and we can't change it. Therefore, we probably just follow their decision.

m-mohr:
I ported the changes proposed here to #529 . Discussions regarding a date query parameter can be continued in #536.

Issue #517, opened 2019-06-26
Align definition of API "time" parameter with WFS 3 "datetime" parameter


philvarner:
The STAC definition for the `time` parameter is currently "Single date, date+time, or a range ('/' separator), formatted to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6)"

There are also several syntax and semantic differences:
1. WFS 3 now calls this "datetime" http://docs.opengeospatial.org/DRAFTS/17-069r1.html#_parameter_datetime .  Our examples of WFS 3 in examples.md are wrong wrt to this.
1. Duration syntax is  not supported. 
2. ranges allow open and closed datetimes and intervals.
3. RFC 3339 defines complete datetime w/ timezone, so a date only (e.g, 2019-01-01) is not a valid string. Our only /stac/search example in examples.md does this. 
4. Our examples with an interval like P1Y2M10DT2H30M are not supported.  In WFS 3, an interval is explicitly two strings separated by `/`, of which the strings must be an RFC3339 datetime or `..`

I think the work is simply to align our name and definition of this parameter with what WFS 3 has.

m-mohr:
So we inherit this from WFS or is this defined in STAC openAPI files? If this is inherited from WFS, this will probably solve with #472

m-mohr:
See #521 for discussions.

m-mohr:
Done with #525.

Issue #512, opened 2019-06-24
Change references from "WFS 3" to "OGC API - Features - Part 1: Core"


philvarner:
Not sure what the acronym for it is, but we should figure it out (OGCAPIFP1C?, lol).

Also mention that OGC API - Features - Part 1: Core was previously WFS 3.

I'm glad to make the changes, but I wanted some indication as to if we should do it or not first.

m-mohr:
OGC still figures out the acronym. Until they made the first release under the new name, I'd leave it as it is. People know it still as WFS3 and OGC ist still working on the new names etc.

m-mohr:
Will it be OAFeat 1.0?! 🤣 

See https://github.com/opengeospatial/oapi_common/issues/35#issuecomment-509598050

Issue #357, opened 2018-11-18
STAC Extensions Prefixes suggestion


mundialis-dev:
In the extension documentation of STAC API the use of prefixes is suggested to include vendor specific properties like:

```
  "properties": {
    "datetime":"2018-01-01T13:21:30Z",

    "dtr:start_datetime":"2018-01-01T13:21:30Z",
    "dtr:end_datetime":"2018-01-01T13:31:30Z",

    "eo:off_nadir_angle": -0.001,
    "eo:cloud_cover": 10.31,
    "eo:sun_azimuth": 149.01607154,
    "eo:sun_elevation": 59.21424700,
    "eo:resolution": 30,
  }
```

I would suggest to avoid prefixes, but to use sub-categories/dict/maps/groups to specify properties of extension:

```
"properties": {
  "datetime":"2018-01-01T13:21:30Z",
  "dtr" : {
    "start_datetime":"2018-01-01T13:21:30Z",
    "end_datetime":"2018-01-01T13:31:30Z"
   },
  "eo": {
    "off_nadir_angle": -0.001,
    "cloud_cover": 10.31,
    "sun_azimuth": 149.01607154,
    "sun_elevation": 59.21424700,
    "resolution": 30,
  }
}
```

This will make the implementation in languages that support "objects to JSON" serialization like Python much easier, since each property can be mapped directly from a class member variable.

In addition one can identify all extensions easier, without the need to parse all parameters to check the prefixes.

jdries:
I also ran into this issue while using an entirely different environment: we use apache NiFi to process metadata records. While trying to use STAC metadata with properties like eo:cloud_cover, I got an error from the avro library which also does not support colons in its property names.
Avro is also a data serialization system, which allows json records to be represented in a binary form, for more efficient processing. 
I have not yet been able to implement a workaround for this, mostly because this is not really a 'programming' environment where I can easily sanitize and desanitize property names without introducing a lot of overhead for our operators.

m-mohr:
Thanks for pointing that out.

1. I see the problem that `:` can be a problem in certain environments. Nevertheless, this is an implementation issue as `:` in JSON keys are valid according to the JSON specification and some tools decided to not implement it fully.
2. The reason to keep the properties flat is that some GeoJSON readers don't support objects/arrays as properties, so in the end everything except "datetime" wouldn't be readable by some tools. This is also an implementation issue as arrays and objects in properties, are valid according to the GeoJSON specification.

[...]

PR #525, opened 2019-07-17
Backporting WFS3 draft 2 to STAC

m-mohr:
Tackles #472, partly #517 and #512.

Therefore, updates WFS3 OpenAPI file, updates API documentation and backports (a subset of) changes to STAC collections. This restores compatibility with the latest draft of WFS3 (aka OGC API - Features - Part 1: Core) as released here: https://github.com/opengeospatial/WFS_FES/releases/tag/1.0.0-draft.2

Notes:
* Several API query parameters changed significantly (e.g. `datetime`). ~~We are currently diverging here (`time` vs `datetime`).~~ I tried to fix the API examples, but please check carefully. ~~Open question: should we align? See #517~~ (Edit: I removed `time` so we are now aligned with WFS. See comments below.)
* ~~In addition to the WFS3 backporting I changed the following sentence in the API spec from `Either a date-time or a period string that adheres to RFC3339.` to `Either a date-time that adheres to RFC3339 or a period string that adheres to ISO8601.` RFC3339 doesn't specify period strings.~~ (Edit: Not relevant any longer after removing the `time` parameter.)
* Merging STAC and WFS OpenAPI files lead to a mal-formed OpenAPI file, therefore un-commented some required fields in the WFS YAML . They are also required in STAC, so that's not a problem.
* WFS links have additional properties `hreflang` and `length`, which are optionally available in the API, but were not ported back to STAC Items/Catalogs/Collections (see https://github.com/radiantearth/stac-spec/issues/472#issuecomment-511880235)
*  Also, I haven't backported the optional fields `crs` and `itemType` to STAC Collections (see https://github.com/radiantearth/stac-spec/issues/472#issuecomment-511880235)
* Extents have changed in STAC Collections. I did not backport `trs` and `crs` in the extents as WFS3 only allows a single value and defaults to it. Therefore, we can simply ignore it and force all values to comply to these defaults (WGS84 / Gregorian Calender).
* I found some issues in the WFS draft which make the OpenAPI file unsuable and reported these issues to Clemens Portele. Those issues are already solved here, but will only be part of the next WFS release. See https://github.com/opengeospatial/WFS_FES/commit/1cd3549a84066885faf837a99cb17fa682606a43 and https://github.com/opengeospatial/WFS_FES/commit/d1fbe39b61054606be59bc94e2965d93dbb03a0a
* This PR also contains some fixes for invalid Collection examples
* For pagination they have `numberMatched` and `numberReturned` as well as links to support paging (link relation next). We diverge from this with our extension. Should we align?

**PR Checklist:**


philvarner:
> Several API query parameters changed significantly (e.g. datetime). We are currently diverging here (time vs datetime). I tried to fix the API examples, but please check carefully. Open question: should we align? See #517

I would like us to get rid of `time` and only have the `datetime` parameter, and with the same definition as in WFS.  While it might be nice to allow different formats and expressions, I don't think the added capability outweighs the additional required implementation burden.

philvarner:
In addition to the WFS3 backporting I changed the following sentence in the API spec from Either a date-time or a period string that adheres to RFC3339. to Either a date-time that adheres to RFC3339 or a period string that adheres to ISO8601. RFC3339 doesn't specify period strings.

I don't know what a period string is.  I would prefer we only support RFC3339 as WFS 3 does, unless there's a compelling reason to support the ISO8601 ones also.

philvarner:
> For pagination they have numberMatched and numberReturned as well as links to support paging (link relation next). We diverge from this with our extension. Should we align?

Links to support paging become problematic when you could possibly submit a multi-megabyte GeoJSON string as a parameter, hence our `next` parameter and `search:metadata.next` field.

philvarner:
> For pagination they have numberMatched and numberReturned as well as links to support paging (link relation next). We diverge from this with our extension. Should we align?

I think we should make them `numberMatched` and `numberReturned` to align.

m-mohr:
@philvarner Thanks for the review.

> I would like us to get rid of `time` and only have the `datetime` parameter, and with the same definition as in WFS.

Fine for me, was proposing this to address concerns raised by @matthewhanson. What do you think, Matt?


> I don't know what a period string is. I would prefer we only support RFC3339 as WFS 3 does, unless there's a compelling reason to support the ISO8601 ones also.

WFS defines the period strings itself (with another name). RFC3339 only defines a single instance in time and therefore, WFS took the notion for intervals (here: period strings) from ISO8601, including the two dots for open intervals. So I guess we should align to the WFS terms here, but we can't really get rid of ISO8601. 

> Links to support paging become problematic when you could possibly submit a multi-megabyte GeoJSON string as a parameter, hence our `next` parameter and `search:metadata.next` field.

I don't get that. Isn't that more or less the same? Can't you simply generate the link with our next token and add it to the links?

m-mohr:
I removed the `time` parameter for now (related discussions also in #521 and #536). If we decide to add `date` again, we can do that in another PR, I think. I have opened issue #536 to streamline the discussions whether we want a date parameter or not.

philvarner:
Are the period strings the interval syntax, `2019-07-22T00:00:00/..`? If not, can you show me an example?

m-mohr:
> Are the period strings the interval syntax, `2019-07-22T00:00:00/..`? If not, can you show me an example?

Yes, period string = interval syntax. "period string" is not in the API specification any longer, we use the WFS description now.

philvarner:
>> Links to support paging become problematic when you could possibly submit a multi-megabyte ? 
>> GeoJSON string as a parameter, hence our next parameter and search:metadata.next field.

> I don't get that. Isn't that more or less the same? Can't you simply generate the link with our next 
> token and add it to the links? @philvarner

You could do that, but then it's misleading because it's only part of the url for the next page.  For example, the url `https://example.com/stac/search?geometry={1mb geojson}` would mean that you either set the `next` link value to `https://example.com/stac/search?geometry={1mb geojson}&next=some_token` or `https://example.com/stac/search?next=some_token`, the first of which is intolerably large and the second of which is only part of the URL required to the the next page of results, but with no indication that that's the case. WFS 3 doesn't have this problem because none of the parameters to their GET endpoint are likely to have large values, like our `geometry` parameter can.

That's why we settled on semantics whereby the response has a `search:metadata.next` value, and the client is responsible for constructing the next query for the next page of results with that value set as the value of the `next` parameter.

m-mohr:
@philvarner Thanks for the clarification. Then I don't see a good reason to align paging and/or search metadata, I think.

philvarner:
> For pagination they have numberMatched and numberReturned as well as links to support paging (link relation next). We diverge from this with our extension. Should we align?

An underlying issue with this is that if we decide to align (which I believe we should), we're going to need to introduce fields with camelCase.  I feel that if getting parts of STAC adopted WFS/OGC is an objective of this project, then we should do as much as possible to align our naming scheme with those specs, even if some people may not like camelCase.

m-mohr:
We are in dependency hell here. openEO API uses STAC, too, and that is also using snake_case to be aligned. We could do it together with #357, if we do it. But just for aligning this little bit, I wouldn't do all the work just for aligning casing. Anyway, this is a completely different topic. ;-)

matthewhanson:
This looks good - totally on board with `datetime` as I always hated the `time` parameter and happy that WFS3 changed it.

m-mohr:
STAC Working Session Notes:

> * Several API query parameters changed significantly (e.g. `datetime`). ~We are currently diverging here (`time` vs `datetime`).~ I tried to fix the API examples, but please check carefully. ~Open question: should we align? See #517~ (Edit: I removed `time` so we are now aligned with WFS. See comments below.)

Align with datetime. We may add a date query parameter in an extension if required.


> * For pagination they have `numberMatched` and `numberReturned` as well as links to support paging (link relation next). We diverge from this with our extension. Should we align?

Don't align at the moment.

PR #527, opened 2019-07-18
Summaries for Catalogs (and Collections)

m-mohr:
PR for #413 as discussed on the last sprint.

It's a first draft for discussion.

Side note: In a separate PR, I'd like to get full JSON Schema support for Item properties in `properties` and `summaries`, but this needs #528 to be solved. This requires rewriting all JSON schemas.

**PR Checklist:**


catalog-spec/examples/summaries-s2.json
```diff
@@ -0,0 +1,180 @@
...
+    ]
+  },
+  "links": [
+    {
+      "href": "item.json",
+      "rel": "item"
+    },
+    {
+      "href": "summaries-s2.json",
+      "rel": "self"
```
cholmes: I think we should only have examples with self links that are absolute, to be in line with our best practices (or requirement? I forget where we landed).

collection-spec/collection-spec.md
```diff
@@ -32,11 +32,14 @@ Implementations are encouraged, however, as good effort will be made to not chan
 | license      | string            | **REQUIRED.** Collection's license(s) as a SPDX [License identifier](https://spdx.org/licenses/) or [expression](https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60) or `proprietary` if the license is not on the SPDX license list. Proprietary licensed data SHOULD add a link to the license text, see the `license` relation type. |
 | providers    | [Provider Object] | A list of providers, which may include all organizations capturing or processing the data or the hosting provider. Providers should be listed in chronological order with the most recent provider being the last element of the list. |
 | extent       | Extent Object     | **REQUIRED.** Spatial and temporal extents.                  |
-| properties   | object            | Common fields across referenced items. May also be used to describe standalone collections better that don't reference any items. See the section 'Common Fields' for more information. |
+| properties   | object            | Common fields across referenced items. |
+| summaries    | Map<string, [*]|Range Object> | A map of property summaries, either a set of values or a range. |
```
cholmes: This should be <Range Object>

catalog-spec/catalog-spec.md
```diff
@@ -119,11 +119,56 @@ details on the two types and how you might use them see the [Static and Dynamic
 | id           | string        | **REQUIRED.** Identifier for the catalog.                    |
 | title        | string        | A short descriptive one-line title for the catalog.          |
 | description  | string        | **REQUIRED.** Detailed multi-line description to fully explain the catalog. [CommonMark 0.28](http://commonmark.org/) syntax MAY be used for rich text representation. |
+| summaries    | Map<string, [*]|Range Object> | A map of property summaries, either a set of values or a range. |
```
cholmes: Typo here, missing the open bracket around 'Range Object'

catalog-spec/examples/summaries-s2.json
m-mohr: I removed the self link as I don't have a reasonable link to put there. Or should I invent one to have an absolute self link?

collection-spec/collection-spec.md
m-mohr: From your comment not sure what exactly you want to have changed, but I assume escaping the pipe? Done!

catalog-spec/catalog-spec.md
m-mohr: Fixed.

m-mohr:
> The one other thing I'd like to see is the 'range object' moved down in the spec a bit

@cholmes Moved it to the end.

> I think in general it makes sense to put the objects in order of 'importance' to a reader.

What is important? Probably required fields first and then? For me Range is more important than Object, but others will have other priorities...

> We also discussed on the co-working call putting links from the property list into the actual object sections, so that order matters less, as people can link right in. But that can be its own issue/PR.

I made an issue for it #541.

> I see that the whole section is optional, but I think it feels a little unclear if when it exists whether it contains a complete list of properties.

@hgs-truthe01 I added the following text to clarify it. Does it help you and does it make sense, too?
"It is recommended to list as many properties as reasonable so that consumers get a full overview about the properties included in the Items. Nevertheless, it is not very useful to list all potential `title` values of the Items. Also, a range for the `datetime` property may be better suited to be included in the STAC Collection. In general, properties that are covered by the Collection specification (e.g. `providers` and `license`) may not be repeated in the summaries."

m-mohr:
Waiting for @matthewhanson to review this PR, don't merge yet!

Issue #439, opened 2019-04-25
Allow mixing different STAC versions


vmx:
The current STAC Catalog Spec [forbids having sub-catalogs with a different version](https://github.com/radiantearth/stac-spec/blob/4053b834204ca339701f9866ccb37b0609005995/catalog-spec/catalog-spec.md#catalog-fields):

> Child Catalogs and child Collections MUST NOT specify a different STAC version

This was introduced and discussed at https://github.com/radiantearth/stac-spec/pull/268.

It is also possible to link to someone else's catalog:

> Thus a published root catalog might be a sub-catalog of someone else's structure.

There you don't have control the version of the catalog. Hence a link to the other catalog might make my own catalog not conforming to the spec if its version changes.

Hence I propose changing the sentence from "MUST NOT" to "SHOULD NOT" with a note that catalogs under your control should have the same version, external ones would be the exception (and the reason for not being a "MUST NOT". I expected that we use the definition of these terms as defined by [RFC 2119](https://www.ietf.org/rfc/rfc2119).

PS: I'm totally new to STAC, so it could well be that I'm missing something :)

matthewhanson:
@vmx 
I think this is a reasonable requirement to relax. I can also see a provider that might eventually have a lot of different collections of historical data (e.g., a drone operator), where you create the collection and items for that mission, but then have a catalog containing all of your collections. It may not be worth it to have to update all those older collections if that STAC version does what you need it to. I think this is OK for static catalogs.

On the other hand, a STAC API is a different story. Because STAC versions might change the API, you really need to have all the collections in the catalog be the same version. *Perhaps* changes in patch version would be ok (e.g., 0.6.x), but this should also be discouraged.

m-mohr:
I agree with @matthewhanson. For APIs it is probably good to stich with a single version and I'd guess quite easy to implement. For static catalogs this is a different story.

This restriction is mainly there to avoid clients running into problems. If you have a 0.4 (static) catalog mixed with 0.6 you'll very likely run into problems when reading it with STAC Browser and other clients. This could only be solved by implementing all versions in client software. I think this was the fear that made us adding the restriction. We'll find out in the long run whether that's something that's practically working or not and therefore get's removed or not. So I'd say leave this issue open and revisit in some time to check existing implementations etc.

cholmes:
Agreed with the other comments. Also I don't think we yet have examples of using a sub-catalog of someone else's structure, so I'm less clear if we should call that out in the spec.

[...]

Issue #278, opened 2018-10-10
Extensions supported


matthewhanson:
It would be useful to know the extensions that are supported by a given collection. In one catalog you may have a collection of EO data, and another collection of pointcloud data. 

I suggest an optional extensions field at the collection level.

```
extensions = ['eo', 'commons']
```

m-mohr:
Could you elaborate more when this is useful and where to put this? Would this be added to the collection link in the item? Or would this be added to the catalog/collection json file itself? If the latter, you could just figure it out by checking for required fields, right? Other than that: What are the names that you are using? eo and commons are not directly specified somewhere. One could get the idea that these values are the prefixes, but commons doesn't has a prefix. How would you reference the datetime range extension (see #274)? dtr (prefix)? datetime-range (folder name), ...?

matthewhanson:
@m-mohr Maybe something in the links would be a better way too, a rel="extension" with a link to the extension definition.

The names I used above would be an extension ID, but those aren't currently defined.

You could check the fields in the Items, but it's a little awkward. You would have to iterate through all the attributes, split them on ':', then get the set of them to know which extensions are used.

One use case is this: I've got a static catalog of collections, and want to ingest them into some other system but I only support certain extensions. 

I was thinking another reason would be if I wanted to ingest all EO data or all pointcloud data in a catalog, but even this wouldn't help. An individual record has no indication what 'type' it is, and an 'extension' isn't synonymous with 'type' anyway.

We can wait on this for now and discuss more later, I'll change the label.

m-mohr:
Thanks for the response.
Your proposal sounds a little related to #264 as it both involves linking schemas.
It seems also a little related to #214 as you mentioned the type - but that only gives an indirect hint as you also noticed.
Yes, I also think we need more time to think about this. Personally, I am not sure yet what a good solution would look like.

cholmes:
Yeah, this could be a good one to talk through - understand the use cases, etc.

I do think there are two types of extensions that we might treat differently. EO is a content extension - just different fields, and can be validated with a JSON Schema. Commons is more of a functionality or behavioral extension. 

For content extensions I would see an item articulating what content extensions it implements. And to me that would ideally link to a JSON Schema. Like it should refer to the extension name / id, as well as the schema that defines the extension. And it would link to all the schemas that define it. 

[...]

PR #531, opened 2019-07-22
Extension supported (#278) / Mixing STAC versions (#439)

m-mohr:
Adds the "extensions supported" field as discussed on the STAC sprint 4, see #278.
Allows to mix STAC versions in catalogs, see #439.

Partially tackles #533 (inherit JSON Schema fields from Catalog) to avoid that extensions need to be added to both the Collection and the Catalog JSON schemas.

**PR Checklist:**


catalog-spec/json-schema/catalog.json
```diff
@@ -42,6 +49,29 @@
...
+              "title": "Reference to a core extension",
+              "type": "string",
+              "enum": [
+                "checksum",
+                "datacube",
+                "datetime-range",
+                "eo",
+                "pointcloud",
+                "sar",
+                "scientific"
```
hgs-truthe01: Should this list include API extensions? Or should we add them separately at /stac explicitly?

matthewhanson: I think that's separate issue regarding the API response, and since we had discussed recently about versioning the API separately, I think we nee to also talk about returning the API version as well.

matthewhanson:
We talked about this a bit during the working session and there are still some questions/issues that could use more discussion:

- do `stac_extensions` need to be at the `item` level?
- if at the `item` level do they go in `properties` so they can be put in the collection by the data provider if desired?
- are `stac_extensions` really needed at the `catalog` level?
- if it can appear the at the `catalog` and `collection` level, is there some sort of inheritance that should occur?
- related, does `stac_version` belong at the `item` level?  We think this should mostly depend on what's best for validation
-

jbants:
> - do stac_extensions need to be at the item level?

I think so.
>  - if at the item level do they go in properties so they can be put in the collection by the data provider if desired?

Yea, I have in my notes that we settled on this.
>  - are stac_extensions really needed at the catalog level?

We discussed having them at the item and collections levels. Not the catalog level
>  - if it can appear the at the catalog and collection level, is there some sort of inheritance that should occur?

Item `stac_extension` trumps collection `stac_extension`
 > - related, does stac_version belong at the item level? We think this should mostly depend on what's best for validation
I think so. The example i had written down was
```
{
"custom_extension_1": "https://....",
"eo": "0.0.1",
"cube": "0.0.1"
}
```
If there is no extension version, use the stac version

catalog-spec/catalog-spec.md
```diff
@@ -116,14 +116,49 @@ details on the two types and how you might use them see the [Static and Dynamic
 | Element      | Type          | Description                                                  |
 | ------------ | ------------- | ------------------------------------------------------------ |
 | stac_version | string        | **REQUIRED.** The STAC version the catalog implements.       |
+| stac_extensions | [string]   | A list of extensions the catalog and its items have implemented. |
```
jbants: Wasn't this going on the collection and item levels?

catalog-spec/catalog-spec.md
```diff
@@ -150,6 +186,7 @@ A typical '_child_' sub-catalog could look similar:
 ```json
 {
   "stac_version": "0.7.0",
+  "stac_extensions": ["eo"],
```
jbants: I had in my notes, that we were going to add versions to the extensions.
Something like this?
```
{
"custom_extension_1": "https://....",
"eo": "0.0.1",
"cube": "0.0.1"
}
```

collection-spec/examples/landsat-collection.json
```diff
@@ -1,10 +1,14 @@
 {
+    "stac_version": "0.7.0",
+    "stac_extensions": [
```
jbants:
```
"stac_extensions": {
        "custom_extension_1": "https://....",
        "eo": "0.0.1",
        "cube": "0.0.1"
    }
```

collection-spec/examples/sentinel2.json
```diff
@@ -1,5 +1,8 @@
 {
   "stac_version": "0.7.0",
+  "stac_extensions": [
```
jbants:
```
"stac_extensions": {
        "eo": "0.0.1",
    }
```

extensions/datacube/examples/example.json
```diff
@@ -1,5 +1,8 @@
 {
   "stac_version": "0.7.0",
+  "stac_extensions": [
```
jbants:
```
"stac_extensions": {
        "cube": "0.0.1"
    }
```

extensions/label/examples/spacenet-roads/roads_collection.json
```diff
@@ -1,5 +1,8 @@
 {
   "stac_version": "0.7.0",
+  "stac_extensions": [
```
jbants:
```
"stac_extensions": {
        "label": "0.0.1"
    }
```

extensions/scientific/examples/example-merraclim.json
```diff
@@ -1,5 +1,9 @@
 {
   "stac_version": "0.7.0",
+  "stac_extensions": [
```
jbants:
```
"stac_extensions": {
        "eo": "0.0.1",
        "scientific": "0.0.1"
    }
```

extensions/README.md
```diff
@@ -136,3 +136,8 @@ These files should be kept together in order to preserve relative `$ref` links.
 See the [EU](eo/) extension file structure as an example.
 * Specification examples should be stored in an `examples` directory. 
 * The specification schema file(s) should be stored in a `json-schema` directory. 
+
+Make sure to choose a meaningful name for the extension folder as it will be the shortcut
```
jbants: If we go the object route for defining extensions, the key can be used as the shortcut for identifying the extension schema.

catalog-spec/json-schema/catalog.json
m-mohr: I'm agreeing with @matthewhanson here. API extensions could go in `/conformance` if we'd like to go the OGC way.

m-mohr:
> > * if at the item level do they go in properties so they can be put in the collection by the data provider if desired?
> 
> Yea, I have in my notes that we settled on this.

At some point we had this idea, but I think in the end we settled to put them (top-level) together with stac_version. 

> > * are stac_extensions really needed at the catalog level?
> 
> We discussed having them at the item and collections levels. Not the catalog level

They should be on the same level as stac_version, I think. They belong together I think.

> > * related, does stac_version belong at the item level? We think this should mostly depend on what's best for validation
> >   I think so. The example i had written down was
> 
> ```
> {
> "custom_extension_1": "https://....",
> "eo": "0.0.1",
> "cube": "0.0.1"
> }
> ```

That's an outdated example. We discussed that at some point, but later agreed that this is not necessary. Instead the version is part of the URL. If not URL is specified, it always aligns with the stac_version. Makes the structure much easier and doesn't need objects, just arrays. See the examples attached to the PR.

extensions/label/examples/spacenet-roads/roads_collection.json
m-mohr: See my comment below. I'm pretty sure we settled on arrays as the version information is basically included in the URL as the referenced schema is versioned.

extensions/datacube/examples/example.json
m-mohr: See my comment below. I'm pretty sure we settled on arrays as the version information is basically included in the URL as the referenced schema is versioned.

collection-spec/examples/sentinel2.json
m-mohr: See my comment below. I'm pretty sure we settled on arrays as the version information is basically included in the URL as the referenced schema is versioned.

collection-spec/examples/landsat-collection.json
m-mohr: See my comment below. I'm pretty sure we settled on arrays as the version information is basically included in the URL as the referenced schema is versioned.

catalog-spec/catalog-spec.md
m-mohr: See my comment below. I'm pretty sure we settled on arrays as the version information is basically included in the URL as the referenced schema is versioned.

extensions/scientific/examples/example-merraclim.json
m-mohr: See my comment below. I'm pretty sure we settled on arrays as the version information is basically included in the URL as the referenced schema is versioned.

catalog-spec/catalog-spec.md
m-mohr: Not sure about this. We should check use cases and see where we need it.

m-mohr:
Discussed on today's call. We agreed on having stac_version and stac_extensions in Items, Catalogs and Collections. They define the content extensions in their context, i.e. stac_extensions in a Collections defines the extensions that are applied to the collection, NOT the extensions that are defined in childs/items. So there's no inheritance.

m-mohr:
Updated according to the recent discussions. I hope I found all Items where I had to add the stac_version.

matthewhanson:
Ah, there's conflicts, must be due to merging #527 , @m-mohr probably best if you resolve these.

PR #532, opened 2019-07-22
Fixed label extension examples and re-generated OpenAPI files

m-mohr:
Two fixes:
1. Fixed issues in the label extension examples 
    - Invalid version numbers
    - Commas after last element in array
2. Re-generated the STAC API files as they seem to be outdated (seems a PR missed to run `npm run generate-all`)

PR #534, opened 2019-07-22
ETag support for optimistic locking + add transaction extension to STAC-extensions.yaml

hgs-truthe01:
While using the transactions extension we discovered some issues with concurrent changes to the documents. I wanted to propose using the ETag header to support optimistic locking to prevent changes from stepping on each other.

In this proposal, the PATCH call doesn't require a ETAG, but optionally supports one. Since it isn't replacing the whole document and is making specific changes I think that makes sense. PUT should require it though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Moved Single Item Extension to core (`license` and `providers` properties for Items)
 
+### Added
+
+- API Transaction extension supports optimistic locking through use of the ETag header.
+
 ## [v0.7.0] - 2019-05-06
 
 ### Fixed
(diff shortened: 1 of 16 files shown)


hgs-truthe01:
Also, I want to mention that I ran all of the generate scripts, so there are some changes that are not part of this PR that ended up in the generated files. I'm not sure if it is better to get everything synced up in this PR or restrict it to just my changes.

m-mohr:
> Also, I want to mention that I ran all of the generate scripts, so there are some changes that are not part of this PR that ended up in the generated files.

You could review #532, merge it in dev and here, and then this should fix stuff here.

I'll have a look at this PR tomorrow.

hgs-truthe01:
Thanks for the help on the cleanup @m-mohr !

cholmes:
Just reviewed and merged #538 as I saw this go by on my email.

m-mohr:
Great, will resolve conflicts later and merge.

m-mohr:
**PR checklist:**

- [x] This PR has breaking changes.
- [ ] I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-spec/blob/dev/CHANGELOG.md).
- [x] API only: I have updated the [generated OpenAPI files](https://github.com/radiantearth/stac-spec/blob/dev/api-spec/README.md#openapi-definitions) by executing `npm run generate-all`.

PR #537, opened 2019-07-23
Add transaction extension to STAC-extensions.yaml

m-mohr:
I added the Transaction extension to STAC-extensions.yaml. It's not clear why it is missing.

We should also add the search extension, which currently only has a JSON Schema but not a YAML fragment.


m-mohr:
Duplicate of #534

PR #538, opened 2019-07-23
search + transaction extension added to STAC-extensions.yaml

m-mohr:
Here's a YAML definition for the search extension as it was missing from STAC-extensions.yaml as of now.

In contrast to the JSON Schema, I did not make `search:metadata` required in the YAML. 

Also added the Transaction extension to the STAC-extensions.yaml.

Plus: Some fixes.

### DIFF
--- a/api-spec/STAC-extensions.yaml
+++ b/api-spec/STAC-extensions.yaml
@@ -1,6 +1,6 @@
 openapi: 3.0.1
 info:
-  title: The SpatioTemporal Asset Catalog API
+  title: The SpatioTemporal Asset Catalog API + Extensions
   version: 0.7.0
   description: >-
     This is an OpenAPI definition of the core SpatioTemporal Asset Catalog API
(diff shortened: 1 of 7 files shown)


api-spec/STAC-extensions.yaml
```diff
@@ -855,6 +1008,25 @@ components:
...
+            - returned
+          properties:
+            next:
+              type: string
+              nullable: true
+            limit:
+              type: integer
+              nullable: true
+              minimum: 0
+            matched:
```
scisco: I think in our discussion we decided that providing a count on search might be too expensive for many implementations. Could we make this optional?

m-mohr: @scisco This hasn't been introduced here. The extension and the requirement for count was introduced in #482. With this PR, I just fixed the issue that it wasn't compiled into our full OpenAPI definition. As I don't change any semantics of the spec here, I would prefer to discuss your note in a separate issue or PR.

scisco: Yeah, it makes sense. Thanks for pointing to the right ticket.

Issue #289, opened 2018-10-13
Asset schema / Further refine adding additional metadata to the asset definition


m-mohr:
*Taken from the old catalog recommendations.*

**Asset definition**

It is of top priority to provide definitions of assets, so clients can know get some information as to 
what is actually contained in the data they are downloading. Assets should contain the metadata that is 
specific to the format of the asset. The asset must have some way to state its format. 

While assets in Catalog APIs can easily represent files that are generated on demand, a static
catalog should only list assets that already exist online. This increases the reliability and speed
of the catalog and makes it easier to fully duplicate or back-up a static catalog.

**Note:** 0.6.0 defines Media types for assets, to provide more information about what is to download, but there is likely
more work to be done to fully define the asset, so providers should feel free to experiment and give feedback on potential asset improvments.

There are a couple of ideas on how to define metadata for assets. Providers could either

* include a number of optional metadata fields directly in the asset that would provide additional definition for clients or
* include a field that specifies a URI to a (JSON) file that has the additional metadata.

When adding the data via a URI, all the needed metadata for a client would be in the product level definition,
which would mean a lot less repeating of fields. But clients would have to be a bit smarter to actually follow the links.
Though ideally clients would be able to cache and reuse product definitions, and hopefully there would be a small number
of canonical product definitions online, that all static catalogs using a particular provider would be able to reference.

m-mohr:
Related to #153

m-mohr:
Do we need to move anything from this old best practices to the new best practices, @cholmes ?

cholmes:
Closed with #552

Issue #153, opened 2018-08-14
Defining some asset fields in collections


matthewhanson:
One of the things we hard originally discussed was having an "available_assets" field. This is a useful field to define in collections, as it tells you what you can expect to find in a single member item.

There's nothing specifically that prevents this from being the case in the definition of collections, but it does make the logic of merging an item with it's collection level metadata more complex. If a collection defines some fields of an asset (the ones that are duplicated across all items such as key, file type, available keys, etc.) and the item defines just the actual URL of those items, then a client that merges them (such as [sat-search](https://github.com/sat-utils/sat-search/) ) needs to make sure it merges all of the available keys into the correct items.

For example, right now with assets as objects (may change, see https://github.com/radiantearth/stac-spec/pull/151) in the collection assets could be defined as just a dictionary of keys with empty values:

assets: {
   'B1': {},
   'B2': {},
   'B3': {}
}

Which informs the user who is examining the collection that B1, B2, and B3 are available assets. Otherwise right now, such as in sat-api, you have to look at a specific item to even know what assets are available, and even in that case you are assuming that the one item is representative of all items in the collection, which is a bad assumption. Some items could define additional assets, but the collection should indicate which assets are available in all items.

cc @cholmes @mojodna @scisco

mojodna:
This still seems achievable using the `key` or `role` attribute in place of object keys, indicating which are expected to have values associated.

matthewhanson:
@mojodna Yes, it is achievable, but only if an id is constrained to be unique among all items in the list. But it also makes it very difficult to merge them as described in #151 whereas a recursive merge of dictionaries is straightforward.

cholmes:
Moving this to 0.7.0, as getting this in is a top priority for that release, but won't happen here.

PR #552, opened 2019-08-15
asset definition extension added

matthewhanson:
**Related Issue(s):** #153 #289 

**Proposed Changes:**

1. An Asset Definition object added to the collection to describe possible assets in member Items.

**PR Checklist:**


extensions/README.md
```diff
@@ -47,9 +47,10 @@ An extension can add new fields to STAC entities (content extension), or can add
 
 | Extension Name (Prefix)                                      | Scope            | Description                                                  | Maturity |
 | ------------------------------------------------------------ | ---------------- | ------------------------------------------------------------ | -------- |
+| [Asset Definition](asset/README.md) (`asset`)                  | Collection | Provides a way to specify details about what assets may be found in Items belonging to a collection | *Proposal* |
```
m-mohr: The extension doesn't specify a prefix in the documentation, but here one is specified? Should probably be (-) or so.

extensions/asset/README.md
```diff
@@ -0,0 +1,29 @@
+# Asset Definition Specification (`asset`)
```
m-mohr: Same here, the extension doesn't use a prefix, but specifies one?!

extensions/asset/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+This extension introduces a single new field, `assets` at the top level of a collection. An Asset Object defined at the Collection level is nearly the same as the [Asset Object in Items]((../../item-spec/item-spec.md#asset-object)), except for two differences. The `href` field is not required, because collections don't point to any data by themselves. Additioanlly the remaining fields, `title` and `type` are required in the Asset Definition, in order for it to adequately describe Item assets.
+
+| Field Name         | Type   | Description                                                  |
+| ------------------ | ------ | ------------------------------------------------------------ |
+| assets | Map<string, Asset Object | **REQUIRED.** A dictionary of assets that can be found in member Items |
+
+### Asset Object
+
+An asset is an object that contains details about the datafiles that will be included in member Items. Assets included at the Collection level do not imply that all assets are available from all Items. However, it is recommended that the Asset Definition is a complete set of all assets that may be available from any member Items.
```
m-mohr: I think it should say that the table below is not a complete list of members and can potentially include other properties, especially from extensions (bands from eo and sar).

matthewhanson:
@m-mohr I have resolved your comments

extensions/asset/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+| Field Name | Type   | Description                                                                           |
+| ---------- | ------ | ------------------------------------------------------------------------------------- |
+| title      | string | **REQUIRED.** The displayed title for clients and users.                                            |
```
m-mohr: A title is not required for assets in Items. So that's a bit contradicting... What should I do with assets that have no title?

extensions/asset/examples/example-landsat8.json
```diff
@@ -0,0 +1,240 @@
...
+    "id": "landsat-8-l1",
+    "title": "Landsat 8 L1",
+    "description": "Landat 8 imagery radiometrically calibrated and orthorectified using gound points and Digital Elevation Model (DEM) data to correct relief displacement.",
+    "keywords": [
+        "landsat",
+        "earth observation",
+        "usgs"
+    ],
+    "version": "0.1.0",
+    "stac_version": "0.7.0",
```
m-mohr: Update to 0.8.0 would make sense.

extensions/asset/json-schema/schema.json
```diff
@@ -0,0 +1,33 @@
...
+      },
+      {
+        "$ref": "#/definitions/asset"
+      }
+    ],
+    "definitions": {
+      "asset": {
+        "type": "object",
+        "required": [
+          "title",
```
m-mohr: See above regarding the required title.

extensions/asset/json-schema/schema.json
```diff
@@ -0,0 +1,33 @@
...
+    "allOf": [
+      {
+        "$ref": "../../../collection-spec/json-schema/collection.json#/definitions/core"
+      },
+      {
+        "$ref": "#/definitions/asset"
+      }
+    ],
+    "definitions": {
+      "asset": {
```
m-mohr: The JSON Schema is incorrect. The asset here is the schema name, but a property asset (which should be required) is missing.

m-mohr:
Thanks, I hadn't finished the review this morning, finished it now with some additional ToDos.

extensions/asset/json-schema/schema.json
matthewhanson: All the schemas I've done today are likely wrong. I'll make an attempt to fix.

extensions/asset/README.md
matthewhanson: Well it's not contradicting in that this Asset object is different. It does not require href (in fact that is not a field), and does require title and type.

The reasoning here is that if an "Asset Definition" is being provided than there should be some sort of label/description of each of the assets.

extensions/asset/examples/example-landsat8.json
matthewhanson: 👍

matthewhanson:
Made the changes requested, except I kept title as required, that's still open for discussion.

extensions/asset/json-schema/schema.json
m-mohr: They are still not validating. I took a stab and pushed an updated version of the JSON Schema for your convenience.

PS: When developing, I always use https://www.jsonschemavalidator.net/ to check JSON Schemas against the examples. :-)

extensions/asset/README.md
m-mohr: I still think it's a problem to require it here. A script that crawls Items and updated the Collection schemas automatically would fail to comply. It either doesn't have a title, leave the schema out or have to invent one. All three options are not good.

matthewhanson: Well, you shouldn't be creating the Collections schema from Items I don't think - what if you only have href defined?  You'll have no asset definition. Or are you ok with requiring type in this extension, in which case how is that different?

matthewhanson:
Ah cool, thanks for the fix, still figuring out this json-schema thing

extensions/asset/README.md
m-mohr: Oh, hell... When I was checking the Item spec on my phone yesterday night, I mixed up href and type and thought type would be required. You are right that type is not required, too. Well, in this case I wouldn't require anything, but I see that having empty objects doesn't help anybody. Hmm... I'll just leave this to others to decide, I think. ;-)

m-mohr: So isn't using this extension basically requiring titles and types in Items as the extension also says that all assets should be listed?

matthewhanson: Not sure if I understand what you are saying here.

This extension doesn't apply to the Item at all, so you can do whatever you want there.  

You also don't need to include *all* assets, although it is recommended. 

The idea here is that if your collection is going to have a definition of the assets it should probably include some minimum amount of information (that is a short description of the asset via `title`, and the type of file via `type`). If those things aren't present, or only one of them is, it's not that useful. 

Users should be able to use the Asset Definition to determine what assets they are interested in accessing.

extensions/asset/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+- [Example](examples/example-landsat8.json)
+- [JSON Schema](json-schema/schema.json)
+
+## Collection fields
+
+This extension introduces a single new field, `assets` at the top level of a collection.
+An Asset Object defined at the Collection level is nearly the same as the [Asset Object in Items](../../item-spec/item-spec.md#asset-object), except for two differences.
+The `href` field is not required, because collections don't point to any data by themselves.
+Additioanlly the remaining fields, `title` and `type` are required in the Asset Definition, in order for it to adequately describe Item assets.
```
cholmes: Minor type on 'Additionally'

extensions/asset/README.md
```diff
@@ -0,0 +1,34 @@
+# Asset Definition Specification
+
+**Extension [Maturity Classification](../README.md#extension-maturity): Proposal**
+
+A Collection extension to provide details about assets that are available in member Items.
```
cholmes: I think we should include more 'why' - what are the use cases for an implementor to make use of this. Not sure if this is the exact right place. But a bit more here would be good, and maybe flesh out more of it down below.

extensions/asset/README.md
cholmes: I think perhaps you need to explain more what this 'title' is for, that it is a short description of the asset for users to determine which assets they are interested in accessing. And maybe provide some more information on use than that. The 'The displayed title for clients and users.' doesn't really help people figure out what they should put here.

Issue #352, opened 2018-11-14
Add track and frame field in the EO/SAR extension


p3dr0:
These fields are defined as 

- track (or path or wrsLongitudeGrid) : string : A string providing the coordinate of the product on the across track (east/west) axis of the reference system used.
- frame (or row or wrsLatitudeGrid) : string : A string providing the coordinate of product along track (north/south) axis of the reference system used.

These are quite generically used in satellites following a polar or sun-synchronous orbit (ENVISAT, LANDSAT, Sentinel-1) so assigning them to a sensor specific extension would be quite cumbersome because we would need to replicate it across several ones. 

Another possibility would be to create an extension for satellite following a polar orbit.

cholmes:
In #350 @matthewhanson said: 

> track and frame don't apply to sensors not using a reference system, such as taskable satellites. I think it probably doesn't belong in the EO extension, but instead as sensor specific metadata, e.g., "landsat:wrs_path", "landsat:wrs_row".

I have been thinking of the EO extension as a set of fields that *can* be used, but don't have to be things that every single sensor will use. Constellation is one that doesn't apply to every satellite. 

I definitely don't think having each specific sensor make an extension for track and frame makes sense. If we don't want it in the EO extension directly then a little extension for satellites following a polar orbit could make a ton of sense. And if it gets wide usages we can evolve it to be a direct part of the EO spec.

matthewhanson:
Well that is certainly true @cholmes, they don't have to be used, and there are quite a few sensors that use some sort of x,y tiling for publication, so maybe it would be a good addition.

@p3dr0 I'm not sure track, frame, row, or path are all generic enough though.  Sentinel retiles their swaths into multiple scenes across track, Landsat has been distributed using other tiling schemes not path/row (WELD, the initial Landsat ARD beta). MODIS uses a horizontal/vertical tile grid.

What about something like "eo:xtile" and "eo:ytile". A provider could then provide a link to the GeoJSON of the reference grid if desired. Maybe 'tile' isn't a great term either though, to avoid confusion with web tiles.

m-mohr:
Generally seems to be a good idea, now we are bikeshedding a name. ;-)
Maybe `eo:grid: [x,y]` or similar? Or `eo:row` and `eo:column`?

fredliporace:
[...]

Issue #35, opened 2017-11-21
Consider z axis


cholmes:
We should consider if height is also part of geometries in STAC, and either explicitly say it's not or make sure that an extension mechanism can handle it, or that it's just ok with geojson definitions.

m-mohr:
I think that could be an important one if we want to cater for all domains especially such as meteorology.
For the datasets we specified a dimensions extension. I am not sure whether that really fulfills all needs yet. What about items?
I already heard voices within openEO that were not very comfortable with GeoJSON as it doesn't allow the z-axis to be specified.
Do we have any expert that we could consult regarding this?

matthewhanson:
GeoJSON does allow for an optional 3rd dimension to allow for altitude @m-mohr and this will be needed for a point cloud extension cc @hobu

m-mohr:
Oh, interesting, thanks for pointing that out. Never saw that in the wild yet, but you are certainly right I just looked it up directly in the spec/RFC.

m-mohr:
The z-axis has been added to the dataset spec with #239. We should consider adding similar things to the item.

matthewhanson:
There is no reason to not allow a 3rd dimension in the Item, and I think we do because it's a GeoJSON geometry which allows for it.
However, our description of Geometry and Bounding Box should be updated to say not just lat/lon, but lat/lon/<alttitude>

I'm moving this to 0.8.0 as it's just a clarification.

m-mohr:
Agreed, we should also check the descriptions for spatial collection extents.

PR #564, opened 2019-08-19
Issue #352: add z-axis to item spec for bbox and geometry

joshfix:
**Related Issue(s):** #35 

**Proposed Changes:**

1. Add documentation to the item spec in support of z-axis for geometry and bbox values.

**PR Checklist:**


m-mohr:
@joshfix I think there are potentially more places, which need an update:
* api-spec.md, line 62 is about bbox and is missing the z-axis ("[west, south, east, north]")
* To align between collection and item, I'd suggest to also update the collection-spec.md around line 67 (height vs. elevation).

Should a CHANGELOG entry be added? Not because it's really a change, but to make implementers aware of that they may have missed implementing z-axis support?

PR #567, opened 2019-08-20
Eo extension update readme

anayeaye:
**Related Issue(s):** #
N/A This PR addresses issue raised by @lossyrob in STAC/Lobby.

> The README here https://github.com/radiantearth/stac-spec/tree/dev/extensions, under the section “Prefixes”, makes it seem like extension prefixes (like eo) are for keys in the properties object of the Item (this is also how https://github.com/radiantearth/stac-spec/blob/dev/extensions/eo/examples/example-landsat8.json is set up). The examples in https://github.com/radiantearth/stac-spec/blob/dev/extensions/eo/README.md make it seem like they go direclty into the body of the Item object. Is this a misplacement in the latter? Is extension Item information explicitly supposed to go into the properties object of the Item?

**Proposed Changes:**

1. Modifies examples in Assets Object section of EO extension README to place `eo:bands` inside of item properties and corrects a linked example JSON.
2. Updates Prefixes section of extension README with `eo:gsd` in place of `eo:resolution`.

**PR Checklist:**

- [x] ~~I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-spec/blob/dev/CHANGELOG.md)~~ **or** a CHANGELOG entry is not required. Documentation changes only.
- [x] ~~API only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-spec/blob/dev/api-spec/README.md#openapi-definitions).~~ No API changes.

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -111,7 +111,7 @@ An example of this can be seen in a Landsat example:
     "eo:cloud_cover": 10.31,
     "eo:sun_azimuth": 149.01607154,
     "eo:sun_elevation": 59.21424700,
-    "eo:resolution": 30,
+    "eo:gsd": 30,
 
     "l8:data_type": "L1T",
     "l8:wrs_path": 153,
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -100,14 +100,14 @@ evolve to make fully resolved namespacing an option.
 
 An example of this can be seen in a Landsat example:
 
-```json
+```
   "properties": {
-	"datetime":"2018-01-01T13:21:30Z",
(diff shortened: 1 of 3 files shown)


m-mohr:
I made some further changes until I realized that this PR - unfortunately - bases on master, not dev. So it's likely that we are working on an outdated example and that's why stac_version, stac_extensions etc are missing. Therefore this PR needs to be done again, basing on dev, sorry.

anayeaye:
@m-mohr Sorry about that! I'll walk though it again and will add your changes as well

PR #577, opened 2019-09-05
Fix yaml definitions

scisco:
**Related Issue(s):** #
N/A

**Proposed Changes:**

1. Updates the open api schemas to update the `next` field definition and match the definitions in the markdown with the ones given in the yaml file

**PR Checklist:**

### DIFF
--- a/api-spec/openapi/STAC.yaml
+++ b/api-spec/openapi/STAC.yaml
@@ -116,8 +116,7 @@ components:
     limit:
       name: limit
       in: query
-      description:
-        $ref: '#/components/schemas/limit/description'
+      description: The maximum number of results to return (page size). Defaults to 10
       required: false
       schema:
         $ref: '#/components/schemas/limit'
@@ -126,8 +125,7 @@ components:
     next:
       name: next
       in: query
-      description:
-        $ref: '#/components/schemas/next/description'
+      description: The token to retrieve the next set of results, e.g., offset, page, continuation token
       required: false
       schema:
         $ref: '#/components/schemas/next'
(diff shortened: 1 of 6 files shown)


hgs-truthe01:
I'm a little confused on this one. I know we said we didn't want page to be part of the spec, but I didn't think we were adding next as a query string parameter? I thought it was intended to link to the next page using whatever parameter the implementer decided on and was returned in the search results. next as a parameter doesn't make a tone of sense to me. Do we need to specify it's format?

scisco:
@hgs-truthe01 I think `next` is only needed as a query parameter when you using the `GET /stac/search`. On `POST /stac/search` you will pass it in the body.

Nevertheless, my PR is only updating the yaml file to reflect what is already added here: https://github.com/radiantearth/stac-spec/blob/dev/api-spec/api-spec.md#filter-parameters-and-fields

api-spec/openapi/STAC.yaml
```diff
@@ -116,100 +116,60 @@ components:
...
+      name: next
       in: query
-      description: |
-        The optional page parameter returns the specified page of results
-        (with each page having size=limit).
-
-        * Minimum = 1
-        * Default = 1
+      description:
+        $ref: '#/components/schemas/next/description'
```
m-mohr: I guess this works in most openAPI implementations, but it is not allowed by openAPI. The description objects (and there are more similar references here) don't allow $refs to be used. Only strings are allowed.

m-mohr: The OpenAPI spec is not JSON Schema and works differently. $refs are not allowed by default, but must be allowed explicitly. See for example the [Components object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#componentsObject) where it explicitly states that Reference Objects (= $refs) are allowed. The descriptions are only of type string and don't mention Reference Object, therefore $refs are not allowed for descriptions.

scisco: @m-mohr updated my PR. Could you take a look again?

Issue #578, opened 2019-09-11
label extension: Label Classes should be list


lossyrob:
In the [New Item properties](https://github.com/radiantearth/stac-spec/tree/fb3320ca65bd96795306e7467aca39ed30b126e1/extensions/label#new-item-properties) section of the label spec, it lists `label:classes` as having type info `Class Object`. Based on the description, which describes "A Class Object defining the list of possible class names __for each__ `label:property`.", as well as existing examples, this type info should be  `[Class Object]` instead (be a list of class  objects, instead of a single instance.

The [JSON Schema](https://github.com/radiantearth/stac-spec/blob/fb3320ca65bd96795306e7467aca39ed30b126e1/extensions/label/schema.json#L22) entry is also incorrect.

PR #579, opened 2019-09-11
Move label:classes to be a list of classes.

lossyrob:
**Related Issue(s):** #578

**Proposed Changes:**

Move `label:classes` property to be a list of class objects.

**PR Checklist:**


lossyrob:
I'm unclear on if this needs a CHANGELOG entry - if so let me know and I'll add it.

m-mohr:
This must be included in 0.8. This just fixes what the examples already show: label:classes is an array. Not having this in 0.8 either leaves us with an invalid schema/spec Or invalid examples.

cholmes:
Ah, gotcha. If that's the case then I think we can just consider it a 'fix' and not have to do another release candidate.

lossyrob:
@cholmes Sorry for the incorrect base, I thought I made it off of dev... :shrug:

The 0.8.0-RC1 release notes have the label extension as 'added'; a 'fixed' changelog entry would make sense if there was a new RC, but if we aren't making another RC for this fix, it might not make sense to have both an 'added' and 'fixed' lines in the changelog for the same feature. But I can put one in if you still think it makes sense, or else I'll merge away.

m-mohr:
It could be added as "fixed" for 0.8.0. "Added" was 0.8.0-rc.1.

lossyrob:
@m-mohr Added changelog entry to the `unreleased` section of the CHANGELOG, which can be moved to 0.8.0 once it moves out of release candidate. If that seems good, looks good to merge - from a process standpoint, should I merge this PR or will one of you?

lossyrob:
From gitter conversation, this is good to go.

PR #594, opened 2019-10-10
Changed GeoTIFF type from `image/vnd.stac.geotiff` to `image/tiff; application=geotiff`

joshfix:
**Related Issue(s):** 
https://github.com/radiantearth/stac-spec/issues/535

**Proposed Changes:**

1. Changed GeoTIFF type from `image/vnd.stac.geotiff` to `image/tiff; application=geotiff`

**PR Checklist:**

- [x] This PR has **no** breaking changes.
- [x] I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-spec/blob/dev/CHANGELOG.md) **or** a CHANGELOG entry is not required.
- [ ] API only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-spec/blob/dev/api-spec/README.md#openapi-definitions).

Issue #535, opened 2019-07-23
GeoTiff media type


m-mohr:
As announced in https://github.com/opengeospatial/geotiff/issues/34#issuecomment-514078289, the official media type for GeoTiff will be `image/tiff; application=geotiff`. 

We should deprecate `image/vnd.stac.geotiff` and recommend the new one here: https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#media-types

The question is: What do we do with  `image/vnd.stac.geotiff; cloud-optimized=true`? That could be converted to `image/tiff; application=geotiff; cloud-optimized=true`, I guess?

philvarner:
I think `image/vnd.stac.geotiff; cloud-optimized=true` was always wrong, since there's nothing vendor specific to STAC about geotiffs.  I say we completely forget about it.

I like `image/tiff; application=geotiff; cloud-optimized=true` well enough and would fully support that being the media type.  Another option would be `image/tiff; application=geotiff; cog=true`, simply because it's shorter.

m-mohr:
> I think `image/vnd.stac.geotiff; cloud-optimized=true` was always wrong, since there's nothing vendor specific to STAC about geotiffs. I say we completely forget about it.

It was simply because there was nothing standardized. So we had to invent our own and had to specify a vendor (us). That's why STAC is in there. As I also found this wrong, I pushed the OGC GeoTiff guys to register one and they proposed the one I posted above.

> I like `image/tiff; application=geotiff; cloud-optimized=true` well enough and would fully support that being the media type. Another option would be `image/tiff; application=geotiff; cog=true`, simply because it's shorter.

I guess we should ask on the COG mailing list what they prefer. They may even come up with an official one.

m-mohr:
Opened a thread on the COG mailing list: http://osgeo-org.1560.x6.nabble.com/Media-type-td5411498.html

philvarner:
@m-mohr Sorry, that sounded more harsh than I meant it to be. I was trying to emphasize that it was chosen because there wasn't a better option, but there's no need to keep it around when a more standard option has emerged. Also, thanks for pushing OGC folks on this.

p3dr0:
I would suggest the usage of the profile link relation
`
       image/tiff; profile=<URI-reference>
`
the value <URI-reference> to be defined (for example it could be http://geotiff.osgeo.org/ ) 
further reading at 
https://tools.ietf.org/html/rfc6906#section-3.1 
and check also 
[...]

PR #595, opened 2019-10-11
Media types once again

m-mohr:
**Related Issue(s):** #594, #535 

**Proposed Changes:**

1. I think PR #594 was missing a deprecation notice containing the old media types.
2. Also, in the COG mailing list discussion it seemed that people were not super happy with a boolean flag, but rather would prefer it being a profile. Therefore proposing to changing from `cloud-optimized=true` to `profile=cloud-optimized`.

I think we should definitely merge point 1. Point 2 is open for discussions and I can revert that if nobody likes it.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Changed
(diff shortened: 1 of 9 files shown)


m-mohr:
I'd like to post this "proposal" as reply to the discussion on the COG mailing list, but I don't know how?!?

sgillies:
@m-mohr this is a good change :+1: , media type profiles are much more normal.

PR #598, opened 2019-10-11
Fix label extension examples

m-mohr:
**Related Issue(s):** None

**Proposed Changes:**

When doing #597, I found several issues in the label extension examples. I ran them all though a JSON Schema Validator and fixed them according to the Schema. Afterwards, all were also passed through a JSON Formatter.

**PR Checklist:**

### DIFF
--- a/extensions/label/examples/multidataset/zanzibar/znz001.json
+++ b/extensions/label/examples/multidataset/zanzibar/znz001.json
@@ -167660,11 +167660,18 @@
   },
   "properties": {
     "datetime": "2019-04-23T00:00:00Z",
-    "label:property": "[building, condition]",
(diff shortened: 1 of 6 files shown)


extensions/label/examples/multidataset/zanzibar/znz001.json
```diff
@@ -167682,15 +167689,17 @@
...
-    "label:overview": {
-      "property_key": "building",
-      "counts": [
-        {
-          "name": "yes",
-          "count": 4440
-        }
-      ]
-    }
+    "label:overview": [
```
lossyrob: Will these have to change again after the pluralization pr merges?

PR #608, opened 2019-10-15
Improve JSON Schema for Items

m-mohr:
**Related Issue(s):** None

**Proposed Changes:**

1. It's not a bug, but the oneOf is not needed in the schema. Therefore removed it for simplicity.
2. Removed the "geometry" and "type" as required property as it is already required in the GeoJSON Feature schema (and we do not redefine it in the schema, therefore it is referring to a non-existing property anyway).
3. Added "bbox" as property so that the "require" flag really works.
4. Fixed a typo: "updated"->"created" 

**PR Checklist:**

### DIFF
--- a/item-spec/json-schema/item.json
+++ b/item-spec/json-schema/item.json
@@ -146,7 +146,7 @@
                   }
                 },
                 "created": {
-                  "title": "Metadata updated at",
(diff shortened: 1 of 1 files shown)


m-mohr:
Need to redo it against dev instead of master. :-(

PR #614, opened 2019-10-24
Extension Tiled Assets

constantinius:
**Related Issue(s):** #

None

**Proposed Changes:**

1. Additional extension for tiled assets

**PR Checklist:**

**TODOS:**

- [ ] check/review alignment with other (especially `eo`) extensions
- [x] add examples
- [x] add schema
- [x] align with [OGC two dimensional tile matrix set specification](http://docs.opengeospatial.org/is/17-083r2/17-083r2.html)


constantinius:
This work-in-progress but any comments are very welcome.

@Schpidi @ungarj

m-mohr:
In general this seems like a good idea and thanks for pushing it forward. Personally, I would like to see this better aligned with STAC. At the moment it looks like it was simply converted from the OGC spec. For example, there is a title field, but title is already available in STAC and could probably be removed here. Similarly with abstract, keywords, identifier, type etc. Bounding Boxes could be aligned as well, it doesn't make much sense to introduce a new structure for bounding boxes if there's already one in STAC. Some fields probably need better descriptions. For example, what is a well-known scale set? Also, what fields are actually required for search and discovery and what fields are more for other use cases?

constantinius:
>In general this seems like a good idea and thanks for pushing it forward. Personally, I would like to see this better aligned with STAC. At the moment it looks like it was simply converted from the OGC spec. For example, there is a title field, but title is already available in STAC and could probably be removed here. Similarly with abstract, keywords, identifier, type etc. Bounding Boxes could be aligned as well, it doesn't make much sense to introduce a new structure for bounding boxes if there's already one in STAC.

It is true, this extension directly uses the JSON schema defined for the TMS spec. I knew this was likely a point of discussion, but I initially decided to go this route first.

I think I could slim down the schema and to better align with the STAC way to do things.

>Could the dimensions be aligned with the data cube extension?

Yes, I think this is feasible. But I think it would be bad to directly reference the datacube extension constructs.

>Some fields probably need better descriptions. For example, what is a well-known scale set? 

Yes, this could be extended. But I think I will wait until the structure has somewhat settled. But on the other hand I don't want to retell the TMS spec. Do you think it is okay to reference the TMS spec at some vital points?

extensions/tiled-assets/examples/example-dimension.json
```diff
@@ -0,0 +1,234 @@
...
+                ],
+                [
+                    14.326257705688475,
+                    48.33992958068847
+                ]
+            ]
+        ]
+    },
+    "properties": {
+        "datetime": "2018",
```
m-mohr: This is not a valid date-time, must conform to RFC3339 date-time.

m-mohr: This is not a valid date-time, must conform to RFC3339 date-time.
I guess you should use the datetime-range extension here.

extensions/tiled-assets/examples/example-dimension.json
```diff
@@ -0,0 +1,234 @@
...
+                "2018-04-02_2018-04-17",
+                "2018-04-18_2018-05-02"
+            ]
+        }
+    },
+    "assetTemplates": {
+        "B02": {
+            "hrefTemplate": "http://cloudless.eox.at/sen2l2a-cog-converted/{timeSpan}_s2_l2a/{timeSpan}_s2_l2a_B02.tif",
+            "title": "",
+            "type": "image/vnd.stac.geotiff",
```
m-mohr: The media type has been replaced by image/tiff; application=geotiff

extensions/tiled-assets/examples/example-dimension.json
```diff
@@ -0,0 +1,234 @@
...
+        }
+    },
+    "links": [
+        {
+            "rel": "self",
+            "href": "http://cloudless.eox.at/sen2l2a-cog-converted.json"
+        },
+        {
+            "rel": "root",
+            "href": "http://cloudless.eox.at/stack.json"
```
m-mohr: Very minor: I guess that was meant to be named stac instead of stack ;-)

extensions/tiled-assets/README.md
```diff
@@ -0,0 +1,127 @@
...
+| lowerCorner | [number] | **REQUIRED.** The lower corner of this bounding box.               |
+| upperCorner | [number] | **REQUIRED.** The upper corner of this bounding box.               |
+
+### Tile Matrix Object
+
+The description of a single tile matrix (zoom level) of the tile pyramid.
+
+| Field Name          | Type                                                          | Description                                                                                                                                                       |
+| ------------------- | ------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| type                | string                                                        | **REQUIRED.** Type of the Tile Matrix. Fixed to `TileMatrixType`                                                                                                  |
```
m-mohr: A field which only allows a single value seems note very useful and could probably be omitted.

extensions/tiled-assets/README.md
```diff
@@ -0,0 +1,127 @@
...
+| title               | string                                                        | Title of this tile matrix set, normally used for display to a human                                                                                               |
```
m-mohr: title, abstract and keywords are probably better directly put in the Item properties or in a collection. title even has a well-defined field already. abstract and keywords are available in collections.

extensions/tiled-assets/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+**steps:** This is meant as a total number of steps. Where the first value is always the `start` and the last step is always the `stop` value.
+
+## Item fields
+
+| Field Name         | Type                                                         | Description                                                  |
+| ------------------ | ------------------------------------------------------------ | ------------------------------------------------------------ |
+| assetTemplates     | Map<string, [Asset Template Object](#asset-template-object)> | An array of Asset Template objects that denote templated arguments | 
+
+### Asset Template Object
```
m-mohr: This looks like it could use the same structure as an asset, but allow replacing variables in hrefs.

extensions/tiled-assets/README.md
```diff
@@ -0,0 +1,127 @@
...
+| matrixWidth         | number                                                        | **REQUIRED.** The number of columns in this tile matrix. Any number between zero and `matrixWidth - 1` can be used to replace the `{TileCol}` template parameter. |
+| matrixHeight        | number                                                        | **REQUIRED.** The number of rows in this tile matrix. Any number between zero and `matrixHeight - 1` can be used to replace the `{TileRow}` template parameter.   |
+| variableMatrixWidth | [Variable Matrix Width Object](#variable-matrix-width-object) | Describes the rows that has variable matix width                                                                                                                  |
+| pixelBuffer         | [Pixel Buffer Object](#pixel-buffer-object)                   | An optional default pixel buffer description object, overriding the default of the pyramid. By default, no pixel buffers are used.                                |
```
m-mohr: Do you really need to know the pixel buffer for downloading the assets?
Similar with many of the fields: What is needed to just find and get the data? Could we reference external metadata (e.g. a static TMS metadata document or so) and only provide the essential fields here?

extensions/tiled-assets/README.md
```diff
@@ -0,0 +1,127 @@
...
+| title             | string                                             | Title of this tile matrix set, normally used for display to a human                                                          |
+| abstract          | string                                             | Brief narrative description of this tile matrix set, normally available for display to a human                               |
+| keywords          | [string]                                           | Unordered list of one or more commonly used or formalized word(s) or phrase(s) used to describe this dataset                 |
+| identifier        | string                                             | **REQUIRED.** Tile matrix set identifier. This can be used as the `{TileMatrixSet}` template parameter.                      |
+| boundingBox       | [BoundingBox Object](#bounding-box-object)         | Minimum bounding rectangle surrounding the tile matrix set, in the supported CRS                                             |
```
m-mohr: Isn`t the bounding box already specified in the item?

m-mohr:
I made a review with specific comments so it is easier to discuss specific points.

> Yes, I think this is feasible. But I think it would be bad to directly reference the datacube extension constructs.

We may also align the data cube extension. In the end it would be great to just not reinvent the wheel multiple times.

> Yes, this could be extended. But I think I will wait until the structure has somewhat settled. But on the other hand I don't want to retell the TMS spec. Do you think it is okay to reference the TMS spec at some vital points?

I like self-contained extensions, but if it is too complex to add here, links are fine, I think.

notthatbreezy:
I am interested in implementing the proposal here and would like to help make changes so that it could make it in the specification as an extension as we are in the process of implementing tiled assets in some of our STAC services at Azavea.

What's the best way to help out with this proposal? Do you have plans to update this PR @constantinius or should I open a new one?

constantinius:
Thanks for helping out, @notthatbreezy, I am indeed in a bit of a struggle to find the time to further develop this spec extension.

What has become clear from the reviews by @m-mohr is that it is to complex and too much leaning on the TMS side instead of being more aligned with STAC. Also it seems to be too unclear what is actually necessary and what is not.

As a first step, I would slim down the spec and remove fields that are not useful (such as these maligned `type` fields, title, abstract, etc).

It was also requested to not duplicate the `Dimension` objects from the `Datacube` extension. So, imho it would be best to completely remove them from this spec and make a reference the `dc` spec objects instead.

One important thing last: the tile matrix set should probably not be included in the stac item itself, but be referenced from somewhere else. (Either a URL or from the collection/catalog). Then there is the thing about the `wellKnownScaleSets` which are basically predefined tile matrix sets in the TMS spec. This would be the third option. And *then*: in a tile set referencing a tile matrix, one can specify the actual limits within this tile matrix set. e.g: specifying the actual zoom levels and actual tile limits.

I think this is demonstrated best by an example: the `Web Mercator Quad TileMatrixSet definition ` defines a global tile grid with 24 tile matrices (zoom levels). If you have a dataset in only a part of the world, you can still align it to the global grid, but limit the zoom levels and then limit the tiles in that matrix of the ones actually intersecting with the data.

This is concept is not yet reflected in any way in the extension and would be rather important.

I think it is better to stay on this PR, but contributions are very welcome!

This turned out to be more  than expected. I hope we can collaborate on this (maybe on gitter?). Please let me know what you think.

constantinius:
@m-mohr, others

I reworked this extension quite a bit and wanted to know your thoughts.

Man focus points:
 * it should be more aligned to STAC now (removed duplicate concepts, parameter naming convention, ...)
 * the whole `tl:tileMatrixSet` object description was removed and now refers to the online docs and schemas of the two dimensional tile matrix spec (http://docs.opengeospatial.org/is/17-083r2/17-083r2.html). The focus now lies on referencing existing tile matrices instead of embedding your own (which is still possible). These tile matrices are now linked using the `Tile Matrix Link` objects, where one can reference such a tile matrix, but can apply modifications such as limits or pixel buffers.
 * the concept of dimension was removed, and now it refers to the datacube extension instead.

Please let me know if this is going into the right direction.

extensions/tiled-assets/README.md
```diff
@@ -0,0 +1,138 @@
...
+| ------------------- | ------------------------------------------------------------ | ------------------------------------------------------------------ |
+| asset_templates     | Map<string, [Asset Template Object](#asset-template-object)> | An array of Asset Template objects that denote templated arguments |
+
+### Asset Template Object
+
+An asset template is an object that contains a link template to data associated with the Item that can be downloaded or streamed. It is allowed to add additional fields.
+
+| Field Name    | Type   | Description                                                                           |
+| ------------- | ------ | ------------------------------------------------------------------------------------- |
+| href_template | string | **REQUIRED.** Link template to the asset object. Relative and absolute links are both allowed. All parts of the template within curly braces `{}` are meant to be template substitution parameters to be replaced |
```
m-mohr: Is this compatible with https://tools.ietf.org/html/rfc6570? Should it be mentioned?

m-mohr:
@constantinius Thanks. Could you resolve the comments I made if you have tackled them? That would give me a better overview of what has been changed and what is still open and may need further discussions.

extensions/tiled-assets/README.md
constantinius: I was not aware of this RFC. It actually looks very promising, and we should definitely find a way to incorporate it.

The only issue I see here is with the use of "namespaced" template parameters, such as `eo:bands` which is not permitted with the RFC (there the colon separates the variable from its `max-length` field). Do you have a suggestion for that issue?

extensions/tiled-assets/README.md
constantinius: In the latest version there is no bounding box definition in the extension spec anymore. For inline tile matrix definitions the bounding box from the 2D TMS schema can still be used.

extensions/tiled-assets/README.md
m-mohr: I haven't looked into the RFC yet, but it seemed relevant so linked it here. If there are big burdens of adoption, we may skip it.

extensions/tiled-assets/README.md
constantinius: The pixel buffer is necessary, as it can actually displace the data, for that amount of pixels. It is optional anyhow, so I would really love to have it in the spec, but if you say it has nothing to do in that, we can still remove it.

extensions/tiled-assets/README.md
constantinius: So instead of re-defining it in this spec, just refer to the Asset definition? Thats also possible. In this spec it just has a different field name `href_template` to make it clear that it is meant as a template.

extensions/tiled-assets/README.md
constantinius: This was removed with the re-treading of the 2D TMS spec.

extensions/tiled-assets/README.md
constantinius: Agreed. This was removed with the re-treading of the 2D TMS spec.

extensions/tiled-assets/examples/example-dimension.json
constantinius: Fixed.

extensions/tiled-assets/examples/example-dimension.json
constantinius: Done. I cannot find the datetime range extension anymore, what has happened to it?

constantinius:
@m-mohr I have answered to the inline comments now.

extensions/tiled-assets/examples/example-dimension.json
constantinius: 👍

extensions/tiled-assets/examples/example-dimension.json
m-mohr: It is part of the core, see https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#date-and-time

extensions/tiled-assets/README.md
```diff
@@ -0,0 +1,138 @@
...
+| title         | string | The displayed title for clients and users.                                            |
+| type          | string | [Media type](../README.md#media-types) of the asset.                                  |
```
m-mohr: This link is broken.

extensions/tiled-assets/README.md
m-mohr: Yes, I'd do so. Reasoning is that changes to assets would not get reflected here. You can already see this as the roles property is missing, which has been added with 0.9 and thus this definition is now outdated and not aligned with the current STAC Item spec any longer. You may even put the template into `href` directly, but that's probably to be discussed with more people.

extensions/tiled-assets/README.md
m-mohr: This link is broken.

extensions/tiled-assets/README.md
m-mohr: STACs primary focus is search and discovery. Are pixel buffers useful for that (i.e. displaying it on a web map)? I'm not saying it must be removed, maybe it helps to clarify in the docs what it is useful for and why it's part of STAC with regard to search and discovery.

extensions/tiled-assets/README.md
constantinius: I fixed the link.

I see your point. Regarding directly linking the `Asset`: we would need put the template into the `href`, as it is mandatory, but I don't see the problem in that, to be honest.

m-mohr: Nevermind, the issue I saw yesterday based on a wrong assumption. So linking directly to asset should be the best option, I think.

extensions/tiled-assets/examples/example-dimension.json
constantinius: Thanks! (It was in the extensions, so I was confused). I updated the examples to also incorporate the datetime range.

extensions/tiled-assets/examples/example-dimension.json
```diff
@@ -0,0 +1,239 @@
+{
+    "stac_version": "0.8.0",
```
m-mohr:
```suggestion
    "stac_version": "0.9.0",
```
I guess this should be updated to be "0.9.0".We are still in 0.9.0-rc.1, but I guess it won't get into STAC before 0.9.0.

extensions/tiled-assets/examples/example-dimension.json
```diff
@@ -0,0 +1,239 @@
...
+                ]
+            ]
+        ]
+    },
+    "properties": {
+        "datetime": "2018-01-01T00:00:00Z",
+        "start_datetime": "2018-01-01T00:00:00Z",
+        "end_datetime": "2018-12-31T23:59:59Z",
+        "eo:gsd": 10,
+        "eo:platform": "sentinel-2",
```
m-mohr:
```suggestion
        "platform": "sentinel-2",
```
Field was moved to core.

extensions/tiled-assets/examples/example-dimension.json
```diff
@@ -0,0 +1,239 @@
...
+        "eo:instrument": "msi",
```
m-mohr:
```suggestion
        "eo:instruments": [
            "msi"
        ],
```
Field was moved to core and is an array now.

extensions/tiled-assets/examples/example-dimension.json
```diff
@@ -0,0 +1,239 @@
...
+        "eo:bands": [
+            {
+                "name": "B2",
+                "common_name": "blue",
+                "gsd": 10
```
m-mohr: gsd is not available in bands any longer.

extensions/tiled-assets/examples/example-dimension.json
```diff
@@ -0,0 +1,239 @@
...
+                    "2018-04-02_2018-04-17",
+                    "2018-04-18_2018-05-02"
+                ]
+            }
+        }
+    },
+    "asset_templates": {
+        "B02": {
+            "href_template": "http://cloudless.eox.at/sen2l2a-cog-converted/{cube:dimensions:timeSpan}_s2_l2a/{cube:dimensions:timeSpan}_s2_l2a_B02.tif",
+            "title": "",
```
m-mohr: Empty titles can also just be removed.

extensions/tiled-assets/README.md
```diff
@@ -0,0 +1,138 @@
+# Tiled Assets specification
+
+**Extension [Maturity Classification](../README.md#extension-maturity): Proposal**
+
+An extension to allow the specification of tiled assets within STAC Items. With this extension it is possible to allow the description of assets using template references and rules to construct those. This extension is tailored for cases where there are too many assets associated with an Item to be practically listed in the `assets` property. For this reason, the new `asset_templates` property allows to specify template URLs where components can be replaced to get the final URLs to the actual files.
+
+This extension is modelled in close alignment to the [OGC Two Dimensional Tile Matrix Set](http://docs.opengeospatial.org/is/17-083r2/17-083r2.html)
```
m-mohr: The extension should tell what other extensions and/or fields may be useful to be used and why (also should have links). For example start_datetime, end_datetime, datacube, ...

extensions/tiled-assets/examples/example-tiled.json
```diff
@@ -0,0 +1,266 @@
+{
+    "stac_version": "0.8.0",
```
m-mohr:
```suggestion
    "stac_version": "0.9.0",
```

extensions/tiled-assets/examples/example-tiled.json
```diff
@@ -0,0 +1,266 @@
...
+            ]
+        ]
+    },
+    "properties": {
+        "collection": "s2cloudless",
+        "datetime": "2018-01-01T00:00:00Z",
+        "start_datetime": "2018-01-01T00:00:00Z",
+        "end_datetime": "2018-01-01T00:00:00Z",
+        "eo:gsd": 10,
+        "eo:platform": "sentinel-2",
```
m-mohr:
```suggestion
        "platform": "sentinel-2",
```

extensions/tiled-assets/examples/example-tiled.json
```diff
@@ -0,0 +1,266 @@
...
+        "eo:instrument": "msi",
```
m-mohr:
```suggestion
        "eo:instruments": [
            "msi"
        ],
```

extensions/tiled-assets/examples/example-tiled.json
```diff
@@ -0,0 +1,266 @@
...
+        "eo:bands": [
+            {
+                "name": "B2",
+                "common_name": "blue",
+                "gsd": 10
```
m-mohr: gsd is not part of bands any longer.

extensions/tiled-assets/examples/example-tiled.json
```diff
@@ -0,0 +1,266 @@
...
+                        "matrixHeight": 16
+                    }
+                ]
+            }
+        }
+    },
+    "asset_templates": {
+        "bands": {
+            "href_template": "http://cloudless.eox.at/s2cloudless_2018/{TileMatrixSet}/{TileRow}/{TileCol}.tif",
+            "title": "",
```
m-mohr: You don't need to provide empty titles.

extensions/tiled-assets/README.md
ungarj: Hi @m-mohr!
Well the reasoning behind storing tiles with overlaps/buffers is processing performance in certain use cases. There are some tile-based processes which require buffers around the processing tiles (e.g. moving window filters such as hillshade) otherwise artifacts are introduced at tile borders.
Now if such a process uses the same tiling scheme as the input dataset and the input dataset does not have tiles with buffers this means that 9 input tiles have to be read (the one in the center matching with the process tile and the 8 neighbours around) but if the tiles are already stored with buffers, the process only has to read one tile.
I agree, for only downloading the assets having to know the pixelbuffer is not essential. It could however introduce confusion if one wanted to determine the exact bounds of a tile using a given tile index and not the metadata stored in the GTiff. If there is no knowledge of pixelbuffers, the bounds would be off by the pixelbuffer width and height.
I would also argue that in many cases the assets are probably used directly as a input of a processing tool and in that case pixelbuffers are an optimized data structure for the beforementioned use cases.

m-mohr: Thank you, @ungarj. I fully understand your point and I often suffer the same thing with STAC. We also use it for processing tasks and need to store related data in STAC, but it's not meant to store such data as the focus was defined to be search and discovery.
@cholmes How to proceed in such cases?

extensions/tiled-assets/examples/example-dimension.json
constantinius: 👍

extensions/tiled-assets/examples/example-dimension.json
constantinius: Reading the spec I that should be "instruments" not "eo:instruments" (as in common metadata).

extensions/tiled-assets/examples/example-dimension.json
constantinius: 👍

extensions/tiled-assets/examples/example-dimension.json
constantinius: 👍

extensions/tiled-assets/examples/example-tiled.json
constantinius: 👍

extensions/tiled-assets/examples/example-tiled.json
constantinius: See above.

extensions/tiled-assets/examples/example-tiled.json
constantinius: 👍

extensions/tiled-assets/examples/example-tiled.json
constantinius: 👍

extensions/tiled-assets/examples/example-dimension.json
constantinius: 👍

extensions/tiled-assets/README.md
constantinius: I read that spec in detail now. I think it is very interesting, but might not be all too useful for our purposes. The spec states that there are other mechanisms of templating, specifically mentioning OpenSearch and WSDL. 

Also the WMTS spec defines a simple templating mechanism, which this extension actually builds upon. So I'd rather reference that, if it is okay with you. It can be found here: http://docs.opengeospatial.org/is/13-082r2/13-082r2.html#17

extensions/tiled-assets/examples/example-dimension.json
m-mohr: Indeed.

extensions/tiled-assets/examples/example-tiled.json
m-mohr: Suggestion fixed.

extensions/tiled-assets/README.md
cholmes: I think I've been consistent on my take in cases when people want to put in non-search metadata - STAC is focused on search and discovery, but we provide extension mechanisms for people to add in what they want. So I'm not opposed to extensions that do more of that type of stuff. 

The eventual line I think will be that 'common metadata' should just be things for search. But there can be a rich ecosystem of extensions that people use for more than the core. Since this is just a proposed extension for now I think it's fine. If it evolves to be part of common metadata then it could make sense to remove it when we fold that in - but I think there are quite a few steps to get there.

constantinius:
@cholmes @m-mohr 

I updated the schema and I tried to better explain the concepts behind the tiled assets. Please have a look at that and let me know if you think that there is something missing.

m-mohr:
Just a minor thing and mostly a matter of taste, but why not use the prefix tiles or tile for the extension? tl seems not very intuitive.
Instead of for example `tl:tile_matrix_sets` it would simply be `tiles:matrix_sets` or `tile:matrix_sets`.

constantinius:
@m-mohr Thats a good point. I'll change it to `tiles`.

@notthatbreezy Have you had a chance to look at this PR? Does it also suit your requirements?

extensions/README.md
```diff
@@ -60,6 +60,7 @@ An extension can add new fields to STAC entities (content extension), or can add
 | [Scientific](scientific/README.md) (`sci`)                   | Item, Collection | Scientific metadata is considered to be data that indicate from which publication data originates and how the data itself should be cited or referenced. | *Proposal* |
 | [Single File STAC](single-file-stac/README.md) (-)           | ItemCollection   | An extension to provide a set of Collections and Items as a single file catalog. | *Proposal* |
 | [Versioning Indicators](version/README.md) (-)               | Item, Collection | Provides fields and link relation types to provide a version and indicate deprecation. | *Proposal* |
+| [Tiled Assets](tiled-assets/README.md) (`tl`)                | Item             | Allows to specify numerous assets using asset templates via tile matrices and dimensions. | *Proposal* |
```
m-mohr: It's still tl here. Also there are merge conflicts for this file.

extensions/tiled-assets/README.md
```diff
@@ -0,0 +1,129 @@
+# Tiled Assets specification
+
+**Extension [Maturity Classification](../README.md#extension-maturity): Proposal**
```
m-mohr: Could you align this header section to how we have it now in dev for all other extension, please? With Field Name Prefix etc?

constantinius: 👍

extensions/README.md
constantinius: Sorry, missed that.
I merged dev and resolved the conflicts now.

constantinius:
Done. The CI checks seem to be unrelated.

extensions/README.md
```diff
@@ -37,30 +37,31 @@ column is the main gating function for an extension to mature. But extension aut
...
+| [SAR](sar/README.md)                           | sar              | sar                 | Item                      | *Proposal* | Covers synthetic-aperture radar data that represents a snapshot of the earth for a single date and time. |
+| [Satellite](sat/README.md)                     | sat              | sat                 | Item                      | *Proposal* | Satellite related metadata for data collected from satellites. |
+| [Scientific](scientific/README.md)             | scientific       | sci                 | Item, Collection          | *Proposal* | Scientific metadata is considered to be data that indicate from which publication data originates and how the data itself should be cited or referenced. |
+| [Single File STAC](single-file-stac/README.md) | single-file-stac | -                   | ItemCollection            | *Proposal* | An extension to provide a set of Collections and Items as a single file catalog. |
+| [Tiled Assets](tiled-assets/README.md)         | tiles            | tiles               | Item                      | *Proposal* | Allows to specify numerous assets using asset templates via tile matrices and dimensions. |
```
m-mohr: The identifier is the folder name.

extensions/tiled-assets/README.md
```diff
@@ -0,0 +1,133 @@
+# Tiled Assets specification
+
+- **Title: Tiled Assets**
+- **Identifier: tiles**
```
m-mohr: The identifier is the folder name.

extensions/tiled-assets/examples/example-dimension.json
```diff
@@ -0,0 +1,220 @@
+{
+    "stac_version": "0.9.0",
+    "stac_extensions": [
+        "datacube",
+        "eo",
+        "tiles"
```
m-mohr: The value to use here (and probably also in all other examples) is the identifier/folder name.

item-spec/json-schema/item.json
```diff
@@ -79,6 +79,7 @@
                       "sar",
                       "sat",
                       "scientific",
+                      "tiles",
```
m-mohr: The value to use here (and probably also in all other examples) is the identifier/folder name.

extensions/README.md
constantinius: 👍

extensions/tiled-assets/examples/example-dimension.json
constantinius: 👍

extensions/tiled-assets/README.md
constantinius: 👍

item-spec/json-schema/item.json
constantinius: 👍

extensions/tiled-assets/README.md
```diff
@@ -0,0 +1,133 @@
...
+When a very large dataset is split into many tiles it is very impractical to list all possible files with the `assets` property.
+
+This extension allows the specification of tiled assets within STAC Items. With this extension it is possible to allow the description of assets using template references and rules to construct those in order to get the tiles the user is interested in. The new `asset_templates` property allows to specify template URLs where components can be replaced to get the final URLs to the actual files.
+
+This extension is modelled in close alignment to the [OGC Two Dimensional Tile Matrix Set](http://docs.opengeospatial.org/is/17-083r2/17-083r2.html). The main access approach and the templating mechanism used is defined in the [OGC WMTS Simple Profile](http://docs.opengeospatial.org/is/13-082r2/13-082r2.html).
+
+- Examples: [Tiled](examples/example-tiled.json), [Dimension](examples/example-dimension.json)
+- [JSON Schema](json-schema/schema.json)
+
+## Item, Collection and Catalog properties
```
m-mohr: Here is says that the field applies to Catalogs and Collections, too, but the scope and JSON schema only mention items?! 🤔

constantinius:
I see your point, tiled items are in of itself some kind of collection of tiles. The issue remains that we do not want an actual STAC Item representing a simple tile (due to the sheer number and repetitive nature of the metadata). When using asset_templates on Collection level this would mean that the collection would be devoid of Items, which is quite a different concept I think.  But I agree, that Items with tiled assets are somewhat between Items and Collections. But I would argue that they are closer to Items than Collections.

extensions/tiled-assets/README.md
constantinius: Oh, yes, good point. The `tile_matrix_sets` can be located at Collection/Catalog level. So I'll add this in the readme and schemas as-well.

Everything else would remain on Item level.

constantinius:
@m-mohr 
I updated the header and schema to allow `tiles:tile_matrix_set` descriptions on Collections and Catalogs as suggested in the Readme. My json-schema skills are questionable at best, so please have a look at it and tell me if it is sound.

Thanks for the feedback, really appreciated!

notthatbreezy:
This looks good to me - my own requirements have shifted slightly (we're only concerned with visual tiles at the moment so have shifted to the [OGC API Tiles Spec](https://github.com/opengeospatial/OGC-API-Tiles/)) so this isn't an immediate need for us at the moment, but I think this would be useful for us when the need arises. I think the extension is thorough and will be useful.

extensions/tiled-assets/json-schema/schema.json
```diff
@@ -3,16 +3,49 @@
...
-  "allOf": [
+  "oneOf": [
     {
-      "$ref": "../../../item-spec/json-schema/item.json#/definitions/core"
-    },
-    {
-      "$ref": "#/definitions/tiles"
+      "allOf": [
+        {
+          "$ref": "../../../item-spec/json-schema/item.json#/definitions/core"
```
m-mohr:
```suggestion
          "$ref": "../../../item-spec/json-schema/item.json"
```

extensions/tiled-assets/json-schema/schema.json
```diff
@@ -3,16 +3,49 @@
...
+      ]
     }
   ],
   "definitions": {
-    "tiles": {
+    "catalog-collection-tiles": {
+      "title": "Tile matrix additions to STAC Collections and Catalogs",
+      "description": "Allows the definition of tile matrix sets on a Collection or Catalog.",
+      "type": "object",
+      "required": [],
```
m-mohr: I think "tiles:tile_matrix_set" should be required here. Otherwise validation would succeed when nothing is specified, which seems wrong is you specify that a collection/catalogs supports this extension but no field is available.

By the way, I just realized nothing in the extension is actually required. What is the minimum set a user needs to specify? That should be required then....

extensions/tiled-assets/README.md
```diff
@@ -3,7 +3,7 @@
 - **Title: Tiled Assets**
 - **Identifier: tiled-assets**
 - **Field Name Prefix: tiles**
-- **Scope: Item**
+- **Scope: Item, Catalog, Collection**
```
m-mohr: The scope in the extensions overview table must be updated, too.

In the JSON Schemas for collections and catalogs you must add the "tiled-assets" value to the stac_extensions enum.

extensions/tiled-assets/json-schema/schema.json
constantinius: > I think "tiles:tile_matrix_set" should be required here. Otherwise validation would succeed when nothing is specified, which seems wrong is you specify that a collection/catalogs supports this extension but no field is available.

I also wanted to allow the definition of the tile matrix sets on the `Collection` or `Catalog` level, similarly to the `eo:bands`. Does this make sense?

> By the way, I just realized nothing in the extension is actually required. What is the minimum set a user needs to specify? That should be required then....

Hm, I think this would be the `asset_templates`, as without it it would be plain `Items` instead. I'll update the schema.

extensions/tiled-assets/README.md
constantinius: I'll do that, thanks!

constantinius: Hm, there is no extensions enumeration in the `catalog.json`.

extensions/tiled-assets/json-schema/schema.json
constantinius: Okay, I think you were right. The extensions to the `Catalog` and `Collection` mean that the only property `tiles:tile_matrix_set` should be required. I changed that.

I also made `asset_templates` mandatory for items.

I think this can be resolved now, what do you think @m-mohr?

extensions/tiled-assets/README.md
m-mohr: Indeed, I'm not 100% sure why we did that. I guess it should look like in the collection.json?! Anyway, so that's nothing we need to change here.

extensions/tiled-assets/json-schema/schema.json
m-mohr: The schema looks good, but the markdown file / README doesn't reflect the required fields as far as I can see...

constantinius: Thats right. I've added those now.

m-mohr:
So it has been merged and thus is scheduled to be released with the next version. :-)
Thanks @constantinius !

constantinius:
Thanks a lot! :-)

PR #625, opened 2019-10-31
Checksum extension: use Multihash

### LINKED ISSUE #503: Using self-describing hashes for Checksum Extension
There's a project called [Multihash](https://github.com/multiformats/multihash) which specifies self-identifying hashes. With bundling the hash function directly with the hash, you wouldn't need to specify it within the JSON. So you would have a single field called `checksum` which contains the hash.

Example: Checksum for a text file with file content `test`.

| Algorithm               | Example                                                      |
| ------------------------ | ------------------------------------------------------------ |
| sha1 (160 bits)            | `1114a94a8fe5ccb19ba61c4c0873d391e987982fbbd3`                           |
| sha2 (256 bits)            | `12209f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08`                   |
| sha2 (256 bits truncated to 160 bits) | `12149f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b` |

vmx:
**Related Issue(s):** #503 

**Proposed Changes:**

 extension(checksum): update checksum extension to use Multihash

Instead of defining the hashing algorithm within the key, use a self-identifiable hash
([Multihash]). This way the stac spec doesn't need to introduce a long list of supported
hash algorithms, but can rely on the ones specified in [Multicodec].

I also changed the examples to use BLAKE2 instead of MD5. While MD5 is still in wide
use, I want to encourage people to update to newer hash function, which makes also
sense if you want to check the integrity of untrusted sources.

[Multihash]: https://github.com/multiformats/multihash
[Multicodec]: https://github.com/multiformats/multicodec/

Closes #503.

**PR Checklist:**


extensions/checksum/README.md
```diff
@@ -2,30 +2,29 @@
...
-| ------------- | ------ | ------------------------------------------------------------ |
-| checksum:md5  | string | [MD5 message-digest algorithm](https://en.wikipedia.org/wiki/MD5) 128-bit checksum for the corresponding file, encoded as hexadecimal string. |
-| checksum:sha1 | string | [SHA-1 (Secure Hash Algorithm 1)](https://en.wikipedia.org/wiki/SHA-1) 160-bit checksum for the corresponding file, encoded as hexadecimal string. |
-| checksum:sha2 | string | [SHA-2 (Secure Hash Algorithm 2)](https://en.wikipedia.org/wiki/SHA-2) checksum for the corresponding file with varying sizes of 224, 256, 384 or 512 bits, encoded as hexadecimal string. |
-| checksum:sha3 | string | [SHA-3 (Secure Hash Algorithm 3)](https://en.wikipedia.org/wiki/SHA-3) checksum for the corresponding file with varying bit sizes, encoded as hexadecimal string. |
+| Field Name | Type   | Description                                                  |
+| ---------- | ------ | ------------------------------------------------------------ |
+| checksum   | string | Multihash for the corresponding file, encoded as hexadecimal (base 16) string with lowercase letters. |
```
m-mohr: I think I'd prefer the field name to be `checksum:multihash` for three reasons:
1. We have a prefix (checksum) as for all other extensions.
2. We could extend it later (e.g. having checksum:newhash)
3. From the field name it is clear how the hash is "named" without needing to read the docs.

vmx:
I just rebased it on-top of the current `dev` branch.

extensions/checksum/README.md
```diff
@@ -2,30 +2,29 @@
...
 
 Checksum for a text file with file content `test`.
 
-| Field Name               | Example                                                      |
-| ------------------------ | ------------------------------------------------------------ |
-| checksum:md5             | `098f6bcd4621d373cade4e832627b4f6`                           |
-| checksum:sha1            | `a94a8fe5ccb19ba61c4c0873d391e987982fbbd3`                   |
-| checksum:sha2 (256 bits) | `9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08` |
-| checksum:sha3 (256 bits) | `36f028580bb02cc8272a9a020f4200e346e276ae664e45ee80745574e2f5ab80` |
+| Field Name         | Algorithm                             | Example                                                                |
```
m-mohr: The field name column doesn't make much sense at the moment and could be removed.

vmx: I also thought about removing it. Though I wanted to leave it in for consistency with other extensions and also to make clear that it is always the same field, but the hash algorithm can change (as it is kind of a new concept).

Issue #592, opened 2019-10-10
Search Extension review


tomkralidis:
As discussed on Gitter, we should review the rationale for the Search Extension given OGC API - Features now provides similar capabilities.

matthewhanson:
@tomkralidis 
I see numberMatched and numberReturned here: https://github.com/opengeospatial/ogcapi-features/tree/master/core/openapi/schemas

One issue is that STAC is snake_case and OGC API uses camelCase. It hasn't really come up before because so far everything has been 1 word.  I would prefer `matched` and `returned` or other 1 word alternatives.

m-mohr:
Fully agree with @matthewhanson, but I doubt we can change it and need to live with casing inconsistencies as WFS has been released already.

Is there anything to do for STAC 0.8? Like should we add numberMatched and numberReturned in responses to be fully compliant with OAF?

tomkralidis:
If STAC (API) is seen as a WFS extension, then do we assume that `numberMatched` and `numberReturned` are already part of same?

m-mohr:
I just checked it and STAC API inherits numberMatched and numberReturned from WFS just fine. So currently, things are duplicated in STAC API and we can think about the next steps during the sprint.

PR #631, opened 2019-11-05
removing next from search metadata and extending links to support POST

hgs-truthe01:
**Related Issue(s):** #592 

**Proposed Changes:**

1. Align with WFS by using next link for paging
2. Add additional link properties to support linking for POST searches

**PR Checklist:**


api-spec/openapi/STAC.yaml
```diff
@@ -328,6 +319,21 @@ components:
           example: 'http://www.geoserver.example/stac/naip/child/catalog.json'
         rel:
           example: child
+        method:
+          type: string
+          enum:
+            - GET
```
philvarner: `get` or `GET`?

m-mohr: Following RFC7231 we should have it uppercase.
The RFC states in section 4.1: 
> By convention, standardized methods are defined in all-uppercase US-ASCII letters.

api-spec/openapi/STAC.yaml
```diff
@@ -328,6 +319,21 @@ components:
...
+            - POST
+          default: get
+          description: Specifies the HTTP method that the link expects
+        headers:
+          type: object
+          description: Object key values pairs they map to headers
+          example:
+            Accept:  application/json
```
philvarner: should this be an object structure where the attribute name is the header name, or one like:
"headers" : [ { "name": "Accept", "value": "application/json", 
"name": "Authorization", "value": "Bearer mySpecialBearerToken"
]

api-spec/openapi/STAC.yaml
```diff
@@ -328,6 +319,21 @@ components:
...
+        body:
+          type: string
+          description: For POST requests, the link can specify the HTTP body as a JSON object.
```
philvarner: How well does this handle www-formencoded and form/multipart?

api-spec/openapi/STAC.yaml
m-mohr: Can there be multiple headers with the same name?

scisco:
@mbucknell-usgs

api-spec/openapi/STAC.yaml
m-mohr: But the default value should be uppercase, too.

api-spec/openapi/STAC.yaml
hgs-truthe01: I was looking at other tools and it seems the common way of specifying them is with an object rather than an array:
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Headers
https://github.com/request/request#custom-http-headers

I thought it would be convenient to just pass it is rather than restructure it.

matthewhanson:
This PR needs to be against dev rather than master

api-spec/openapi/STAC.yaml
hgs-truthe01: I think formencoded should be fine, but multipart might be a little difficult. 

I don't think multipart has an obvious JSON representation, but perhaps we can lean on existing tools? It seems many client side libraries make use of https://github.com/form-data/form-data

Of particular interest is the request library that provides supports formData as an object. 

I don't think it would solve the file streaming issue, but it would allow for basic multipart support. Also, I would recommend that any file transfers should be referenced for additional page requests rather than resubmitted. It would be in the servers control to make that decision with this approach.

We might want to consider making body an object instead with some rules for how different content types are supported? application/json  would just be passed through for example.

api-spec/openapi/STAC.yaml
hgs-truthe01: Good catch, fixed.

hgs-truthe01:
I branched from dev, guess I wasn't paying enough attention when I created the PR. Fixed!

joshfix:
I think it would be good to include some examples, specifically of what the POST response looks like.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+- Removed "next" from the search metadata and query parameter, added POST body and headers to the links for paging support
```
scisco: Could you move this under the `### Changed` header

api-spec/STAC-extensions.yaml
```diff
@@ -548,24 +547,14 @@ components:
...
-      style: form
     ids:
       name: ids
       in: query
       description: >
         Array of Item ids to return. All other filter parameters that further
         restrict the number of 
 
-        search results (except `next` and `limit`) are ignored
+        search results are ignored
```
scisco: I think `limit` is still ignored, right?

api-spec/openapi/STAC.yaml
scisco: I thought we decided that the body would be "string"

api-spec/STAC-extensions.yaml
hgs-truthe01: Do we support paging when using ids? It seems a little strange as the user is requesting the exact items they want.

api-spec/openapi/STAC.yaml
hgs-truthe01: Sorry, that was a change when discussing with Phil. He prefers that the key value pairs be expressed as JSON and the client would know how to apply them to the body. I'm happy to facilitate more discussion, but I'm not sure I feel strongly either way. Let's get a discussion going today.

api-spec/openapi/STAC.yaml
tschaub: Agreed an array is more correct for headers.  Many libs accept an object for convenience.  But I don't think it is too onerous if we require that people provide an array.

PR #632, opened 2019-11-05
Renamed STAC endpoints

m-mohr:
**Related Issue(s):** https://github.com/opengeospatial/ogcapi-features/issues/154

**Proposed Changes:**

- The STAC API endpoint `/stac` has been merged with `/`
- The STAC API endpoint `/stac/search` is now called `/items`

**PR Checklist:**

### DIFF
--- a/extensions/eo/examples/example-landsat8.json
+++ b/extensions/eo/examples/example-landsat8.json
@@ -249,7 +249,7 @@
     "links": [
         {
             "rel": "self",
-            "href": "https://odu9mlf7d6.execute-api.us-east-1.amazonaws.com/stage/stac/search?id=LC08_L1TP_107018_20181001_20181001_01_RT"
+            "href": "https://odu9mlf7d6.execute-api.us-east-1.amazonaws.com/stage/items?id=LC08_L1TP_107018_20181001_20181001_01_RT"
         },
         {
             "rel": "parent",
--- a/extensions/eo/examples/example-landsat8.json
+++ b/extensions/eo/examples/example-landsat8.json
@@ -249,7 +249,7 @@
     "links": [
         {
             "rel": "self",
-            "href": "https://odu9mlf7d6.execute-api.us-east-1.amazonaws.com/stage/items?id=LC08_L1TP_107018_20181001_20181001_01_RT"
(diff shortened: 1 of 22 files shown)


philvarner:
I feel it needs to be named something other than `/items`, as to avoid semantic confusion with `/collections/:id/items`, particularly in the context on the Transactions extension were `POST` will create new Items rather than being another way to do a perform a query. 

I'm in favor of it just being `/search`

hgs-msmith:
I rather like /items, because it makes it clear that search is really just accessing the /items collection with a filter.  By making search just filtered access at the items level, we open the door to allowing the same semantics at every level in the hierarchy, effectively achieving client-controlled search scope. It also appeases the REST purists who would object to verbs in the path.

However, Phil's point is also true that we now have a semantic collision with POST on the transaction extension. But it seems to me that since search is a much more common activity than transactions, the clean semantics should go search, and we should consider the adjustment on the transaction extension.

hgs-truthe01:
There was a pretty big discussion when search was originally selected. I think the feeling was that it isn't RESTful with POST not creating an item, but we could interpret POST to /search as "creating" a short lived search record. I think it would cause confusion to change it to /items.

matthewhanson:
The `/items` came about from chats with OAF peeps who have been discussing adding it for cross collection searches, and it fits in with the current nomenclature.

m-mohr:
We discussed in a bigger group, also with OAF guys, and we agreed on /search. Peter Vretanos will also bring that back to OAF. I'll change the PR accordingly.

extensions/eo/examples/example-landsat8.json
```diff
@@ -249,7 +249,7 @@
     "links": [
         {
             "rel": "self",
-            "href": "https://odu9mlf7d6.execute-api.us-east-1.amazonaws.com/stage/stac/search?id=LC08_L1TP_107018_20181001_20181001_01_RT"
+            "href": "https://odu9mlf7d6.execute-api.us-east-1.amazonaws.com/stage/items?id=LC08_L1TP_107018_20181001_20181001_01_RT"
```
joshfix: Should `/items` here be `/search`?

m-mohr: Thanks, fixed.

PR #633, opened 2019-11-06
Search context

matthewhanson:
**Related Issue(s):** #
#592 

**Proposed Changes:**

1. Rename `search` API extension to `context`
2. Change `search:metadata` field to `context`

**PR Checklist:**


api-spec/STAC-extensions.yaml
```diff
@@ -1154,6 +1143,23 @@ components:
           example: NAIP Child Catalog
         length:
           type: integer
+        method:
+          type: string
+          enum:
+            - GET
+            - POST
+          default: get
```
m-mohr: The default value should be upper-case.

PR #634, opened 2019-11-06
Bulk CRUD

gsf-sellis:
**Proposed Changes:**

1. Rename Transaction Extension to CRUD Extension since bulk operations are not transactional. They are equivalent to WFS batch operations, i.e. individual operations succeed or fail independently.
2. ETags to support optimistic locking for PUT, PATCH, and DELETE are always optional.
3. Add spec for bulk POST and DELETE within a collection.
4. Consistently use the term `item` when appropriate instead of `feature`.

**PR Checklist:**

- [x] API only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-spec/blob/dev/api-spec/api-spec.md)


m-mohr:
Is this aligned with the simple Transaction extension the OAF guys have drafted?
See https://github.com/opengeospatial/ogcapi-features/tree/master/extensions/transactions

joshfix:
Following up with what @m-mohr said, it would be good if changes such etags could be added to the OAF spec and the STAC transaction extension would simply point to the OAF link Matthias posted.

gsf-sellis:
@m-mohr The only place where this PR is not in alignment with the OAF "Simple Transaction" proposal is here:

`PATCH /collections/{collectionId}/items/{featureId}`

In the STAC Transaction Extension from 2 sprints ago, it was decided to use the [JSON Merge spec](https://tools.ietf.org/html/rfc7386), and the current OAF proposal is to use the [JSON Patch spec](https://tools.ietf.org/html/rfc6902) for that operation.

Here is a good discussion of the pros and cons of both specs: https://erosb.github.io/post/json-patch-vs-merge-patch/

The major trade-off for the simplicity of JSON Merge is its inability to manipulate the contents of an array. Arrays can only be replaced with an new value.

Since the idea is to keep simple transactions simple, perhaps the OAF proposal could be changed to use the simpler JSON Merge spec for "Simple Transactions," and the more complex and powerful JSON Patch spec for "Complex Transactions?" E.g. a JSON Patch [can be easily reversed for a rollback operation](https://news.ycombinator.com/item?id=11715421), which is not true of a JSON Merge.

This PR does not conflict at all with the OAF proposal's `/transactions` endpoint for "Complex Transactions." It only adds bulk POST and DELETE on the `/collections/{collectionId}/items` endpoint.

@joshfix The comments about ETags to support optimistic locking for safe concurrent PUT, PATCH, and DELETE could certainly be moved up into the OAF spec for "Simple Transactions" as a helpful detail for implementors.

m-mohr:
So then I think we should talk through this during this sprint and agree on something so that we can be fully aligned for simple transactions and just replace our extension with the OAF extension whenever it has been published.

gsf-sellis:
@m-mohr Ok I opened a new issue for discussion here: https://github.com/opengeospatial/ogcapi-features/issues/290

cholmes:
Closing this, as it's for the API which is now in https://github.com/radiantearth/stac-api-spec and I don't think you can just transfer PR's. Tracking issue in that repo at https://github.com/radiantearth/stac-api-spec/issues/14

PR #636, opened 2019-11-06
update maturity levels for extensions

philvarner:
**Related Issue(s):** #

N/A

**Proposed Changes:**

1. Update maturity levels of Extensions based on group decisions at Sprint 5

**PR Checklist:**


extensions/README.md
```diff
@@ -45,18 +45,18 @@ stable for over a year and are used in twenty or more implementations.
...
-| [Label](label/README.md) (`label`)                           | Item             | Items that relate labeled AOIs with source imagery | *Proposal* |
-| [Point Cloud](pointcloud/README.md) (`pc`)                   | Item             | Provides a way to describe point cloud datasets. The point clouds can come from either active or passive sensors, and data is frequently acquired using tools such as LiDAR or coincidence-matched imagery. | *Proposal* |
-| [SAR](sar/README.md) (`sar`)                                 | Item             | Covers synthetic-aperture radar data that represents a snapshot of the earth for a single date and time. | *Proposal* |
-| [Scientific](scientific/README.md) (`sci`)                   | Item +Collection | Scientific metadata is considered to be data that indicate from which publication data originates and how the data itself should be cited or referenced. | *Proposal* |
-| [Single File STAC](single-file-stac/README.md) (-)           | ItemCollection | An extension to provide a set of Collections and Items as a single file catalog. | *Proposal* |
```
m-mohr: Are there 3+ implementtions of it? I'm wondering the same for Asset Definition. Also not sure for SAR (I know of two).

joshfix: It's an interesting point, given that we currently define that in order to reach candidate maturity, there must be 3 implementations.

matthewhanson:
We talked about this and before merging we'd like to get people to actually add implementations to each of the extensions pages before we can state for sure that they have reached some level of maturity based on implementations.

notthatbreezy:
I have a PR to update the implementations for the label extension: https://github.com/radiantearth/stac-spec/pull/648

extensions/README.md
m-mohr: Another thought: Are three clients/general-purpose servers implementing the spec enough without an actual catalog? Should we say something like there must be at least 3 catalogs from distinct vendors and there should be a client implementing it?

matthewhanson:
We discussed this again...the problem is that there are some large changes to the extensions for 0.9.0, so many of these effectively go back to "Proposal".   We also haven't included implementations on the extensions page.

So I'm going to close this for now. The next step, after all these PRs are merged, is to get some implementations using the new extensions up, get those listed under Implementations for each extension, and then we update the maturity level.

PR #637, opened 2019-11-07
Asset rel field for purpose

philvarner:
**Related Issue(s):**  n/a

**Proposed Changes:**

1. add a field `rel` to Asset to be used for the purpose of the asset, similarly to `rel` in Link

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Changed
+- Added attribute `purpose` to Asset, to be used similarly to Link `rel`
+
+### Fixed
+
+n/a (yet)
+
 ## [v0.8.1] - 2019-11-01
 
 ### Changed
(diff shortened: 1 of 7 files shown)


jjrom:
Why not call it rel if it's similar to rel in link ?

philvarner:
> Why not call it rel if it's similar to rel in link ?

I think that's fine -- I thought there was some desire to have it named something that didn't mean 'relation'.

m-mohr:
We discussed this in a small group and think also purpose is not good, but rel is good neither. With rel we always refer to the IANA relation type list in the docs and that list lists nothing that is really useful for assets. So it's not really rel and I wouldn't explain the field with as being similar to rel. Maybe it's better to name it "role" and have a better description what it should be use for.

philvarner:
`role` sounds good to me. Could be confused with permissions.

m-mohr:
Indeed, although you'd usually not see it standalone. Either it would be for example `role=thumbnail` and then the values explains that it's not about permissions. Or you'd see it in docs and there's an explanation coming with it. Anyway, it's bike-shedding and doesn't make me not approve this PR. 

Oh, I just realized: PR is against master, not dev.

philvarner:
https://github.com/radiantearth/stac-spec/pull/651

Issue #149, opened 2018-08-14
Drone profile


mojodna:
(This is mostly a placeholder, but I'd like to capture metadata items that are common to drone imagery.)

Imagery captured using drones has some different characteristics than satellite imagery, in part because it is typically produced by stitching together multiple images taken at varying headings, angles, exposures, and times.

* `startDateTime` - time at which the first item contributing to the image was captured
* `endDateTime` - time at which the last item contributing to the image was captured
* ...

matthewhanson:
@mojodna as you and I talked about, start_datetime and end_datetime (no mixing camelcase in!) applies to a lot of data, even the EO profile and maybe deserves being made either optional core fields or it's own extension.  Will start new ticket

crypticsymbols:
Does STAC contain an elegant structure to reference several source images that may have been used to generate one single mosaic?

m-mohr:
Hi @threewordphrase, we are still thinking about mosaics in STAC, see https://github.com/radiantearth/stac-spec/issues/150

Currently, you can add several derived_from links to the Item, which link to the source imagery.

m-mohr:
Some related extensions have been proposed:
- https://github.com/linz/stac/tree/master/extensions/aerial-photo
- https://github.com/linz/stac/tree/master/extensions/camera

PowerChell:
Existing extensions cover this well, closing.

PR #639, opened 2019-11-07
[WIP] Aerial extension

geospatial-jeff:
**Related Issue(s):**
- #149 

**Proposed Changes:**

1. Adds the `aerial` extension for describing imagery coming from airborne sensors.  This is a WIP and only contains a couple fields with the goal of beginning a conversation about what fields should be included in the extension.

**PR Checklist:**


geospatial-jeff:
Some initial ideas on different fields that the `aerial` extension should support:
1. Fields specific to historical aerial imagery.  Some examples from historical imagery program are:
    - National High Altitude Photography (NHAP) uses metadata fields like `roll`, `frame`, and `film_type` (see [here](https://lta.cr.usgs.gov/DD/nhap.html))
    - USGS Aerial Photo Mosaics uses metadata fields like `flying_height`, `image_quality`, `image_type`, `recording_technique`, and `stereo_overlap`. (see [here](https://lta.cr.usgs.gov/DD/photo_mosaic.html#image_type))

2. Support EXIF data which is primarily used for aerial data (specifically drone).  This would allow reading EXIF data without having to first download the file.

matthewhanson:
@geospatial-jeff We have added an "instrument" extension - #658 , which has these same fields and would be used for both sat and aerial data, so for nor I'm going to close this PR.

I think there's still a need for an aerial extension with some of the fields you mentioned above, but since we aren't sure for now what those would be (and no one is currently working on creating a catalog of aerial imagery), then we can defer that to a future PR.

Issue #555, opened 2019-08-15
Sensor extension to combine some properties from EO and SAR extensions


matthewhanson:
Related to some of the discussion that happened at the beginning of #357, I was looking through the EO and SAR extensions for some commonalities. 

Pulling out these fields: 
- constellation
- platform
- instrument

into another extension serves not only the EO and SAR extensions but also future ones such as drones, or even ground based instruments.

- Constellation: a group of individual sensors often used together
- Platform: the unique ID of a single instrument
- Instrument: The name of the instrument

Eventually we might envision other fields such as sensor type (msi, sar, hyperspectral).

I'd also propose that this extension not use a prefix.

m-mohr:
Isn't this a duplicate of #356? Not sure whether we get this done for 0.8. There were some discussions on the STAC sprint 3, but we were stuck with some issues. Unfortunately, I can't recall what issues we faced. Do you remember, @jbants ?

jbants:
If I recall, we were starting to get into the issue of extension inheritance. If a user implements the `SAR` or `EO` extension are they required to implement the following non-prefix fields?

* constellation
* platform
* instrument

matthewhanson:
Hmm, yes this is somewhat of a duplicate, although I was moving away from a `sat` extension, because these 3 fields basically can apply to any sensor at all.

Not sure this needs to be for 0.8.0, but pulling stuff out of extensions and renaming things I'd rather we do sooner than later.

As far as extension inheritance....I'm not sure, should it be required?  I think so.   Someone could still have data where they didn't want to include details about the platform or instrument (perhaps it's sensitive info), but they could still just not populate those fields.

matthewhanson:
Also discussed in #356  is bands. The concept of bands is in both SAR and EO, but they contain some different fields.

But bands in general are really just variables, and any sensor is potentially capable of collecting data over multiple variables, so I could see it included more generally here as well, and with extensions adding fields to it.

cholmes:
I'm going to move this to 0.9.0, as I think we need to have a larger discussion on this and get a plan and philosophy of how to approach this. Overall I like the move, towards some 'core extensions' that don't need prefixes, but aren't in the absolute core. But we need to decide if it's one larger extension that lists all the non-prefix stuff, or a set of extensions with some special designation. I lean a bit towards the former to find them all in one place. 

I hear you on changing things earlier, but we do have decent versioning now so I don't think it should mess up too many people, and I think we should push out 0.8.0 asap.

[...]

Issue #216, opened 2018-08-29
eo:platform and eo:instrument may have multiple values


simonff:
eo:platform and eo:instrument are not necessarily single-value fields. Eg, for MODIS MCD* datasets platform is MODIS Terra + MODIS Aqua, and for combined Sentinel-1 or -2 data the platform is both S1A/S1B etc. As for the instrument, Landsat 8 has both OLI and TIRS instruments, and the main L8 product has bands from both.

m-mohr:
We have a similar problems with the licenses and tried to solve it by using SPDX license expressions which allows something like `Apache-2.0 OR CC-BY-4.0`. I'm not suggesting that we use exactly this, but I think both fields mentioned by you should somehow stick to be strings as they are used inside the Items properties as well, which are not allowed to hold arrays or objects. Therefore we may just add a recommendation how to generally separate fields with multiple values, e.g. `|` or `/` or `;`.

cholmes:
I like @m-mohr's thinking on this. Having a single value field to search on makes things a lot easier / gives us a simpler spec. 

Also I think 'constellation' in #169 was also supposed to somehow help S1A vs S1B, though I still need to see more concrete examples to really understand it.

m-mohr:
By the way, the eo extension already has a paragraph on this: 
> eo:instrument is the name of the sensor used, although for Items which contain data from multiple sensors this could also name multiple sensors. For example, data from the Landsat-8 platform is collected with the OLI sensor as well as the TIRS sensor, but the data is distributed together and commonly referred to as OLI_TIRS.

This could be improved, for sure, e.g. by describing how multiple values should be separated.

m-mohr:
For collections and catalogs it is solved with the summaries as proposed in #413. For Items, I think need further evaluation.

m-mohr:
So the summaries now allow this for collections. The question here is whether there are really many Items that have multiple values for platform and instrument? Usually Sentinel-2A and Sentinel-2B images would be a separate items. And for "merged" imagery/mosaics/... I'm not sure whether it's worth the effort to change anything here. It could just be a string such as "Instrument1 + Instrument2" or so. Anyway, I think we should come to a decision before 1.0.

PR #644, opened 2019-11-07
sat extension

matthewhanson:
**Related Issue(s):** #

#216 
#356 
#555 

**Proposed Changes:**

Consolidates satellite related fields from sar and eo extensions into a single sat extension

1. New sat extension proposal
2. remove fields from eo and sar that are in new proposal

**PR Checklist:**


extensions/sat/README.md
```diff
@@ -0,0 +1,91 @@
...
+| orbit_state        | string  | The state of the orbit. Either `ascending` or `descending` for polar orbiting satellites, or `geostationary` for geosynchronous satellites |
+| relative_orbit        | integer       | The relative orbit number at the time of acquisition. |
+
+**platform** is the unique name of the specific platform the instrument is attached to. For satellites this would 
+be the name of the satellite, whereas for drones this would be a unique name for the drone. Examples include 
+`landsat-8` (Landsat-8), `sentinel-2a` and `sentinel-2b` (Sentinel-2), `terra` and `aqua` (part of NASA EOS, 
+carrying the MODIS instruments), `mycorp-uav-034` (hypothetical drone name), and `worldview02` 
+(Maxar/DigitalGlobe WorldView-2).
+
+**instruments** is an array of all the sensors used in the creation of the data. For example, data from the Landsat-8 platform is collected with the OLI sensor as well as the TIRS sensor, but the data is distributed together so would be specified as ['oli', 'TItirsRS']. Other instrument examples include `msi` (Sentinel-2), `aster` (Terra), and `modis` (Terra and Aqua).
```
m-mohr: TItirsRS?

m-mohr:
I have concerns about several of the changes here.

First of all, the SAR extension was written in a consensus process with I think about 6 to 8 parties involved. Changing some of the decisions should also go through a broader review process again. For example, pass_direction seems to be a commonly used term in SAR, but orbit state is something new I have rarely heard before. absolute_orbit was requested from SAR folks (see related issues/PRs) and what is the reason to remove it again?

I'm also not sure whether we are moving too many fields to sat. The sun angles are not so relevant for SAR and could remain in EO, I think.

Also, we were trying to remove the replication of instrument(s), platform etc. But now it would be in sat and aerial again so it is duplicated again.

Don't get me wrong: I think aligning SAR and EO makes sense and I appreciate the effort going into this, but I'm afraid of changing too many things so close to 1.0. This is bringing back many things that were above proposal state back into the proposal phase just before 0.9 and a close 1.0. So we should be careful here as this will also break many implementations and catalogs.

matthewhanson:
orbit state is the same as pass direction, and it is used in both EO and SAR.  Orbit state is just a more general term as it works for geostationary orbits as well, not just polar. I'm fine with keeping it as 'pass_direction', but it should still have the 3 states: ascending, descending, geostationary

I've been using the SAR extension quite a bit and I think that a lot of the fields that were initially included are not fields important for search at all, and seem to be more "let's put all SAR related fields in here".  I reached out to SAR folks and no one could think of any reason at all for needing absolute orbit.

The `sat` extension is intended to conceptually include fields related to positioning and geometry.  It would seem a shame to keep some of those fields in EO when this could be a pretty tightly defined extension.

I would recommend also dropping the prefix in Aerial, just hadn't gotten to commenting on that PR yet.

I get your concerns overall, but my main worry for a while now is that we would be stuck with bloated extensions that seemed to be more of a catch-all then a clearly defined extension, like EO currently. I've been wanting to make these changes but they've always been pushed off to the next version.   I'm less concerned overall with breaking existing catalogs, as we are still pre 1.0 and I think that's always been a risk. I think it's more important to get this right than worry about backwards compatibility.

m-mohr:
> orbit state is the same as pass direction, and it is used in both EO and SAR. Orbit state is just a more general term as it works for geostationary orbits as well, not just polar. I'm fine with keeping it as 'pass_direction', but it should still have the 3 states: ascending, descending, geostationary

I'm fine with both if all communities are used to these terms. We should ask members of the communities about that, I guess?

> I've been using the SAR extension quite a bit and I think that a lot of the fields that were initially included are not fields important for search at all, and seem to be more "let's put all SAR related fields in here". I reached out to SAR folks and no one could think of any reason at all for needing absolute orbit.

absolute_orbit is arguably one of the fields that could probably be removed as it's really not really well suited for search. What else do you think are non-searchable fields?

> The `sat` extension is intended to conceptually include fields related to positioning and geometry. It would seem a shame to keep some of those fields in EO when this could be a pretty tightly defined extension.

Fair enough. This should be explained in the extension description.

On the other hand, it feels strange that there are multiple instrument related fields in SAR, but that actual instrument is not defined there and SAR doesn't require sat to be used.

> I would recommend also dropping the prefix in Aerial, just hadn't gotten to commenting on that PR yet.

That would lead to having multiple potential definitions per field name, which was (usually) not the case before. So to know which field exactly it is, I would have to check stac_extensions. That sounds too complex.

> I get your concerns overall, but my main worry for a while now is that we would be stuck with bloated extensions that seemed to be more of a catch-all then a clearly defined extension, like EO currently.

Fair point, but IMHO has nothing to do with moving fields to a sat extension. Instead we could just clean the extensions.

> I think it's more important to get this right than worry about backwards compatibility.

The question is whether we get this right with this approach. I'm not yet convinced at all, especially as we duplicate fields again in aerial and sat and some fields in sat don't apply to all use cases. At the moment it feels like the previous approach was cleaner as there was just one extension per use case which you had to look at. On the other hand we better separate sat from aerial...

m-mohr:
This should now have a sat prefix, right?

matthewhanson:
Should it?  
Are we doing core extensions - or will all extensions move out of this repo except the ones that move into the Item spec?

I would say that if prefixes are not required, then I wouldn't use one here, the field names are descriptive enough on their own.

extensions/sat/json-schema/schema.json
```diff
@@ -0,0 +1,75 @@
...
+  "definitions": {
+    "eo": {
+      "type": "object",
+      "required": [
+        "properties"
+      ],
+      "properties": {
+        "properties": {
+          "type": "object",
+          "required": [
```
matthewhanson: @m-mohr is this sufficient for requiring the fields from the core Item?

matthewhanson:
Updated to use a `sar` prefix, for now. We can discuss if it makes sense to remove the prefix for this extension

extensions/sat/json-schema/schema.json
m-mohr: I think that should work, let'y try it out tomorrow in a validator.
Are these requirements specified in the written spec?

extensions/sar/examples/envisat.json
```diff
@@ -21,8 +23,12 @@
...
   "properties": {
     "datetime": "2009-05-20T02:40:46.763325Z",
-    "dtr:start_datetime": "2009-05-20T02:40:01.042784Z",
-    "dtr:end_datetime": "2009-05-20T02:41:31.483865Z",
+    "start_datetime": "2009-05-20T02:40:01.042784Z",
+    "end_datetime": "2009-05-20T02:41:31.483865Z",
+    "platform": "envisat",
+    "constellation": "envisat",
+    "instruments": ["asar"],
+    "pass_direction": "descending",
```
m-mohr: Prefix missing.

extensions/sar/examples/sentinel1.json
```diff
@@ -28,6 +29,8 @@
     "platform": "sentinel-1a",
     "constellation": "sentinel-1",
     "instruments": ["c-sar"],
+    "pass_direction": "ascending",
+    "relative_orbit": 33,
```
m-mohr: Prefix missing

extensions/sat/json-schema/schema.json
```diff
@@ -0,0 +1,76 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "schema.json#",
+  "title": "Sat Extension",
+  "description": "STAC Sat Extension to a STAC Item.",
+  "allOf": [
+    {
+      "$ref": "../../../item-spec/json-schema/item.json#/definitions/core"
```
m-mohr: You can remove `#/definitions/core`

extensions/sat/json-schema/schema.json
```diff
@@ -0,0 +1,76 @@
...
+    },
+    {
+      "$ref": "#/definitions/sat"
+    },
+    {
+      "$ref": "../instrument/json-schema/schema.json"
+    }
+  ],
+  "definitions": {
+    "eo": {
```
m-mohr: must be sat

PR #645, opened 2019-11-07
fixing examples to include method

hgs-truthe01:
**Proposed Changes:**

1. Fixed issue with pagination examples
2. Added GET based example as well

**PR Checklist:**

### DIFF
--- a/api-spec/examples.md
+++ b/api-spec/examples.md
@@ -73,8 +73,27 @@ OGC API supports paging through hypermedia links. STAC follows the same pattern 
 
 In addition, to avoid passing large post body values to and from the server a "merge" property can be specified. This indicates that the client should send the same post body that it sent on the previous request, but with the specified body value merged on top. This allows servers to indicate what needs to change to get to the next page without mirroring the entire query structure back to the client.
 
-Request:
+Simple GET based search just uses the standard next link:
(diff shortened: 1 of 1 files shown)


api-spec/examples.md
```diff
@@ -103,7 +123,7 @@ Response:
 ```
 
 This tells the client to post to the search endpoint with a body of:
-```
+```json
```
m-mohr: This is not valid JSON due to the POST ... header

api-spec/examples.md
```diff
@@ -114,18 +134,19 @@ POST /search
...
+    "method": "POST",
     "body": {
         "next": "token"
     }
 }
 ```
 The above link tells the client not to merge (default of false) so it is only required to pass the token.
 
-```
+```json
```
m-mohr: This is not valid JSON due to the POST ... header

api-spec/examples.md
hgs-truthe01: Yeah, but I'm pretty sure it's just a style hint so it shouldn't hurt anything and makes it look prettier right?

m-mohr: Seems like that. Sometimes it makes things worse, but here it doesnt.

PR #649, opened 2019-11-08
Simplified format for fields parameter when using GET

philvarner:
**Related Issue(s):** 

N/A

**Proposed Changes:**

1. Add a simplified format for API Fields Extension when using a GET query parameter
2. Make language more explicit that the `fields` value is a suggestion/hint to the server, and that the semantics of it are entirely up to server (with recommendations)

**PR Checklist:**

- [ ] This PR has **no** breaking changes. - this PR changes the format of a parameter


m-mohr:
PR should be against dev, not,master.

PR #650, opened 2019-11-08
Add simplified GET fields syntax

philvarner:
**Related Issue(s):** #
n/a

**Proposed Changes:**

* Add a simplified format for API Fields Extension when using a GET query parameter
* Make language more explicit that the fields value is a suggestion/hint to the server, and that the semantics of it are entirely up to server (with recommendations)

**PR Checklist:**


api-spec/extensions/fields/README.md
```diff
@@ -2,32 +2,68 @@
...
+polygons can be very large when reprojected to EPSG:4326, as in the case of a highly-decimated sinusoidal polygons.
+Implementations are also not required to implement semantics for nested values whereby one can include an object, but
+exclude attributes of that object, e.g., include `properties` but exclude `properties.datetime`.
+
+No error should be returned if a specified field has no value for it in the catalog.  For example, if the attribute 
+"properties.eo:cloud_cover" is specified but there is no cloud cover value for an Item or the API does not even 
+support the EO Extension, a successful HTTP response should be returned and the Item entities will not contain that 
+attribute. 
+
+If no `fields` are specified, the response is **must** be a valid ItemCollection. If a client excludes
```
m-mohr: Maybe add a link to the ItemCollection spec. It's not so well known (yet).

api-spec/extensions/fields/README.md
```diff
@@ -2,32 +2,68 @@
...
+attributes that are required in a STAC Item, the server may return an invalid STAC Item. For example, if `type` 
+and `geometry` are excluded, the entity will not even be a valid GeoJSON Feature, or if `bbox` is excluded then the entity
```
m-mohr: Just a (stupid?) thought, but maybe it would be better to set some(?) fields to null instead of removing them. A null geometry would be valid GeoJSON, but removing it would be invalid GeoJSON. Depends on how important it is for us to return valid GeoJSON in most cases, but still allow removing large fields... Same applies for STAC Item compliance. An empty object would be compliant with the spec, but removing would not be compliant. Similar things apply for (most?) objects/arrays, I guess.

matthewhanson: It looks like while technically it is a GeoJSON with a null geometry, most clients won't be able to handle it and generate an error.
GeoJSON must also contain `properties`, even if empty, and also must contain the `type` field, so a minimal FeatureCollection would look like:
```
{
   "type": "FeatureCollection"
   "features": [
      {
         "type": "Feature",
         "geometry": null,
         "properties": {}
      }
   ]
}
```

It's really not a bad idea to keep our response as valid GeoJSON, otherwise using the fields query will sometimes be valid and sometimes not.

m-mohr:
Similarly to #513, I'm wondering whether it would actually be a good idea to have something that can properly be defined in OpenAPI and translates better between GET and POST. We can't properly describe the current GET behavior except for textual descriptions. Having two completely different approaches for GET and POST makes things more complicated to implement. Also, the documentation speaks about `exclude` and `include`, but you cant find these fields in GET.

Couldn't we simply translate POST and GET as defined by OpenAPIs `form` parameter style?

GET: `include=properties.datetime,id,bbox&exclude=geometry`

POST:
```
{
  "include": ["properties.datetime", "id", "bbox"],
  "exclude": ["geometry"]
}
```

api-spec/extensions/fields/README.md
philvarner: I feel the problem with this is that the result should never contain non-nested attributes with  values that are not the actual values of the Item.

m-mohr: If we restrict to type, geometry and properties, I think it would be fine.

matthewhanson: I also do not have a strong opinion here.  It seemed like it might be a good idea to ensure responses are always valid, but I'm also fine with saying "if you use fields the response may not be valid GeoJSON".

PR #655, opened 2019-11-11
allow multiple roles in asset

philvarner:
**Related Issue(s):** 

n/a

**Proposed Changes:**

1. Allow multiple roles on an Asset

**PR Checklist:**


item-spec/json-schema/item.json
```diff
@@ -209,8 +209,8 @@
           "title": "Asset type",
           "type": "string"
         },
-        "role": {
-          "title": "Asset role",
+        "roles": {
```
m-mohr: The schema is not correct.

Issue #619, opened 2019-10-25
Proposed SAR changes


matthewhanson:
I've been working with SAR recently and some things have come up WRT the `sar` extension.

- `sar:pass_direction` is required, but it is specific to satellites, and a SAR instrument could be mounted on an airborne platform  (which is why we need `sat` extension)
- sar:bands -> data_type: polarizations can be represented as different products (i.e., assets) within the same collection. If I have SLC and GRD products for HH and HV in the same collection I would need 4 different bands rather than 2. It seems that data_type is more a property of an asset, rather than a band (also, a single asset is not going to contain different data_types)

Also, more generally, the extension seems like it includes a lot of fields that just don't fit the purpose of "search and discovery" and I wonder if some could be pruned out.   I've got some questions out to some SAR folks to get input on that.

@m-mohr

m-mohr:
> * `sar:pass_direction` is required, but it is specific to satellites, and a SAR instrument could be mounted on an airborne platform  (which is why we need `sat` extension)

For airborne SAR you can simply set the field to null: "Direction of the orbit, either ascending, descending or **null if not relevant**."

> * sar:bands -> data_type: polarizations can be represented as different products (i.e., assets) within the same collection. If I have SLC and GRD products for HH and HV in the same collection I would need 4 different bands rather than 2. It seems that data_type is more a property of an asset, rather than a band (also, a single asset is not going to contain different data_types)

I thought that's how bands work at the moment for EO as well?! I remember the discussions about asset specific metadata and thought the solution for now is to have more bands?

> Also, more generally, the extension seems like it includes a lot of fields that just don't fit the purpose of "search and discovery" and I wonder if some could be pruned out. I've got some questions out to some SAR folks to get input on that.

Not sure whether it's really *a lot* of fields. Which one do you think about? The only fields that are not so related to search and discovery (from my not so expert perspective) are data_type and unit in the bands. And these are just really there for better understanding from a human perspective. Maybe pixel_spaing_* is one field not so relevant.

matthewhanson:
Yeah, but if it is allowed to be `null` then why would it be required?  Why not just make all fields required, but then set them to `null` if not relevant?  To me the point of something being required is that it is relevant in all cases.  By saying that it can be set to null if not relevant you are saying that it is optional, but you must still provide a value.


There is no data type of other asset specific metadata in EO bands: https://github.com/radiantearth/stac-spec/tree/master/extensions/eo#band-object


[...]

PR #662, opened 2019-11-13
Sar updates

matthewhanson:
**Related Issue(s):** #
#619

**Proposed Changes:**

1. change `sar:type` to `sar:product`
2. move `sar:bands` attributes into assets

**PR Checklist:**


matthewhanson:
@m-mohr This removes `sar:center_wavelength`, moves `sar:type` to `sar:product`

I've also moved `sar:bands` into assets which we can discuss, I did not update the schema yet but wanted to show an example. It drops the description field because it's not specific to SAR, and we might consider adding it to the assets.   The other fields: `data_type`, `unit`, and `polarization` definitely vary by asset, and in fact `band` is actually not a great name for these terms because `data_type` and `unit` arise from processing, not how the data is collected like with eo:bands.

m-mohr:
* I'm not yet convinced that it's really a good idea to remove sar:center_wavelength (or sar:center_frequency) as they are more specific as sar:frequency_band.
* Fine with renaming type to product or product_type. Just thinking, but would that be a field which also applies to EO (with a different set of values)?
* As with the bands for EO, I'm not yet sure I like the bands in the assets. But that's a discussion to be held in a bigger group anyway and until then we shouldn't change it.

matthewhanson:
This only has sar:center_wavelength being removed, it's kinda redundant having both since you can calculate one from the other based on the medium the waves are traveling through.

I do think product is a rather general term that could be useful for other types of data.  For EO it might be Radiance, TOA, Surface Reflectance, etc.

matthewhanson:
The more I think about the bands here the more I think that both `data_type` and `processing` are general asset level metadata fields.  I don't see how they fit along with polarizations at all, they have to do with how the data is processed, not how it is collected.

Once you remove those fields from `sar:bands`, and `description` which is certainly general - then all you have left is the polarization which seems silly to have to reference from the assets.

```
'myasset': {
    "polarizations": ["HH"]
}
```
OR

```
'sar:bands': [
   {
       "name": "HH",
       "polarization": "HH"
   }
]
---
'myasset': {
    "polarizations": [0]
}
```

PR #442, opened 2019-04-29
Making the Commons extension Core / Standalone Collections

m-mohr:
Making the Commons extension Core as optional field in the collections, see issue #392 for a description and discussions. 

Combined this with the issue to add language how to combine (standalone) collections with extensions #344.

PR #665, opened 2019-11-19
Clean up: Commons Extension and remove summaries from Catalogs

m-mohr:
**Related Issue(s):** None?

**Proposed Changes:**

This is based on discussions during the sprint. @cholmes encouraged us to think about a clean up / simplification. Consensus (afaik) was to: 
1. make the Collection properties (Commons extension) a extension again (this is basically a revert of PR #442 with some additional changes to conform the recent version of the spec)
2. remove summaries from catalogs.

**PR Checklist:**


m-mohr:
This PR needs to align with #670 or the other way round, depends on which gets merged first.

collection-spec/collection-spec.md
```diff
@@ -146,107 +145,19 @@ Implementors are free to add other derived statistical values to the object, for
...
+This makes much sense for fields such as `platform` or `eo:epsg`, which are often the same for a whole collection, but doesn't make much sense for `eo:cloud_cover`, which usually varies heavily across a Collection.
+The data provider is free to decide, which fields are reasoable to be used.
 
 ## Extensions
 
+Important related extensions for the STAC Collection Specification:
+
+* [Commons extension](../extensions/commons/README.md), which primarily allows to add shared Item metadata to Collections,
+  but could also be used to describe Collections better that are not referring to Items by adding additional fields from Item extensions.
+  Please note that this extension is only in '[proposal](../extensions/README.md#extension-maturity)' stage.
```
anayeaye: Can I interpret this 'which primarily allows to add shared Item metadata to Collections,
  but could also be used to describe Collections better that are not referring to Items by adding additional fields from Item extensions.'

As this
`which can be used for fields with shared values for all Items in a Collection. Core, extension, and supplemental user-defined Item properties can be defined using this extension.`?

As in: this extension allows users to add wild-card fields at the collection level to describe the items in the collection that do not require validation but support a better understanding of the collection contents?

m-mohr: Sorry, this PR was merged without me having a chance to answer. Will answer tomorrow and potentially issue a new PR, if I think there's room for a better description.

m-mohr: > which primarily allows to add shared Item metadata to Collections,

This part is still valid.

> but could also be used to describe Collections better that are not referring to Items by adding additional fields from Item extensions.

This part is not. It was used for additional fields before we introduced summaries. Therefore I'll remove this part and issue a new PR.

Issue #663, opened 2019-11-13
Add `description` to asset


matthewhanson:
In the SAR extension sar:bands is really used to describe characteristics of different assets due to processing or collection configuration.

In that there was `description`, which seems more general than just SAR for assets in general.  Especially with an asset definition, `description` seems like it may be useful to better describe what an asset is beyond the title.

m-mohr:
Asset descriptions could indeed be useful, although such descriptions could probably also live in the item description.

ucalyptus:
Can I be assigned this? @anayeaye  @matthewhanson

Issue #656, opened 2019-11-11
Combining eo:bands into assets


matthewhanson:
The original thinking for a separate eo:bands field for spectral data was that we could define the bands on the sensor, then reference those from the assets so that we would not have to repeat information when a band is used in more than one asset.   So we have an asset:

```
        "B2": {
            "type": "image/x.geotiff",
            "eo:bands": [
                1
            ],
            "title": "Band 2 (blue)"
        },
```

Then our bands property is:
```
        "eo:bands": [
            {
                "name": "B1",
                "common_name": "coastal",
                "gsd": 30,
                "center_wavelength": 0.44,
                "full_width_half_max": 0.02
            },
            {
                "name": "B2",
                "common_name": "blue",
                "gsd": 30,
                "center_wavelength": 0.48,
                "full_width_half_max": 0.06
            }
     ]
```

However a few things have happened since then. 

1. We now have an asset definition as a way to provide Asset information at the collection level
2. the realization that 'gsd' in eo:bands is really a characteristic of an asset, not a band. We can get around this by adding in another band with a different gsd, but then we are really not reusing bands like originally intended
3. it seems that in practice, it isn't common to have bands referenced in multiple assets

One option that can greatly simplify this is to just put the eo:bands info in the asset, so:
```
    "B2": {
        "title": "Band 2 (blue)",
        "type": "image/x.geotiff",
        "eo:bands": [
            {
                "common_name": "coastal",
                "gsd": 30,
                "center_wavelength": 0.44,
                "full_width_half_max": 0.02
            }
        ]
    },
```

This avoids the entire referencing back to another array to find out what's in the asset.  And the spectral band info is still at the Collection level, but consolidated so easier for humans.

The drawback is that the asset info is duplicated at the Item level the way we have things right now, unless we allow for partially described assets at the Item level (e.g., just href, and requires deep merge with asset definition to get the whole thing).

m-mohr:
Sorry, I don't get it. Why is gsd still at the item level if it is asset specific? How is the relation between asset-level and item-level bands? Why is the band name missing from the asset-level band? Would I be allowed to also use other fields at the asset-level? Maybe a full item example would help...

m-mohr:
Related PR, which had some discussions and was closed: https://github.com/radiantearth/stac-spec/pull/270

matthewhanson:
GSD is still at the Item level because it represents the best possible resolution across all bands for the Item. This change wouldn't affect that at all.

[...]

PR #670, opened 2019-11-19
SAR Extension changes

matthewhanson:
**Related Issue(s):** #
#619 

**Proposed Changes:**

1.  Remove `sar:absolute_orbit` and `sar:center_wavelength` fields from `sar` extension.
2. Remove `data_type` and `unit` from the `sar:bands` object in the `sar` extension.
3. change `sar:type` to `sar:produc_typet`.

**PR Checklist:**


m-mohr:
This PR needs to align with #665 or the other way round, depends on which gets merged first.

m-mohr:
I think I'd prefer product_type. Would a product_type for eo also be useful? 

If we remove the fields from the bands, I'd propose to mention in the description of the band description field that catalog creators should add info about data type and unit in the text.

matthewhanson:
product_type is fine.  
I think that it could be useful for EO, but there's no real set of values for what those would be.  Maybe products for Surface reflectance, TOA, Surface leaving radiance, etc.   I think it's fine leaving it in SAR and just enumerating the values for SAR. 

For data type and unit, I think it's better putting that in a description for the asset (see #663 ).

m-mohr:
Hmm... So in that case what purpose has sar:bands? 🤔

matthewhanson:
sar:bands doesn't serve a purpose other than providing a polarization, which doesn't make sense to add a description to.

For that matter, sar:bands as a separate thing doesn't make sense as it's own object.  We should just add `sar:polarizations` to the Asset in this extension.  Especially since polarization is not a "band", as discussed in https://github.com/radiantearth/stac-spec/issues/619#issuecomment-554146028

matthewhanson:
This last commit removes `sar:bands` as there was only two attributes left: `description`, which is handled by an asset description now, and `polarization`.  

Instead it adds a `sar:polarizations` array to the asset field.

In so doing I noticed that for the JSON schema neither the SAR extension nor the EO Extension included the `sar:bands`/`eo:bands` in the Asset definition.

The proposed JSON schema removes the `sar:bands` property, but does not add an additional attribute to the Asset definition. Once we determine how to do this it should be done for both the SAR and EO extensions.

m-mohr:
I'd like to first discuss the future directions with asset specific metadata first before approving this. Related #656

extensions/sar/README.md
```diff
@@ -56,26 +54,10 @@ properties. The table below shows the common name based on the wavelength and fr
...
-
-## Associating assets with bands
-
-Asset definitions that contain band data should reference the band index. Each asset should provide a `sar:bands` property that is an array of 0 based indexes to the correct [Band Objects](#band-object).
-
 ### Item [`Asset Object`](../../item-spec/item-spec.md#asset-object) fields
 | Field Name | Type     | Description                                  |
 | ---------- | -------- | -------------------------------------------- |
-| sar:bands  | [number] | Lists the band names available in the asset. |
+| sar:polarizations  | [string] | Lists the polarizations available in the asset, in order. One of `HH`, `VV`, `HV`, `VH` |
```
m-mohr: This should be "any of" instead of "one of", right?

CHANGELOG.md
```diff
@@ -20,7 +20,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - Support for [CommonMark 0.29 instead of CommonMark 0.28](https://spec.commonmark.org/0.29/changes.html)
 - Collection field `property` and the merge ability moved to a new extension 'Commons'
+<<<<<<< HEAD
```
m-mohr: Here's a merge issue...

extensions/checksum/examples/sentinel1.json
```diff
@@ -44,9 +44,7 @@
       "href": "./measurement/s1a-ew-grd-hh-20181103t235855-20181103t235955-024430-02ad5d-001.tiff",
       "title": "Measurements",
       "type": "image/tiff",
-      "sar:bands": [
-        0
-      ],
+      "sar:polarizations": ["HH"],
```
m-mohr: Please open an issue on asset-specific metadata before we merge this so that we don't forget to discuss it for 0.9.

extensions/sar/README.md
```diff
@@ -36,6 +34,20 @@ If the data has been collected by a satellite, it is strongly recommended to use
 
 **sar:polarization** specifies a single polarization or a polarization combination. For single polarized radars one of `HH`, `VV`, `HV` or `VH` must be set. Fully polarimetric radars add all four polarizations to the array. Dual polarized radars and alternating polarization add the corresponding polarizations to the array, for instance for `HH+HV` add both `HH` and `HV`.
 
+**sar:product_type**: The product type defines the type of procesed data contained in the assets. A generalized list of suggestions include:
+
+| sar:product_type | Data Type | Description |
+|-------------------|----------------|----------------|
+| SSC | complex | Single-look Slant-range Complex image (standard SLC) |
+| MGD | amplitude | Multilooked Ground-range Detected Image (GRD for Sentinel-1) |
```
m-mohr: So should Sentinel-1 use GRD or MGD? It may need changes in the examples.

extensions/sar/examples/envisat.json
```diff
@@ -38,28 +38,15 @@
...
-    ]
+    "sar:product_type": "GM1_1P"
   },
   "assets": {
     "product": {
       "href": "ASA_GM1_1PNPDE20090520_023957_000001022079_00118_37747_3607.n1.gz",
       "type": "application/gzip",
-      "sar:bands": [0,1]
+      "description": "Amplitude and Intensity. Intensity equals the square of the amplitude band",
+      "sar:polarizations": ["HH", "HH"]
```
m-mohr: Two times HH doesn't make much sense.

extensions/sar/README.md
```diff
@@ -56,32 +68,15 @@ properties. The table below shows the common name based on the wavelength and fr
...
-
-## Associating assets with bands
-
-Asset definitions that contain band data should reference the band index. Each asset should provide a `sar:bands` property that is an array of 0 based indexes to the correct [Band Objects](#band-object).
-
 ### Item [`Asset Object`](../../item-spec/item-spec.md#asset-object) fields
 | Field Name | Type     | Description                                  |
 | ---------- | -------- | -------------------------------------------- |
-| sar:bands  | [number] | Lists the band names available in the asset. |
+| sar:polarizations  | [string] | Lists the polarizations available in the asset, in order. Any of `HH`, `VV`, `HV`, `VH` |
```
m-mohr: Do we validate this in the schema? Could be useful... 
Probably also applies to EO?!

Issue #664, opened 2019-11-15
Unclear what to do for proprietary license with no url


jisantuc:
Per the collection spec, `license` is:

> REQUIRED. Collection's license(s) as a SPDX License identifier or expression. Alternatively, use proprietary if the license is not on the SPDX license list or various if multiple licenses apply. In these two cases links to the license texts SHOULD be added, see the license link relation type.

In the proprietary case, I `SHOULD` add a link to the license. But sometimes I don't have a link. Sometimes I'm just making a catalog and don't want to license it with a permissive license (e.g. I'm making a catalog out of data that someone else owns that I can't license in any particular way), but I also have no idea where a link for a proprietary license would be.

I'm proposing an additional option for `license`, which is `unlicensed`, and then a revision:

> Collection's license(s) as a SPDX License identifier or expression. Alternatively, use proprietary if the license is not on the SPDX license list or various if multiple licenses apply. In these two cases links to the license texts MUST be added, see the license link relation type. Alternatively, use unlicensed if no SPDX License applies to the collection and you cannot provide a link to a license.

This could also include a link to GitHub's [Choosing a License](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/licensing-a-repository#choosing-the-right-license) docs to help people understand the trade-offs of licensing/permissively licensing/not licensing catalogs, though I'm not sure if the same rules apply for software and data catalogs.

m-mohr:
Thanks for the thoughts, but I don't agree with them.

1. "unlicensed" would imply that the owner never chose a license, but what you indeed want to tell is that you don't know the license.
2. IMHO unlicensed data/code/whatever is not useful at all. If you don't know the license, you can't use it without getting in trouble. So the data is not of any use for the user.
3. I don't think a specification is the right place to educate the user how to license. That's maybe something for a best practice, but not for a spec. We also don't teach user how to write a good description etc.

If there's really a case where you don't know the license, I'd say simply use "proprietary" and don't provide a link (it's only "SHOULD" in the spec, not "MUST"). That indicates the user needs to look himself for a license.

Out of curiosity, why would someone make a catalog for data where he doesn't even know whether somebody can actually use it? Is the data you are speaking about a public dataset? Really curious what it is...

jisantuc:
[...]

PR #676, opened 2019-11-20
Specify how to read no URL proprietary license

jisantuc:
**Related Issue(s):** #664

**Proposed Changes:**

1. Explicitly state the interpretation of absence of URL when `license` on collections is proprietary

**PR Checklist:**

- ~API only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-spec/blob/dev/api-spec/README.md#openapi-definitions).~


matthewhanson:
Thanks for the PR @jisantuc , can you update it to be against the `dev` branch rather than `master`?

jisantuc:
It is done -- I went looking for previous chatter about the default branch, is there any particular reason to keep it `master` instead of `dev`, if `dev` is what PRs should target?

m-mohr:
Looks good, with the exception that in the Collection the addition is in the link rel types. In the Item it is in the Field Description and in the Collection it should also be described there instead of in the link rel type, I think.

matthewhanson:
@jisantuc the idea is that the `master` branch is the last released version, so is the branch that people see when they come to the page, while `dev` is the current (under development) version.  PRs should always target `dev`.

m-mohr:
Looks good, thanks. I added a Changelog entry for you.

Issue #692, opened 2019-12-13
Limitations (Inconsistencies) of the operators in the STAC Query Extension 


m-mohr:
I think there are some limitations due to how the operators (e.g. "eg", "lt", "startsWith") are defined:

```yaml
            eq:
              description: Find items with a property that is equal to the specified value. For strings, a case-insensitive comparison must be performed.
            neq:
              description: Find items that *don't* contain the specified value. For strings, a case-insensitive comparison must be performed.
            gt: 
              type: number
              description: Find items with a property value greater than the specified value.
            lt:
              type: number
              description: Find items with a property value less than the specified value.
            gte:
              type: number
              description: Find items with a property value greater than or equal the specified value.
            lte:
              type: number
              description: Find items with a property value greater than or equal the specified value.
            startsWith:
              type: string
              description: Find items with a property that begins with the specified string. A case-insensitive comparison must be performed.
            endsWith:
              type: string
              description: Find items with a property that ends with the specified string. A case-insensitive comparison must be performed.
            contains:
              type: string
              description: Find items with a property that contains with the specified string. A case-insensitive comparison must be performed.
            in:
              type: array
              items:
                type: string
              description: Find items with a property that matches one of the specified strings. A case-insensitive comparison must be performed.
```

Here's the example from the spec with the mentioned inconsistencies:
```yaml
        eo:cloud_cover:
          lt: 50
        providers:
          eq: 'Planet'
        published:
          gt: '2018-02-12T00:00:00Z'
          lte: '2018-03-18T12:31:12Z'
        pl:item_type:
          startsWith: PSScene
        product:
          in:
            - foo
            - bar
            - baz
```

Problems I see at the moment:
* `lt`, `lte`, `gte`, `gt` can not be used with (temporal) strings. That means I can't search for all timestamps newer or older than x. Interestingly, the example (see below) contains this use-case for the `published` field, but officially it's not allowed.
* `in` only works on strings. Why can I not search in a array of integers for example?
* The example uses the `eq` operator for searching in the `providers` array. How would that work?
* There's no way to search in objects (e.g. find provider with name abc or find band with name B4).
* Probably a bit nit-picking, but for `in` it says:
  > Find items with a property that matches one of the specified strings
  
  Is it really "one of" (i.e. exactly one must match)? Or is it really an "any of" (i.e. at least one must match)?

cc @hgs-truthe01 @matthewhanson @joshfix

m-mohr:
I think we should solve this as follows:

[...]

PR #699, opened 2019-12-18
Query polish

philvarner:
**Related Issue(s):** #692 

**Proposed Changes:**

1. Update YAML types to reflect what's being done in practice
2. Update examples to be correct wrt spec

These are technically breaking changes, but I think they more accurately reflect the handful of implementations of this.

**PR Checklist:**

### DIFF
--- a/api-spec/STAC-extensions.yaml
+++ b/api-spec/STAC-extensions.yaml
@@ -1880,11 +1880,12 @@ components:
         $ref: '#/components/schemas/queryProp'
       example:
         'eo:cloud_cover':
+          gt: 8
           lt: 50
-        providers:
-          eq: Planet
-        published:
-          gt: '2018-02-12T00:00:00Z'
+        platform:
+          eq: landsat-8
+        datetime:
+          gte: '2018-02-12T00:00:00Z'
           lte: '2018-03-18T12:31:12Z'
         'pl:item_type':
           startsWith: PSScene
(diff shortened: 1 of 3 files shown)


api-spec/STAC-extensions.yaml
```diff
@@ -1906,50 +1911,74 @@ components:
               description: >-
                 Find items with a property that is equal to the specified value.
                 For strings, a case-insensitive comparison must be performed.
+              oneOf:
```
m-mohr:
```suggestion
              nullable: true
              oneOf:
```

api-spec/STAC-extensions.yaml
```diff
@@ -1906,50 +1911,74 @@ components:
...
+                - type: string
+                - type: number
+                - type: boolean
             neq:
               description: >-
                 Find items that *don't* contain the specified value. For
                 strings, a case-insensitive comparison must be performed.
+              oneOf:
```
m-mohr:
```suggestion
              nullable: true
              oneOf:
```

m-mohr:
Looks good, but eq/neq schould support null, too.

philvarner:
Updated nullable support

m-mohr:
> Updated nullable support

Yes, thanks, but you forgot to run npm run generate-all, right?

philvarner:
@m-mohr I did, update now

PR #701, opened 2019-12-19
Collection: Fix type of interval array, should be string instead of number

philvarner:
**Related Issue(s):** n/a

**Proposed Changes:**

1. Update Collection Temporal Extent interval field definition to be a string rather than a number.  The description and all our examples are RFC3336 strings, but the type was number. Also replaced the confusing term "timestamp" with "datetime" per RFC3336 term. 

**PR Checklist:**


philvarner:
@m-mohr updated the bbox description to be clearer and fixed the examples to be doubly-nested

PR #705, opened 2019-12-23
Release 0.9.0-rc1

matthewhanson:
**Related Issue(s):** #
Update version numbers for 0.9.0-rc1

**Proposed Changes:**

1. update versions

**PR Checklist:**


m-mohr:
I updated some additional version numbers.

anayeaye:
Should the version in api-spec/package-lock.json be updated (I'm not sure that it needs to be in sync)?

m-mohr:
Doesn't necessarily need to be in sync, but could be. We don't publish on npm so the file doesn't have any influence anyway.

PR #718, opened 2020-01-17
Respond with PreconditionFailed when If-Match header invalid

jisantuc:
**Related Issue(s):** #

none

**Proposed Changes:**

1. Respond with as specific a response as possible when we know what went wrong with an item update in api transactions extension PUT. `BadRequest` and `NotFound` don't tell the user that they should refresh the object and try again, while a `PreconditionFailed` for a request with an `If-Match` header is extremely clear

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


cholmes:
Thanks for the contribution! Will be in 0.9.0-RC2, as discussed in the call, since it is a minor improvement and we're going to do the rc2 regardless.

PR #717, opened 2020-01-15
Pagination documentation update

joshfix:
**Related Issue(s):** https://github.com/radiantearth/stac-spec/issues/716

**Proposed Changes:**

1. Documentation updates to clarify how to implement new pagination mechanism for POST requests

**PR Checklist:**

- [x] This PR has **no** breaking changes.
- [x] I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-spec/blob/dev/CHANGELOG.md) **or** a CHANGELOG entry is not required.
- [x] API only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-spec/blob/dev/api-spec/README.md#openapi-definitions).

PR #714, opened 2020-01-09
Added Radiant MLHub to implementations

kbgg:
**Related Issue(s):** #

**Proposed Changes:**

1. Added Radiant MLHub to the list of API implementations

**PR Checklist:**

- [x] This PR has **no** breaking changes.
- [x] I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-spec/blob/dev/CHANGELOG.md) **or** a CHANGELOG entry is not required.
- [x] API only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-spec/blob/dev/api-spec/README.md#openapi-definitions).

PR #712, opened 2020-01-08
Common metadata updates

anayeaye:
**Related Issue(s):** #
N/A

**Proposed Changes:**

1. Corrects item jsonschema to evaluate all common metadata schemas 
2. Item-spec and common-metadata documentation updated with more information about common metadata fields.

**PR Checklist:**

- [x] This PR has **no** breaking changes.
- [x] I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-spec/blob/dev/CHANGELOG.md) **or** ~~a CHANGELOG entry is not required.~~
- [x] API only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-spec/blob/dev/api-spec/README.md#openapi-definitions). *N/A*

Issue #711, opened 2020-01-08
Inconsistency in STAC.yaml and api-spec.md for search endpoint supported methods.


sharkinsspatial:
The [STAC.yaml](https://github.com/radiantearth/stac-spec/blob/dev/api-spec/STAC.yaml#L137) has been updated to reflect `GET` as an optional method for the `/search` endpoint but the [api-spec.md](https://github.com/radiantearth/stac-spec/blame/dev/api-spec/api-spec.md#L18) seems to only show `POST` support?  I could just be misreading it but should this be updated to reflect the new optional `GET` method?

m-mohr:
The whole section is a bit confusing, I think. The section title is "HTTP Request Methods and Content Types" but the text is basically only describing which content types are supported so I'd propose to change the title to "HTTP Content Types". The text just says that you can use some content types with POST /search, but it doesn't say it's not allowed to use GET. It just doesn't specify a content type for GET /search. Also, I think the chapter should be moved down a bit and be above "Filter Parameters and Fields". And then we should probably add a "Method(s)" column for the "... endpoints" sections. That should clarify it.

m-mohr:
Hey @philvarner, I think you wrote the section "HTTP Request Methods and Content Types". I'm trying to understand the "HTTP Request Methods and Content Types" can you explain it? I don't get it. What POST endpoints is spoken about? What does the bullet point about GET tells me? I think we should clarify that section.

philvarner:
I'll take a look at it

philvarner:
This is a change from my previous view on this -- I think we should now do the following:

* GET w /query params is **required** (maybe some discussion around how this works with a geometry?)
* POST w/ JSON is **optional** but highly recommended (e.g., your service is compliant but pretty trash if you don't implement this)
* POST w/ application/x-www-form-urlencoded is **optional** but recommended. For most web service frameworks, this is either supported by default when GET w/ query params is implemented or it only requires some additional config or a bit of code.
* POST w/ multipart/form-data is not part of this spec. if someone wants to implement that they can, but it is **not** part of the STAC API spec. (for my implementation in Scala and Akka, I couldn't figure out how to get this to work, so I don't support it)

comments? @joshfix

joshfix:
[...]

PR #719, opened 2020-01-22
Release 0.9.0-rc2

cholmes:
Fixed up changelog and updated version numbers

**Related Issue(s):** #
#718, #717, #714, #712, #710, #708, 

**Proposed Changes:**

1. Update version numbers to 0.9.0-rc2
2. Added changelog info for asset description and data role, and moved paging changelog info from rc1 to rc2

**PR Checklist:**


cholmes:
I'm not sure if I need to do more on the `npm run generate-all`. I just did a big find and replace on the 0.9.0-rc1 string. After I tried running the npm and `git diff` said no files changed, so I think this may be good?

m-mohr:
We have a problem with the CHANGELOG here due to an unclear process. I thought the next version would be stable so I didn't add my PRs to the CHANGELOG as usually we merge the rc and the stable CHANGELOG and they are only fixing things that have been introduced with 0.9, so no changelog entry for 0.8 -> 0.9 required. We should clarify how we do RCs and CHANGELOGs. I always feel strange about merging the RC changelogs into stable as then you don't have changelogs between RCs and it leads to this confusion. Also, we would need to remove some of Alex changes when merging as they are not changes between 0.8 and 0.9. So we need to decide for one path and then update the changelog.

cholmes:
I guess I naturally went to option 1) - it seems like there should be a way for people to track what happened between an RC1 and an RC2. Though most changes are minor it is possible a bigger change would occur, and we'd want an implementor to be able to easily see that it changed. And just feels better to track everything.

I think the one tweak I could see is when we make the release notes for 0.9.0 or any released candidate then we combine all the changes for the candidates. And when we blog it out we talk about all the changes. But it seems like the change log should track all the changes.

As for #711 - I missed that one. I had thought at the end of the last call we wanted to try to move ahead on an RC with what we had. I'm cool to wait, but we should try to get that fixed asap. I'm mostly keen to start on splitting the repos, but I suppose we also could branch 0.9.x if we do want it to bake for longer. That may also make a 0.9.1 release easier if we need it, since dev will start to change substantially.

cholmes:
@matthewhanson - you need to approve this, since I made the original PR, even though you did the update. 

But I think we should consider including #734 - or at least something that addresses the concerns raised, even if not that exact breakdown. We haven't pushed out 0.9.0, but it makes a bunch of substantial changes, based on assumptions about angles being different in aerial imagery, which they were not. 

So let's decide on that one on monday and then try to release.

PR #723, opened 2020-01-28
Small schema corrections

anayeaye:
**Related Issue(s):** #
N/A

**Proposed Changes:**
1. Fix an old ref link to instrument extension in sat schema
2. Updates the list of stac_extensions in item and collection schemas to match current extensions directories

**PR Checklist:**

### DIFF
--- a/collection-spec/json-schema/collection.json
+++ b/collection-spec/json-schema/collection.json
@@ -39,7 +39,8 @@
                   "asset",
                   "commons",
                   "checksum",
-                  "scientific"
+                  "scientific",
+                  "version"
                 ]
               }
             ]
--- a/collection-spec/json-schema/collection.json
+++ b/collection-spec/json-schema/collection.json
@@ -39,6 +39,7 @@
                   "asset",
                   "commons",
                   "checksum",
+                  "datacube",
                   "scientific",
                   "version"
                 ]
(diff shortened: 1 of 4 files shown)


item-spec/examples/landsat8-sample.json
```diff
@@ -75,7 +75,6 @@
     "metadata": {
       "href": "http://landsat-pds.s3.amazonaws.com/L8/153/025/LC81530252014153LGN00/LC81530252014153LGN00_MTL.txt",
       "type": "mtl",
-      "role": "metadata",
```
m-mohr: What is this change about?

collection-spec/json-schema/collection.json
```diff
@@ -39,7 +39,8 @@
                   "asset",
                   "commons",
                   "checksum",
```
m-mohr: I think datacube is also missing.

item-spec/examples/landsat8-sample.json
anayeaye: 'role' is replaced by 'roles' a few lines down

collection-spec/json-schema/collection.json
anayeaye: Thanks, I missed that

PR #724, opened 2020-01-30
update language around HTTP methods and content types

philvarner:
**Related Issue(s):** #711 Inconsistency in STAC.yaml and api-spec.md for search endpoint supported methods.

**Proposed Changes:**

1. Remove mention of multipart/form-data
2. Wordsmithing

**PR Checklist:**


m-mohr:
What are the opinions regarding https://github.com/radiantearth/stac-spec/issues/711#issuecomment-579662883?

api-spec/api-spec.md
```diff
@@ -38,22 +38,21 @@ STAC API.
...
-The OAFeat and STAC APIs follow a RESTful model. A core principal of this is the use of HTTP Request Methods (verbs) and
-the `Content-Type` header to drive behavior. This section describes how these are used in the OAFeat and STAC endpoints. 
+The OAFeat and STAC APIs follow a RESTful model. A core principal of this is the use of HTTP Request Methods ("verbs") and
+the `Content-Type` header to drive behavior on resources ("nouns"). This section describes how these are used in the OAFeat and STAC endpoints. 
 
-1. **Required** GET (both OAFeat and STAC)
-2. **Recommended** POST `Content-Type: application/x-www-form-urlencoded` with the corresponding content body format.
-3. **Recommended** POST `Content-Type: multipart/form-data` with the corresponding content body format.
-4. **Optional** **STAC endpoint /search only** POST `Content-Type: application/json`, where the content body is a JSON 
+1. **Required** Both OAFeat and STAC: GET using query string parameters
```
m-mohr: does this translate to: all GET endpoints are required?

api-spec/api-spec.md
```diff
@@ -38,22 +38,21 @@ STAC API.
...
+1. **Recommended** Only STAC: POST `Content-Type: application/x-www-form-urlencoded` with the corresponding content body format equivalent to the parameters defined for GET.
```
m-mohr: The list is strange... (1, 1, 1, ...?)

Also, I'm still not sure what it means. Does this translate to: "All POST endpoints are recommended to implement this content type" (and does it include the Transaction extension)? Or is this only valid for POST /search?

api-spec/api-spec.md
philvarner: Yes. Implementing all OAFeat endpoints and `GET /search` is required be considered "compliant" with STAC API. I took this approach for two reasons: 1) STAC is supposedly an extension of OAFeat, so I didn't see how that could be the case an NOT require the OAFeat endpoints to be implemented 2) We've had the convention elsewhere that **something** has to be required in each extension.

api-spec/api-spec.md
philvarner: Markdown doesn't care what number you actually put there, it just generates a sequence, so I tend to just use 1. so I don't have to constantly renumber them when modifying the order.

philvarner: This is specifically referring to `/search`, as that is the only endpoint part of the base STAC API. 

This would not apply to other API extensions that have POST endpoints, such as Transactions, as they don't also have`GET + query string` defined. Also, IIRC, all of the Transactions endpoints are just the OAFeat ones with POST + JSON applied. I can clarify this. 

Maybe we should also **recommend** that the OAFeat GET endpoints also have `POST form-urlencoded` implemented?

philvarner:
> I still don't understand the list and it confuses me. Can we get this into full sentences?

Yes, let's discuss the semantics more and then I'll rework the wording.

api-spec/api-spec.md
cholmes: I'd be cool to recommend that.

cholmes:
@philvarner - what's the additional semantics discussion needed? I'd love to get this merged asap so we can push out the 0.9.0 rc1 release.

m-mohr:
I looked into the issue today, and here's the wrap up: https://github.com/radiantearth/stac-spec/issues/711#issuecomment-583308460 So there's still quite a bit to do.

matthewhanson:
Weird, I didn't merge this one, I merged #733, but perhaps because that one used this PR as a base it counted as those commits getting merged?

cholmes:
Oh yeah, I've had that before - I think since all the commits were merged it updates the PR as merged. Which doesn't seem wrong... But yes, a little surprising.

Issue #732, opened 2020-02-07
Add x-www-form-urlencoded recommendation


cholmes:
In #724 we agreed that it would be good to have `x-www-form-urlencoded` as a recommendation. But we decided to take it out for RC2 because it is not specified anywhere, and has no explanation as to why someone might use it.

With that text it should easily get back into the spec, but we want to get rc2 / 0.9.0 out asap.

PR #733, opened 2020-02-07
Api spec updates

cholmes:
**Related Issue(s):** #724 'update language around HTTP methods and content types' was the start of this PR.
#711 'Inconsistency in STAC.yaml and api-spec.md for search endpoint supported methods' is also
related. #732 also fell out of this work, as we want to move it back in.

**Proposed Changes:**

1.  Remove mention of `multipart/form-data` and `x-www-form-urlencoded`. The former we decided should not be in the spec at all. The latter we agreed should be in the spec, but it needs more specification on how to implement it, and a bit of explanation as to the purpose. #732 will track that.
2. Various wordsmithing and clarification.

**PR Checklist:**


cholmes:
I didn't run the API generation, but I don't think it's necessary as it's just text changes? @matthewhanson - would be great if you could confirm that, and run it if needed.

Also can we confirm that the openapi docs don't have any reference to the x-www-form-urlencoded stuff? To make sure it's fully removed.

api-spec/api-spec.md
```diff
@@ -38,28 +40,40 @@ STAC API.
...
-1. **Required** GET (both OAFeat and STAC)
-2. **Recommended** POST `Content-Type: application/x-www-form-urlencoded` with the corresponding content body format.
-3. **Recommended** POST `Content-Type: multipart/form-data` with the corresponding content body format.
-4. **Optional** **STAC endpoint /search only** POST `Content-Type: application/json`, where the content body is a JSON 
-object representing a filter, as defined in the [STAC API OpenAPI specification document](STAC.yaml).  
-5. **Prohibited** **OAFeat endpoints only** POST `Content-Type: application/json`, where the content body is a JSON 
-object representing a filter. This is prohibited due to conflict with the 
+#### GET
+
+**Required**: OAFeat 1.0 only specifies GET query parameters, and they are all required. STAC's cross-collection `/search` also requires
```
m-mohr: Are we talking about endpoints or parameters here? I think originally it was about endpoints.

matthewhanson:
API generation is not needed. OpenAPI docs look fine. 

I'm approving, we just need another.

Issue #738, opened 2020-02-12
Extension title/name/prefix definitions are a bit confusing


joshfix:
The extensions directory contains a table that lists all of the extension names and prefixes, however the "name" value is actually a "title" and not the name that should be used in the extensions field.  Additionally, each extension's readme doesn't always clearly state what the prefix is even though it's included in parenthesis after the title.  The only clear documentation on what value to use for any given extension's name is at the bottom of the parent extension directory's README.

I propose updating the table in the extension directory's README to clearly distinguish between title, name, and prefix.  In addition, we should clearly add the title, name, and prefix at the top of each extension's README.

PR #739, opened 2020-02-12
Updated extension documentation to clearly define title/name/prefix

joshfix:
**Related Issue(s):** # https://github.com/radiantearth/stac-spec/issues/738

**Proposed Changes:**

1. Updated extension documentation to clearly define each extension's title, name, and prefix

**PR Checklist:**


extensions/README.md
```diff
@@ -45,22 +45,22 @@ stable for over a year and are used in twenty or more implementations.
...
-| [Satellite](sat/README.md) (`sat`)                           | Item             | Satellite related metadata for data collected from satellites. | *Proposal* |
-| [Scientific](scientific/README.md) (`sci`)                   | Item, Collection | Scientific metadata is considered to be data that indicate from which publication data originates and how the data itself should be cited or referenced. | *Proposal* |
-| [Single File STAC](single-file-stac/README.md) (-)           | ItemCollection   | An extension to provide a set of Collections and Items as a single file catalog. | *Proposal* |
-| [Versioning Indicators](version/README.md) (-)               | Item, Collection | Provides fields and link relation types to provide a version and indicate deprecation. | *Proposal* |
-| [View Geometry](view/README.md)                              | Item             | View Geometry adds metadata related to angles of sensors and other radiance angles that affect the view of resulting data | *Proposal* |
+| Extension Title                                | Name             | Prefix   | Scope                     | Description                       | Maturity |
```
m-mohr: Should we replace "name" with "identifier"? "name" is so ambiguous. Then we could say in the "stac_extensions" definitions that the identifier needs to be used. And then at the end of this doc where it explains that stac_extension uses the folder names say that the identifier is the folder name.

m-mohr: And maybe put the name/identifier and prefix column at the end? From a user perspective seems to be the least interesting information in the overview table.

joshfix: I'm good with all of these changes, but I don't think it's necessary to convey to the users that the directory name is the identifier.  The `identifier` field in the extension's README and the `identifier` in each extension's README should make it crystal clear.

m-mohr: It's more as a clarification for extension writers. Not sure whether it's completely clear to "newbies" that the identifier = folder name?! And I'm really just speaking to put it at the end of the file where it already speaks about folder names.

joshfix: I totally agree, but so far the confusion has come from implementors because it's a bit weird to only be able to identify an extension by it's folder name.  It's just yet another nuance -- The directory name should match the identifier, but the instructions to consumers/users shouldn't be to go look at an extension's path to find the directory name, it should just be something that is defined in the extension itself (which it is with this PR).  So I'm just modifying the text to state that extension creators should name the directory with the extension's identifier.

joshfix: How stuck are you on these fields being at the end of the table?  It seems most tables like this in tech documentation will have all of the various identifiers, unique values, etc at the beginning and longer text descriptions as the last column.

I don't know if we want to make this decision based on the neatness of the raw markdown, but due to the varying links of each entry's description, adding fields after that makes the markdown very nasty.  

It may look cleaner in both the raw and rendered markdown if we instead move the maturity field ahead of the description and leave the description as the last column?  Thoughts?

m-mohr: Agreed.

Not so stuck on the order. I just looks a bit messy on the rendered GitHub html pages and repeats quite a lot (EO, eo, eo).

joshfix:
meh, git didn't like the backticks i used in my commit message so it cut it short

collection-spec/collection-spec.md
```diff
@@ -23,19 +23,19 @@ Implementations are encouraged, however, as good effort will be made to not chan
...
-| title        | string            | A short descriptive one-line title for the collection.       |
-| description  | string            | **REQUIRED.** Detailed multi-line description to fully explain the collection. [CommonMark 0.29](http://commonmark.org/) syntax MAY be used for rich text representation. |
-| keywords     | [string]          | List of keywords describing the collection.                  |
-| license      | string            | **REQUIRED.** Collection's license(s), either a SPDX [License identifier](https://spdx.org/licenses/), `various` if multiple licenses apply or `proprietary` for all other cases. |
-| providers    | [[Provider Object](#provider-object)] | A list of providers, which may include all organizations capturing or processing the data or the hosting provider. Providers should be listed in chronological order with the most recent provider being the last element of the list. |
-| extent       | [Extent Object](#extent-object) | **REQUIRED.** Spatial and temporal extents.    |
-| summaries    | Map<string, [*]\|[Stats Object](#stats-object)> | A map of property summaries, either a set of values or statistics such as a range. |
```
anayeaye: Something funny happened to the formatting on this line

extensions/datacube/README.md
```diff
@@ -1,6 +1,10 @@
-# Data cube Extension Specification (`cube`)
+# Data cube Extension Specification
 
-**Extension [Maturity Classification](../README.md#extension-maturity): Proposal**
+- **Title: Data Cube**
+- **Identifier: datacube**
+- **Prefix:cube-**
```
anayeaye: extra `-` on prefix

extensions/view/README.md
```diff
@@ -1,6 +1,10 @@
-# View Geometry Extension Specification (`view`)
+# View Geometry Extension Specification
 
-**Extension [Maturity Classification](../README.md#extension-maturity): Proposal**
+- **Title: View Geometry**
+- **Identifier: view**
+- **Prefix: -**
```
anayeaye: prefix: view

extensions/datacube/README.md
m-mohr: And a missing space ;-)

collection-spec/collection-spec.md
m-mohr: That's the added space in `Map<string, [*]\  |[Stats Object](#stats-object)>`. I think Josh thought it's a table separator, but it was only an "or" (i.e. an escaped `|`) and thus corrupts the table.

extensions/checksum/README.md
```diff
@@ -1,6 +1,10 @@
-# Checksum Extension Specification (`checksum`)
+# Checksum Extension Specification
 
-**Extension [Maturity Classification](../README.md#extension-maturity): Proposal**
+- **Title: Checksum**
+- **Identifier: checksum**
+- **Prefix: checksum**
```
m-mohr: Should we change the label to "Field Name Prefix" or so to be a bit more clear?

collection-spec/collection-spec.md
```diff
@@ -23,19 +23,19 @@ Implementations are encouraged, however, as good effort will be made to not chan
...
+| title          | string            | A short descriptive one-line title for the collection.       |
+| description    | string            | **REQUIRED.** Detailed multi-line description to fully explain the collection. [CommonMark 0.29](http://commonmark.org/) syntax MAY be used for rich text representation. |
+| keywords       | [string]          | List of keywords describing the collection.                  |
+| license        | string            | **REQUIRED.** Collection's license(s), either a SPDX [License identifier](https://spdx.org/licenses/), `various` if multiple licenses apply or `proprietary` for all other cases. |
+| providers      | [[Provider Object](#provider-object)] | A list of providers, which may include all organizations capturing or processing the data or the hosting provider. Providers should be listed in chronological order with the most recent provider being the last element of the list. |
+| extent         | [Extent Object](#extent-object) | **REQUIRED.** Spatial and temporal extents.    |
+| summaries      | Map<string, [*]\  |[Stats Object](#stats-object)> | A map of property summaries, either a set of values or statistics such as a range. |
```
m-mohr:
```suggestion
| summaries      | Map<string, [*]\|[Stats Object](#stats-object)> | A map of property summaries, either a set of values or statistics such as a range. |
```

extensions/README.md
```diff
@@ -45,22 +45,22 @@ stable for over a year and are used in twenty or more implementations.
...
+| [Satellite](sat/README.md)                     | sat              | sat      | Item                      | *Proposal* | Satellite related metadata for data collected from satellites. |
+| [Scientific](scientific/README.md)             | scientific       | sci      | Item, Collection          | *Proposal* | Scientific metadata is considered to be data that indicate from which publication data originates and how the data itself should be cited or referenced. |
+| [Single File STAC](single-file-stac/README.md) | single-file-stac | -        | ItemCollection            | *Proposal* | An extension to provide a set of Collections and Items as a single file catalog. |
+| [Versioning Indicators](version/README.md)     | version          | -        | Item, Collection          | *Proposal* | Provides fields and link relation types to provide a version and indicate deprecation. |
+| [View Geometry](view/README.md)                | view             | -        | Item                      | *Proposal* | View Geometry adds metadata related to angles of sensors and other radiance angles that affect the view of resulting data |
```
m-mohr:
```suggestion
| [View Geometry](view/README.md)                | view             | view     | Item                      | *Proposal* | View Geometry adds metadata related to angles of sensors and other radiance angles that affect the view of resulting data |
```

extensions/datacube/README.md
m-mohr:
```suggestion
- **Prefix: cube**
```

extensions/view/README.md
m-mohr:
```suggestion
- **Prefix: view**
```

collection-spec/collection-spec.md
joshfix: I actually went back and tried to make some of the tables a little cleaner.  In this instance, it just so happens, when looking at the raw markdown, that the pipe lines up with the other table separator pipes.  However it's mere coincidence -- I simply didn't make every row have the same fixed-width for the Type field because there are only 3 lines that are much longer than all of the others.  I'll just go ahead and extend them all out so it looks cleaner.

joshfix: Or maybe I did introduce new spaces before the "or"?  Lol not sure, but I'll fix it up either way.

m-mohr: It was the latter, see the fix below :-)

joshfix:
I just noticed the label extension has a schema.json file in the `label` directory and not in a `json-schema` subdirectory.  Should I move it as a part of this PR?

m-mohr:
> I just noticed the label extension has a schema.json file in the `label` directory and not in a `json-schema` subdirectory. Should I move it as a part of this PR?

Yes, good idea!

joshfix:
I guess one other thing real quick... Since we are tending to use very human readable long-form titles, should we re-title "EO" to "Electro-Optical"?

m-mohr:
Would help to clarify, so yes from my side.

joshfix:
> Would help to clarify, so yes from my side.

Ok... super small change, but just re-requesting reviews to ensure nothing was missed.

PR #742, opened 2020-02-20
Fixed links in the Landsat example in the collection-spec.

m-mohr:
**Related Issue(s):** None

**Proposed Changes:**

1. The links in the landsat example were too API centric and linked with wrong rel types to an endpoint that contains data not part of the core spec. (.../items). See related Gitter conversations starting from Feb. 19 15:41 by @ycespb, @joshfix, @cholmes and me.

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Fixed
+- Fixed links in the Landsat example in the collection-spec.
+
 ## [v0.9.0-rc2] - 2020-02-12
 
 ### Added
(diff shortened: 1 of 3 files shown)


m-mohr:
The CI fails due to some connectivity issue, it's not related to this PR.

joshfix:
I think it is confusing to have an arbitrary path prefix that isn't related to STAC in the examples, especially because it's "stac-api" which seems to imply a relationship to the STAC API spec. 
 IMHO it should be `https://example.com/collections/landsat-8-l1` instead of `https://example.com/stac-api/collections/landsat-8-l1`.

matthewhanson:
Discussed this in working session, closing in favor of a new PR that uses static catalog instead of API.   The sat-api prefix as the root catalog is confusing, and there is no browse endpoint in the spec.

PR #746, opened 2020-02-27
Proj Tweaks

cholmes:
**Proposed Changes:**

1. Added links for the projection extension
2. Removed proj4 strings as an option, see discussion in the PR

**PR Checklist:**


extensions/projection/README.md
```diff
@@ -28,8 +28,8 @@ projections per Asset is not currently handled by this extension.
 | Field Name       | Type                     | Description |
 | ---------------- | ------------------------ | ----------- |
 | proj:epsg        | integer\|null   | **Required** [EPSG code](http://www.epsg-registry.org/) of the datasource |
-| proj:proj4       | string \|null   | PROJ4 string representing the Coordinate Reference System (CRS) that the `proj:geometry` and `proj:bbox` fields represent |
-| proj:wkt2        | string \|null   | WKT2 string representing the Coordinate Reference System (CRS) that the `proj:geometry` and `proj:bbox` fields represent |
+| proj:proj4       | string \|null   | [PROJ4 string](https://proj.org/usage/projections.html) representing the Coordinate Reference System (CRS) that the `proj:geometry` and `proj:bbox` fields represent |
```
hobu: I personally think that allowing proj4 strings as definitions for CRSs is a mistake going forward. PROJ is trying to undo them because there are so many things that are not representable (and will never be). I would think proper authority codes (EPSG, ESRI, IAU, etc), and then WKTv2 plus maybe PROJJSON if you wanted to be really progressive would be sufficient. What prompts the desire for proj4 strings?

cholmes: @philvarner - any thoughts on just taking out proj4 strings? Curious if you had a specific use case for them, or just included them to give options.

m-mohr: In openEO we deprecated them, but still allow them for some time as quite a lot of people use them to process data. For STAC, I think it's reasonable to remove (or at least deprecate) them as we are not handing input of an average user, but we have either dedicated clients or data providers who can take the burden of converting to any of the alternatives once.

cholmes: Ok, discussed on stac call 3/30/20 - removing the proj4 string.

m-mohr:
I removed some more proj4 references.

m-mohr:
@cholmes This has no changelog entry, I realized. Could you add this?

cholmes:
Ah shoot. Yeah.

cholmes:
Ok, just added #758 - sorry about this one, I checked the box when it was just link changes, and forgot when the bigger change came out of the PR conversation.

Issue #749, opened 2020-03-02
Remove API checklist item from issue template


matthewhanson:
With API spec moved to new repo, this can be removed from the issue template (and should probably be added into https://github.com/radiantearth/stac-api-spec issue template)

m-mohr:
Not 100% sure about this, I think there should still be a reference to API spec. How do we ensure that the API always catches up with recent changes in the "core" spec (e.g. update collection fields or so)?

cholmes:
Sorry, made the PR before I saw @m-mohr's comment. But this is just the line in the template that says 'I have run `npm run generate-all` to generate the files'. 

I could see how we may want to use PR templates to help make sure people are updating the api spec. But I also could see that its the responsibility of the api spec to update fields / examples when it does a release. And it can choose the stac core version it wants to depend on, and may even skip a minor version. Like stac-api-spec 1.0.1 might depend on stac-spec 1.0.5, since it was just a lot of minor changes that didn't really matter to the api.

I think the api spec needs to declare what version of core it depends on, and then part of its release process needs to be to update all its openapi / examples to align.

PR #752, opened 2020-03-04
Remove api building requirement

cholmes:
**Related Issue(s):** #749

**Proposed Changes:**

1. Removes the requirement to run the api script, since we moved the api stuff out.

**PR Checklist:**


m-mohr:
Could we add something like: "I opened a PR #XXX for the API"? Not as required, but as indicator for the API spec people so that they can go through the spec PRs and afterwards and know what they still need to include into the API and what is already done?

cholmes:
@m-mohr - good idea, just added something that I think captured your intent.

PR #753, opened 2020-03-18
Update types in label extension for clarity

jisantuc:
**Related Issue(s):** #

none

**Proposed Changes:**

1. Change `label:properties` in the label extension from a list of optional strings (`["abcde", null, "foo"]`) to an optional list of strings (`["abcde", "foo"]` or `null`)
2. Update class objects in labels from a list of strings or numbers (`["abcde", 34, "foo"]`) to a list of strings or a list of numbers (`["abcde", "foo"]` or `[0, 1, 2]`)

**PR Checklist:**

- ~API only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-spec/blob/dev/api-spec/README.md#openapi-definitions).~

**Notes**

These changes come out of trying to interpret the types + words while working on azavea/stac4s#48


m-mohr:
Hi! We can't merge this into master. Please redo the PR for dev. Also, there's a bit much space around the pipes. Other than that the PR looks good.

m-mohr:
@jisantuc I just realized that the JSON Schema is not updated. `label:properties` doesn't allow `null` there and `classes` only allows strings. So either the readme or the schema needs to be changed so that both are aligned.

jisantuc:
@m-mohr I think this is good for another look

extensions/label/json-schema/schema.json
```diff
@@ -1,119 +1,135 @@
...
+  "id": "stac-label-item-schema.json#",
+  "title": "STAC Label Data Item Extension Spec",
+  "type": "object",
+  "properties": {
     "properties": {
+      "type": "object",
+      "required": [
+        "label:properties",
+        "label:classes",
+        "label:description"
```
m-mohr, line 12: label:type is missing

extensions/label/json-schema/schema.json
```diff
@@ -1,119 +1,135 @@
...
-                    "type": "string"
+          ]
+        },
+        "label:classes": {
+          "title": "Classes",
+          "type": "array",
+          "items": {
+            "type": "object",
+            "required": [
+              "name",
```
m-mohr, line 35: The doc says name and classes aren't required. Either fix it here or in the docs.

extensions/label/json-schema/schema.json
```diff
@@ -1,119 +1,135 @@
...
+        },
+        "label:methods": {
+          "title": "Method",
+          "type": "array",
+          "items": {
             "type": "string",
             "enum": [
-              "raster",
-              "vector"
+              "automated",
```
m-mohr, line 88: This doesn't follow the doc (highlighted text by me):

> Recommended to be a subset of 'automated' or 'manual', but **may be an arbitrary value.**

extensions/label/README.md
```diff
@@ -59,10 +59,10 @@ Some additional notes are given here for some of the core STAC Item fields and w
 | label:overviews   | \[[Label Overview Object](#label-overview-object)] | Overview | An Object storing counts (for classification-type data) or summary statistics (for continuous numerical/regression data). |
 
 #### Class Object
-| Field Name | Type              | Name    | Description |
-| ---------- | ----------------- | ------- | ----------- |
-| name       | string\|null      | Name    | The property key within the asset's each `Feature` corresponding to class labels. If labels are raster-formatted, use null. |
-| classes    | \[string\|number] | Classes | The different possible class values within the property `name`. |
+| Field Name | Type                 | Name    | Description |
+| ---------- | -------------------- | ------- | ----------- |
+| name       | string\|null         | Name    | The property key within the asset's each `Feature` corresponding to class labels. If labels are raster-formatted, use null. |
```
m-mohr, line 64: Wouldn't it make sense to require something here? Same for the other objects below...

extensions/label/json-schema/schema.json
matthewhanson: I don't think they should be required, so this should be fixed in the docs.

extensions/label/json-schema/schema.json
matthewhanson: So I think the whole enum just needs to be removed then.

extensions/label/json-schema/schema.json
jisantuc: @matthewhanson given your comment, I'm removing `required` here, though I'm not certain that's the right thing given that they are currently required here and not in the docs.

extensions/label/README.md
m-mohr: Yeah, agreed. We should make a separate issue for that.

extensions/label/json-schema/schema.json
m-mohr: I think the required should be in place and be added to the docs. How does an empty object makes any sense? Now you can specify [{}, {]] and it's valid?!

jisantuc: Required in both as of 04edb4c

extensions/label/README.md
jisantuc: I see -- I misunderstood the relationship between your comments. I believe that 04edb4c addresses this as well.

PR #754, opened 2020-03-20
adjusted required assets extension fields to agree with json-schema

anayeaye:
**Related Issue(s):** #
N/A

**Proposed Changes:**
1. The `Asset Object` fields in the `assets` Collection extension did not reflect the required fields in the json-schema
2. Change not made: `type` is required for `Asset Object`s in the Collection extension but not in Items. Do we want to remove the type field requirement for Collection-Item agreement and/or open an issue to discuss?

**PR Checklist:**

- [x] ~~API only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-spec/blob/dev/api-spec/README.md#openapi-definitions).~~


m-mohr:
Aligning JSON schema and the documentation is good, but I'm wondering whether the schema is actually correct? I understand that there should be something valuable in the asset, but the item spec itself only requires `href`. What if I just want to specify roles and type for example? @matthewhanson How was that meant to be?

anayeaye:
I am also in favor of having more flexibility with which asset fields to implement.

m-mohr:
We could set the `minProperties` to 1 or 2...

matthewhanson:
I agree that title and type probably should not be required.

cholmes:
What's the status of this with all the recent moves? Is it ready to be reviewed / merge in?

m-mohr:
No, the points above still need to be changed.

m-mohr:
As this PR has a heavy merge conflict due to the item asset rename, I also have some fixes in PR #797 and this minProperties thing is still in discussion (I opened issue #799 to keep treck), I'd think we can close this and make a new PR once we have consensus on #799. Do you agree, @anayeaye ?

Issue #681, opened 2019-11-25
Move Item Recommendations to Best Practices?


m-mohr:
There's a section "Recommendations" at the end of the Item spec. Wouldn't those things make more sense in the best practices? 

cc @cholmes

matthewhanson:
Agree to put these in one place

cholmes:
Yeah, I think it makes sense to put all recommendations in one place, in best practices.

Reading the metadata linking section though, I think we should be sure to mention a couple key things in the main item spec:

1) The item spec is the core set of fields, and most any real world implementation will add in their own fields, either from extensions or defining their own. In my quick read of the main spec we don't really say that - we just start talking about the required fields. I think it'd probably be good to flesh out a bit more in best practices on how to pick the right fields (like first check extensions for reuse, make schemas for your own defined data, etc)
2) The focus of STAC is on search - I think we should also say that in the main item spec.

In best practices we can make the point that you can reference a metadata document as an asset.

PR #764, opened 2020-04-12
Item spec rework [WIP]

cholmes:
**Related Issue(s):** #681 

**Proposed Changes:**

1. Move the two best practices in item spec to the 'best practices' document
2. Add rel 'alternate' with type html recommendation to communicate that there is an html version. (can make an issue of this to discuss)

**PR Checklist:**

### DIFF
--- a/api-spec/extensions/sort/README.md
+++ b/api-spec/extensions/sort/README.md
@@ -24,13 +24,13 @@ provided before the field name, it will be assumed to be "+".
 
 Examples of `sortby` parameter:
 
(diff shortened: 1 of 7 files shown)


cholmes:
May want to also add in a bit more of a mention in the item spec that fields should be added for 'search'. We do mention the reason for the item spec is search at the top, but a bit more might be good.

m-mohr:
There are also changes in the PR coming from #763 though...

cholmes:
Ah shoot, I'll redo it. My git skills are still not great.

Issue #693, opened 2019-12-16
Published/Created/Updated dates


palmerj:
I'm interested in item/collection properties for:

- Created Date time | When the collection/item data was created
- Published Date time | When the collection/item data was published

Would there be a community appetite if I proposed an extension for this?

m-mohr:
Related issue: #351

cholmes:
Just working on my Planet STAC mapping, and we have updated and published (in addition to 'acquired', which maps to STAC datetime). Would be nice to use a shared definition for these for sure.  

So @palmerj - a proposal for these would be great, and I'd try them out in my Planet stuff and give feedback. I'm not sure of your difference between created and published, but happy to see the full proposal. Actually looks like #351 lays out differences and cites other works, so would be good to be aligned with that.

matthewhanson:
We have added `created` and `updated`, meaning when the metadata was first created, and when it was last updated.  They are in the common metadata for Items:
https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md

In using these however I do wish the terms were a bit more clear.  `created` and `updated` imply to me a True/False, or Yes/No.    `created_at` and `updated_at` may be better, or `created_datetime` and `updated_datetime` to be explicit.

m-mohr:
> In using these however I do wish the terms were a bit more clear. `created` and `updated` imply to me a True/False, or Yes/No. `created_at` and `updated_at` may be better, or `created_datetime` and `updated_datetime` to be explicit.

I disagree. I don't find them confusing and they are aligned with many other standards, e.g. Dublin Core and DCAT, which also just use the verb without any suffix. Although, they use these terms usually to specify dates about the data (i.e. assets) and not the metadata. That's something we may need to change. I guess it could be an option to rename them to metadata_created or so and leave created to the actual creation time of the data itself. Although a single date for multiple assets may be a bit problematic?!

palmerj:
> So @palmerj - a proposal for these would be great, and I'd try them out in my Planet stuff and give feedback. I'm not sure of your difference between created and published, but happy to see the full proposal. Actually looks like #351 lays out differences and cites other works, so would be good to be aligned with that.

Thanks @cholmes. I'm on holiday for the next month, but when I return I will put a proposal together, as well as for #150 (Needed for our full use case).

> I guess it could be an option to rename them to metadata_created or so and leave created to the actual creation time of the data itself.

@m-mohr I like this idea.

[...]

PR #769, opened 2020-04-13
Changed created to published

cholmes:
**Related Issue(s):** #693 

**Proposed Changes:**

1. Change 'created' to 'published', as it feels clearer. The 'creation' of the metadata file makes me think about when the file was created, and what creation of metadata even means - is it when the image arrives in the system? Published to me makes it clear that it's when the record first appears in the catalog.

**PR Checklist:**

### DIFF
--- a/item-spec/common-metadata.md
+++ b/item-spec/common-metadata.md
@@ -142,5 +142,5 @@ Fields to describe the metadata file itself. These fields do NOT describe the as
 
 | Field Name | Type   | Description |
 | ---------- | ------ | ----------- |
-published    | string | Date and time the metadata file was first published. This is NOT the timestamp the asset was created. MUST be formatted according to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). |
+| published  | string | Date and time the metadata file was first published. This is NOT the timestamp the asset was created. MUST be formatted according to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). |
 | updated    | string | Date and time the metadata file was updated last. This is NOT the timestamp the asset was updated last. MUST be formatted according to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). |
--- a/item-spec/common-metadata.md
+++ b/item-spec/common-metadata.md
@@ -142,5 +142,5 @@ Fields to describe the metadata file itself. These fields do NOT describe the as
 
 | Field Name | Type   | Description |
 | ---------- | ------ | ----------- |
-| created    | string | Creation date and time of the metadata file. This is NOT the timestamp the asset was created. MUST be formatted according to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). |
+published    | string | Date and time the metadata file was first published. This is NOT the timestamp the asset was created. MUST be formatted according to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). |
 | updated    | string | Date and time the metadata file was updated last. This is NOT the timestamp the asset was updated last. MUST be formatted according to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). |
--- a/item-spec/examples/sample-full.json
+++ b/item-spec/examples/sample-full.json
@@ -34,7 +34,7 @@
         "url": "https://cool-sat.com/"
       }
     ],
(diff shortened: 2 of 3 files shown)


m-mohr:
Not 100% sure, but why don't we just add published instead of replacing it?

And then there are lint errors for the PR... ;-)

cholmes:
Oh nice, the CI is working! I was, uh, just testing to see if it would ;)

From talking to @matthewhanson his use of 'created' is equivalent to 'published', so I think it'd be even more confusing to have both options, as well as 'datetime' which is supposed to be 'acquired'.

I'm ok if we just stick with created, just wanted to raise it before 1.0-beta1 as I believe it's our last chance to change it. And the discussion in #693 stalled out a bit.

palmerj:
As mentioned in https://github.com/radiantearth/stac-spec/issues/693#issuecomment-612701159 I'm ok with replacing `created` with published `created`. I also think the added complexity of having both will create confusion for users.

m-mohr:
Okay, I just wanted to figure out whether there may be people who later define `created` again as they think it's useful in addition to `published`. But if published generally is better and there's no real need for a separately defined `created`, then we could afford the change, I guess.

matthewhanson:
I think this is a good change, and agree having both `created` and `published` would be confusing. To me I don't see a need for differentiating when metadata is created and when it is published.

@cholmes just need a CHANGELOG entry here

m-mohr:
Looking at https://github.com/radiantearth/stac-spec/issues/693#issuecomment-613354106 again, I think we should rename `created` to `metadata_published` and `updated` to `metadata_updated`. Reason is that we use datetime without prefix for data/asset-related info so I'd expect updated and created to refer to data/asset related timestamps, too. With the metadata prefix it is clear what is meant, I think.

cholmes:
@m-mohr - I'm conflicted on the metadata prefix... In general I like more specificity, future-proofing to potentially add in more specific time fields in later. But I also think that fields for the full timeline of potential changes is overkill, and may make sense to have something more general in the core.

The case I think is easiest to make for 'published'. I don't see any realistic case where data_published and metadata_published would be meaningfully different. The act of publishing data is creating metadata and putting it in a catalog - to me the 'publish' applies to both the metadata and the data. And I would be very confused if there were two fields and I had to figure out which to pay attention to. So I think 'published' should apply to both - I'm fine if we specify that it's technically the metadata, but with STAC I don't see a case where people publish their metadata and then publish the data - indeed to me it's important that we don't have metadata records bereft of actual data.

For updated I agree that there *could* potentially be value in data_updated different than metadata_updated. But at planet we just use 'updated' to apply to both - when we bundle adjust the image changes slightly, but often the change is too small to affect the bounding box. But we also use it like when we achieve ground lock, and then the ground lock metadata field changes. I see value in a single field that can be used flexibly by data providers, that means either data or metadata has changed. And then there can be extensions that provide more specificity. That could be data_updated and metadata_updated fields (and then they could choose to still use updated to mean either has, or could choose to not use it), or it could be more of like a field that provides information on the latest change, potentially using versioning to make like a changelog. 

So I've actually been confused in the past if the created and updated fields would really work for Planet, since the spec defines it as just applying to the metadata and not the data. I think my preference is to have published to be defined as when you published your metadata and data. And updated to be defined to be able apply to either changes in the data or the metadata.

m-mohr:
Okay, I just feel like there's a long-term strategy yet on temporal fields and I guess it would feel more comfortable to agree on things once we have a final decision on how we want temporal fields to ultimately behave in STAC (see #693). I think it would be a good idea to actually go through the timeline in the issue and see how we want to handle each of them and then it's clear how the current fields fit in there.

matthewhanson:
@cholmes @m-mohr 
I don't have a strong opinion here, other than I'm already using 'created' in the Sentinel-2 catalogs and would need to update 18 million Items. 

So I'm fine with leaving it as is and revisiting later, or just changing it to published.

m-mohr:
For openEO, I can say that we'll stick with created until we have a field that describes the creation date of the data. published doesn't capture it for us, but otherwise I'm pretty neutral on the change...

@matthewhanson Have you exported the S2 in 0.9 or 1.0? Because in 0.9 is will still be created, no need to change.

cholmes:
Ok, between Matt having to change lots of records (I think he was targeting a 1.0-beta) and created working better for Matthias I think it's fine to just leave this as is. Planet's need isn't urgent, and they'd be fine to use created if that's how things go, or to use published if we end up promoting it in stac - I just personally find created a bit more confusing.

Issue #766, opened 2020-04-13
Clean up warnings / work in progress notes


cholmes:
The spec has had a lot of warnings about how it's a work in progess and may change. As we move to 1.0-beta1 we should tone those down - emphasize that things can change and encourage people to give feedback, but also let people know things are pretty stable.

PR #771, opened 2020-04-13
Item spec tweaks

cholmes:
**Related Issue(s):** #681 #766 

**Proposed Changes:**

1. Move the two best practices in item spec to the 'best practices' document
2. Add rel 'alternate' with type html recommendation to communicate that there is an html version. (can make an issue of this to discuss), and updated example
3. Removed 'warning' on item spec, as it has been stable and we are moving to 1.0-beta

**PR Checklist:**

### DIFF
--- a/best-practices.md
+++ b/best-practices.md
@@ -2,7 +2,8 @@
 
 ## Table of Contents
 
-* [Fields and ID's](#fields-and-ids)
+* [Field and ID formatting](#field-and-id-formatting)
+* [Field selection and Metadata Linking](#field-selection-and-metadata-linking)
 * [Static and Dynamic Catalogs](#static-and-dynamic-catalogs)
 * [Catalog Layout](#catalog-layout)
 * [Use of Links](#use-of-links)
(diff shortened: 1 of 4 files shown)


best-practices.md
```diff
@@ -32,6 +33,19 @@ the value consist of only lowercase characters, numbers, `_`, and `-`. Examples
...
+## Field selection and Metadata Linking
+
+In general STAC aims to be oriented around **search**, centered on the core fields that users will want to search on to find 
+imagery. The core is space and time, but there are often other metadata attributes that are useful. While the specification is 
+flexible enough that providers can fill it with tens or even hundreds of fields of metadata, that is not recommended. If 
+providers have lots of metadata then that can be linked to in the [Asset Object](item-spec/item-spec.md#asset-object) 
+(recommended) or in a [Link Object](item-spec/item-spec.md#link-object). There is a lot of metadata that is only of relevance 
+to loading and processing data, and while STAC does not prohibit providers from putting those type of fields in their items, 
+it is not recommended. For very large 
+catalogs (hundreds of millions of records), every additional field that is indexed will cost substantial money, so data
```
m-mohr, line 45: Although I basically agree with the indexing issue, I doubt that people blindly index everything, do they? Like you could skip everything complex (i.e. objects like bands) because the API can't search them anyway. If I'd implement something like that, I'd probably implement a white list of fields to be indexed beforehand...

cholmes: People definitely blindly index everything. Oftentimes the person doing the index is a dev who has no idea about geospatial or the use case of search. And they're asked by someone who has no idea about software/databases/indexes, and when the dev asks them if they should index everything they'd be inclined to say yes, as they don't understand the effects, and why not? That's why I think it's important to have the recommendation - to spark the conversation between diverse people.

matthewhanson: I know which fields should be indexed, and I still index everything, just because I hadn't gotten around to figuring out how I should be indexing certain fields in Elasticsearch but still store the entire record.

Issue #756, opened 2020-04-02
"grids" extension


matthewhanson:
There is a PR open for a "grids" extension: #688 , but there was no corresponding issue.

The grids extension is meant to provide information on the size in pixels and geolocation for each asset:
For each asset:
- the size of the 2 or 3-dimensional array stored within the file, as reported by rasterio or GDAL
- the "geo-transform" array - a 6 element array indicating the top left native pixel coordinate and coefficients for x and y in each of the 2 dimensions.

While this metadata is not used for searching on, it can be useful as it provides information that is otherwise only available by opening and reading the header of each asset. This can be a significant overhead cost when reading lots of files.

matthewhanson:
This extension is only going to be useful if used in conjunction with the `proj` extension, as that provides data on the native projection that the geo-transform units are provided in.

I don't find the name `grids` very helpful in determining the purpose of this extension.  I would propose either renaming it to something more descriptive such as `geo`, or better yet to combine it with the `proj` extension, since it must be used with that extension and thematically fits in with it.

matthewhanson:
cc @abhijeetkumarga @alexgleith @cholmes

alexgleith:
Hey @matthewhanson, I have no problem with merging it in with the `proj` extension.

I don't think `geo` is any better than `grids`, though.

cholmes:
I agree that geo is no better than grids. And my first reaction is that
merging with proj also makes sense, but want to think on it a bit more.

matthewhanson:
Just a note that most of the discussion here is within the PR #688 

@Kirill888 @alexgleith 

We discussed and will be proposing #760 , which will allow STAC assets to override properties that are described at the Item level.   Also agree that the grids extension is better merging with proj.

So now, the new fields from the grids extension become simpler, just two new fields to be added to an Item:

```
proj:shape
proj:transform
```

Then, by using the asset specific metadata option from #760 you can override `proj:shape` and `proj:transform` at the asset level.  In the case of Sentinel where there are 3 different transforms used due to different resolutions, we would define the most common one in Item properties (the 6 20m bands), then would override those for the 4 10m bands and the 3 60m bands (only 2 for L2 data since B10 is not included in L2).  

This does mean some redundant information, but it is a small number of bytes and we didn't feel it was worth the added complexity.

thoughts?

Kirill888:
@matthewhanson sounds good to me.

matthewhanson:
[...]

PR #780, opened 2020-04-17
Update proj extension to include grid

alexgleith:
**Related Issue(s):** #756

**Proposed Changes:**

1. Incorporate the proposed grid extension changes into the projections extension
2. Include two new properties, `proj:shape` and `proj:transform`

**PR Checklist:**

### DIFF
--- a/extensions/projection/README.md
+++ b/extensions/projection/README.md
@@ -74,7 +74,13 @@ defined in latitude and longitude, even if the data coordinate system does not u
 The number of pixels should be specified in Y, X order. If the shape is defined in an item's properties it is used as 
 the default shape for all assets that don't have an overriding shape.
 
-**proj:transform** - An array of affine transformation coefficients that are compatible with Rasterio as defined in the GDAL [`GetGeoTransform`](https://gdal.org/api/gdaldataset_cpp.html#_CPPv4N11GDALDataset15GetGeoTransformEPd) or the Rasterio [`Transform`](https://rasterio.readthedocs.io/en/stable/api/rasterio.io.html#rasterio.io.BufferedDatasetWriter.transform). If the transform is defined in an item's properties it is used as the default transform for all assets that don't have an overriding shape.
+**proj: transform** - Linear mapping from pixel coordinate space (Pixel, Line) to projection coordinate space (Xp, Yp). It is a `3x3` matrix stored as a flat array of 9 elements in row major order. Since the last row is always `0,0,1` it can be omitted, in which case only 6 elements are recorded. This mapping can be obtained from GDAL([`GetGeoTransform`](https://gdal.org/api/gdaldataset_cpp.html#_CPPv4N11GDALDataset15GetGeoTransformEPd)) or the Rasterio ([`Transform`](https://rasterio.readthedocs.io/en/stable/api/rasterio.io.html#rasterio.io.BufferedDatasetWriter.transform)).
+```
+  [Xp]   [a0, a1, a2]   [Pixel]
+  [Yp] = [a3, a4, a5] * [Line ]
+  [1 ]   [0 ,  0,  1]   [1    ]
+```
+If the transform is defined in an item's properties it is used as the default transform for all assets that don't have an overriding shape. 
 
 ### Item [`Asset Object`](../../item-spec/item-spec.md#asset-object) fields
 | Field Name | Type     | Description                                  |
--- a/extensions/projection/README.md
+++ b/extensions/projection/README.md
@@ -11,11 +11,13 @@ for "projection", and not a reference to the use of the PROJ/PROJ4 formats.
 
 The field names defined herein should be added as fields in the Item Properties object. 
 
-When specified on an Item, the values are assumed to apply to all Assets in that Item.  For example, an Item may have 
+When specified on an Item, the values are assumed to apply to all Assets in that Item. For example, an Item may have 
 several related Assets each representing a band or layer for the Item, and which typically all use the same CRS, 
-e.g. a UTM Zone.  However, there may also be assets intended for display, like a preview image or thumbnail, that have 
-been reprojected to a different CRS, e.g., Web Mercator, to better accommodate that use case.  This case of differing 
-projections per Asset is not currently handled by this extension.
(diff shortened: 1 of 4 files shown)


extensions/projection/README.md
```diff
@@ -52,20 +58,41 @@ If the data does not have a CRS, such as in the case of non-rectified imagery wi
...
+**proj:centroid** - Coordinates representing the centroid of the item in the asset data CRS.  Coordinates are
 defined in latitude and longitude, even if the data coordinate system does not use lat/long.
 
+**proj:shape** - An array of integers that represents the number of pixels in the most common pixel grid used by the item's assets.
+The number of pixels should be specified in Y, X order. If the shape is defined in an item's properties it is used as
+the default shape for all assets that don't have an overriding shape.
+
+**proj: transform** - Linear mapping from pixel coordinate space (Pixel, Line) to projection coordinate space (Xp, Yp). It is a `3x3` matrix stored as a flat array of 9 elements in row major order. Since the last row is always `0,0,1` it can be omitted, in which case only 6 elements are recorded. This mapping can be obtained from GDAL([`GetGeoTransform`](https://gdal.org/api/gdaldataset_cpp.html#_CPPv4N11GDALDataset15GetGeoTransformEPd)) or the Rasterio ([`Transform`](https://rasterio.readthedocs.io/en/stable/api/rasterio.io.html#rasterio.io.BufferedDatasetWriter.transform)).
```
m-mohr, line 79: There's a space in the field name ;-)

extensions/projection/README.md
```diff
@@ -52,20 +58,41 @@ If the data does not have a CRS, such as in the case of non-rectified imagery wi
...
+  [1 ]   [0 ,  0,  1]   [1    ]
+```
+
+If the transform is defined in an item's properties it is used as the default transform for all assets that don't have an overriding transform.
+
+### Item [`Asset Object`](../../item-spec/item-spec.md#asset-object) fields
+
+| Field Name | Type     | Description                                  |
+| ---------- | -------- | -------------------------------------------- |
+| proj:shape    | \[number] | See item description.  |
```
m-mohr, line 93: Above it's said that the shape consists of integers, so I'd propose to use \[integer].

extensions/projection/json-schema/schema.json
```diff
@@ -85,6 +85,24 @@
...
                 }
               }
+            },
+            "proj:shape":{
+              "title":"Shape",
+              "type":"array",
+              "minItems":2,
+              "maxItems":2,
+              "items":{
+                "type":"number"
```
m-mohr, line 95: Above it's said that the shape consists of integers, so I'd propose to use type: integer.

extensions/projection/README.md
```diff
@@ -9,30 +9,36 @@
...
 | ---------------- | ------------------------ | ----------- |
-| proj:epsg        | integer\|null   | **Required** [EPSG code](http://www.epsg-registry.org/) of the datasource |
+| proj:epsg        | integer \|null  | **Required** [EPSG code](http://www.epsg-registry.org/) of the datasource |
 | proj:wkt2        | string \|null   | [WKT2](http://docs.opengeospatial.org/is/12-063r5/12-063r5.html) string representing the Coordinate Reference System (CRS) that the `proj:geometry` and `proj:bbox` fields represent |
 | proj:projjson    | [PROJJSON Object](https://proj.org/usage/projjson.html) \|null   | PROJJSON object representing the Coordinate Reference System (CRS) that the `proj:geometry` and `proj:bbox` fields represent |
 | proj:geometry    | [Polygon Object](https://geojson.org/schema/Polygon.json)  | Defines the footprint of this Item. |
 | proj:bbox        | \[number]       | Bounding box of the Item in the asset CRS in 2 or 3 dimensions. |
 | proj:centroid    | Centroid Object | Coordinates representing the centroid of the Item in the asset CRS |
+| proj:shape       | \[number]       | Number of pixels in Y and X directions for the default grid |
```
m-mohr, line 40: Below it's said that the shape consists of integers, so I'd propose to use [integer].

extensions/projection/README.md
alexgleith: Oops!

extensions/projection/json-schema/schema.json
alexgleith: Good idea

extensions/projection/README.md
alexgleith: 👍

alexgleith:
Hey @m-mohr, thanks for the review. I think I've resolved all your comments. If you can check the schema, as it's a bit confusing as to where to put the asset level example. I think I have it in the right place!

matthewhanson:
@m-mohr @alexgleith I don't think there should be a schema for the asset....because this extension only adds Item properties.   It's a not-yet-created asset specific metadata extension that would allow these Item properties to be specified at the asset level.   I'm not sure how schema validation would happen there (perhaps allow any Item properties to be in the asset), but regardless it doesn't need to be handled in this extension.

m-mohr:
@matthewhanson I disagree, the extension defines 

> Item [`Asset Object`](../../item-spec/item-spec.md#asset-object) fields

So there should be validation for it OR the section mentioned above should be removed.

matthewhanson:
Ah you are right, thought that had been changed.

@alexgleith the fields should be moved out of the Asset object and in only the Item properties.

I'll get a PR up by end of this week with the Asset specific metadata extension

alexgleith:
Ah ok. Do we think that we should be explicit in the documentation that the item-level information can be over-ridden?

It's hard to discover if it isn't explicit in this extension.

alexgleith:
Ok, I've pushed a change to the schema and readme, but left a note in on the two properties that says "If the shape is defined in an item's properties it is used as the default shape for all assets that don't have an overriding shape."

Does that make sense?

matthewhanson:
Yes, I think this is one of the places it should say it can be overridden at the asset level because these are some of the fields for which that makes sense to do.

alexgleith:
> Looks mostly good, but some issues with number/integer. See the comments for details. Additionally, there's no JSON Schema for the Assets.

I think this has been resolved now, @m-mohr

Issue #783, opened 2020-04-26
Move eo:gsd into common Item metadata


matthewhanson:
gsd (ground sample distance), a measure of resolution, is useful for other data types beyond those using the `eo` spec. 

This should be moved from `eo:gsd` to `gsd` and appear in the common-metadata for an Item.

PR #683, opened 2019-11-26
EO: GSD -> Resolution #672

m-mohr:
**Related Issue(s):** #672

**Proposed Changes:**

1. Rename gsd to resolution in EO

**PR Checklist:**

- [ ] This PR has **no** breaking changes.
- [x] I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-spec/blob/dev/CHANGELOG.md) **or** a CHANGELOG entry is not required.
- [x] API only: I have run `npm run generate-all` to update the [generated OpenAPI files](https://github.com/radiantearth/stac-spec/blob/dev/api-spec/README.md#openapi-definitions).

Issue #672, opened 2019-11-19
EO: gsd -> resolution


m-mohr:
A thought from the working session:
Should we rename gsd to resolution?

I think that makes it easier to understand, so +1.

cc @matthewhanson

matthewhanson:
I originally pushed for GSD (Ground Sample Distance) as a way to be explicit about what the term is.

In digital cameras resolution is often given "Pixel Resolution", as total pixels in x and y (e.g., 10,000 x 8,000).

Sometime sensors specify "Spatial Resolution" (such as in medican imaging and with scanners) and is given as pixels-per-inch (ppi), or more generally pixels-per-some-unit

For optics, resolution is given in terms of angular separation. 

In remote sensing, resolution is usually provided in terms of Ground Sample Distance, or the distance on the ground between two adjacent pixel centers.

However, if we provide `resolution`, since we specify the units are to be in meters, I don't think there's any confusion what's being referred to and is clearer than gsd.

So +1 from me.

jbants:
I'd prefer to leave it as GSD as resolution can mean many different things like Matt was suggesting. Another "resolution" example would be "spectral resolution" where we look at the resolving power of a band to identify features in the EM spectrum. 

Leaving it as GSD is specific enough not to cause confusion with the other resolution types.

-1 for me

m-mohr:
Telco: Agreed to not rename it.

m-mohr:
From @jlaura:

> Here’s a legit medium article trying to help folks differentiate between GSD / resolution, etc. I suspect most non-geospatial nerds make this distinction? I know planetary scientists out of many labs do not.
> 
> https://medium.com/sentinel-hub/the-most-misunderstood-words-in-earth-observation-d0106adbe4b0

PR #784, opened 2020-04-27
Move eo:gsd to gsd

matthewhanson:
**Related Issue(s):** #
#783 

**Proposed Changes:**

gsd (ground sample distance) is a measure of resolution and is applicable to any raster data, not just to the eo extension. 
Moving `eo:gsd` into the Common metadata for items (under the instruments sub-heading) as an optional field for anyone to be able to specify this measure of resolution.

**PR Checklist:**


m-mohr:
Out of curiosity: Is gsd a term also used in other domains? resolution comes to mind as a broader alternative...

jjrom:
> Out of curiosity: Is gsd a term also used in other domains? resolution comes to mind as a broader alternative...

"resolution" is also the term used in OGC 13-026r8 OpenSearch Extension for Earth Observation (http://docs.opengeospatial.org/is/13-026r8/13-026r8.html)

cholmes:
So I do think there's value in aligning with opensearch for eo. But their structure is more flexible, where you specify the unit (meters, etc) and the value of your gsd. 

My worry about using a more general term would be that other domains define it differently.

matthewhanson:
We've discussed the idea of switching to resolution as the term - but it's a heavily overloaded term, see #672 for explanation and the PR that we closed without merging: #683

Issue #762, opened 2020-04-12
Remove Commons metadata / rework bands object


cholmes:
Opening an issue for discussion that came up on our last STAC call - considering removing the '[commons](https://github.com/radiantearth/stac-spec/blob/v0.9.0/extensions/commons/README.md)' extension. A few reasons came up:

* It's confusing, to have 'commons extension' and '[common metadata](https://github.com/radiantearth/stac-spec/blob/v0.9.0/item-spec/common-metadata.md)' 
* We now have 'summaries' which are a cleaner way to  do one of the things commons metadata was used for.
* The only case it seems really useful is the 'bands' object.

The key thing to do is to figure out another way to handle the bands. I wonder if one option is to just make it something that can be added at the collection level?

matthewhanson:
Some additional thoughts:

The Commons extension served two purposes originally.

One was to present users with what fields were the same across all Items in a Collection. As @cholmes points out the summaries extension now does that in a better way. Fields with a single value have a summary of one value, but can also includes summaries of all the fields.

The other was to avoid redundancy. However, in practice there's really only a couple fields (with the exception of `eo:bands`) that this ends up being used for, such as `instruments` or `constellation`. This is not much savings. And in an API it makes more sense to not use commons as it makes querying extremely difficult. Furthermore, the confusion to the user when they look at an Item and it does not actually contain all the values does not seem worth it.


The additional issue with the `eo:bands` properties is that it's a list of objects, so it's not queryable in any way currently. I also find it confusing to map out assets to bands.

I would propose two related changes:
- return the Assets definition to core as an integral part of a collection
- define spectral bands using the eo extension in the asset definition. Each asset will have an associated `spectral_bands` array of objects indicating the spectral bands in that asset.  This might introduce some redundant fields (multiple assets which contain the same band), but since this is at the collection level and note items this really isn't a big deal.

I suspect @m-mohr might have a problem with this since he has a use case where he doesn't have Items, but in that case I'm wondering how you use the `eo:bands` object anyway.

m-mohr:
If you are exposing data as cloud API / service as Google and openEO do, you don't really want to expose the structure of the assets to users so that's why assets won't work well. What we do is to just specify the bands in the summaries as with all other fields. (+ openEO uses the data cube extension and bands are a dimension.)

So, just having bands in assets is not ideal and we discussed last call that the "properties"/"summaries" should still include something if there's a field specified in the assets (like with gsd). Still thinking myself about the best solution...

[...]

Issue #779, opened 2020-04-16
Collection-level assets


m-mohr:
This issue is to capture some ideas, thoughts and related issues on collection-level assets. This came up several times in the past and we should document our thoughts and decisions.

Recently, collection-level assets came up in the following issues:
* #713: How to expose ZARR data?
* #778: How to expose collection thumbnails?
* #614: How to expose tiled assets? (There it was decided to go for a single item, but collection-level was also an option, but STAC had no option for collection-level assets).

A related extension is the "asset" extension for Collections. It allows to specify assets exposed in Items, but misses the actual link (`href`). Issue #661 asks for promoting this extension to Core. 

One idea I have is based on my similar ideas in #713 and #778: We could extend the `asset` property from the asset extension so that it optionally allows to have a `href`. If there's a href, it's a collection-level asset. Otherwise clients can assume that there's an item-level asset...

What do you think?

cc @cholmes @matthewhanson

matthewhanson:
Overall I agree that having Collection level assets could be useful.  I think thumbnails are a valid use case here, as are Zarr or other similar types of formats that represent a large collection of data spanning a very large region (such as the whole global) that are traditionally represented as individual tiled files.

However, I think the `asset` extension is clearly a different thing entirely.  The asset extension is for indicating the types of assets that can be expected in member items, whereas these are assets for the entire collection.

What if you have a collection with Items and use both the asset extension as defined, but also a collection level asset?  Is it just that one of the assets has an href?

One possibility is to treat Collection level assets the same as for Items (so under an `asset` field), but change the asset definition to something else that clearly indicates that it is a definition of assets rather than assets themselves.

m-mohr:
@cholmes @matthewhanson So how do we proceed regarding the Zarr dicussion and upcoming mini-sprint? All my examples and thoughts went towards collection-level assets and I still think that's a good idea. If you think it's better going forward with Items, you should propose a new example and we should also start doing PRs for whatever needs to be changed (datetime changes?) or at least make a list of ToDos. Otherwise we don't haven't good a good basis for the mini-sprint and I really don't want to discuss these things there.

cholmes:
[...]

PR #785, opened 2020-04-27
Move eobands to collection

matthewhanson:
**Related Issue(s):** #
#762 

**Proposed Changes:**

This moves eo:bands from being an Item property to the Collection level only (top-level).
eo:bands was not searchable, being a nested object, and being at the Item level lent itself to potentially a lot of unnecessary duplication.

**PR Checklist:**


m-mohr:
In principle, I like the change, but there are some things we should think about:
* We've seen EO as pretty stable and it's pretty commonly used. So I think we should survey some more users than the usual reviewers whether this changes is fine for most.
* How would a "single item" with bands in a single file look like? I have a use case where we respond to processing requests in our HTTP API with a single item and this change would lead us to some drastic changes, I guess?!
* Having bands and "band references" (this array indices pointing to the bands in the assets) in separate files makes them even more fragile to errors, I think (at least if you haven't used commons before). If you change something in the Collection, you must somehow ensure that all your Items are also updated whereas before it was just integrity you had to ensure for a single file.

m-mohr:
A general question: As we are moving the "complex" structure out of the Items, is that basically a "best practice" for extensions with "complex structures" to put them in Collections?

matthewhanson:
I think the best practice is "don't use complex objects" in Item properties, and instead separate them out into individual fields.

However, we do use complex objects at the Item and Collection level, but always (well, if we move eo:bands out) at the top-level, not in the properties.

So I think the general recommendation is complex structure never in properties, and instead put the in the top level, then on the API side consider supporting the querying of those with new query parameters.


ugh, looks like I've got some merge conflicts to work through

m-mohr:
I'm wondering whether it could make sense to put eo:bands in summaries? Then there's a central place for collection metadata and it may be easier to index/search on collection metadata at some point. (I'm actually thinking to do a "worldwide" STAC collection index at some point to experiment with collection-level search...)

matthewhanson:
Summaries are defined as being a summary of what is present in the Item properties. Since eo:bands is not in the Item properties this seems strange, it wouldn't be working like everything else in summaries. It would also be a dictionary quite unlike the stats objects that we see in summaries.

jjrom:
> So I think the general recommendation is complex structure never in properties, and instead put the in the top level, then on the API side consider supporting the querying of those with new query parameters.

I don't see why complex structure - or any other item metadata property - should be put at the top level of the item. The big advantage to keep it within properties block is that any GeoJSON reader will display it which is not the case if you move it at the top level

m-mohr:
> Summaries are defined as being a summary of what is present in the Item properties.

Yes and no... Summaries have a focus on that, but are not restricted to that. Otherwise we couldn't put any data in there for Collections without Items. So they could also contain bands. 

> It would also be a dictionary quite unlike the stats objects that we see in summaries.

No, bands is an array! As such it would have exactly the same schema as it is defined now.

> I don't see why complex structure - or any other item metadata property - should be put at the top level of the item. The big advantage to keep it within properties block is that any GeoJSON reader will display it which is not the case if you move it at the top level

I agree with that.

matthewhanson:
Doh! Right, an array - wishful thinking on my part.

Well, I'm fine with moving it to summaries if I get some agreement on it, I can update the PR tonight.

jbants:
+1 on summaries. 

@m-mohr for your single item use case, could you use the single-file stac extension?

m-mohr:
> @m-mohr for your single item use case, could you use the single-file stac extension?

@jbants Yes, in theory I could. In practice the extension just blows up your JSON structure to a degree that I'm not willing to push that. It's like breaking a butterfly on a wheel. Then I'm more likely to not implement STAC at all for that endpoint or just go with collection-level assets (#779).

m-mohr:
There's another idea for bands in this PR: https://github.com/radiantearth/stac-spec/pull/786#discussion_r416440166

Bands could also be part of the asset definition in collections.

matthewhanson:
Ok, will redo PR to put into summaries...

matthewhanson:
Strike that, based on discussion in #786 going to consider moving to asset definition instead.

matthewhanson:
Closing this, see summary of discussion here: https://github.com/radiantearth/stac-spec/issues/762#issuecomment-621294872

Issue #685, opened 2019-11-26
Asset specific metadata


matthewhanson:
This has come up again in discussions on EO (#247) and SAR (#670).

In SAR we had `sar:bands` in the Assets, and with #670 it will be `sar:polarizations`, which is also an Item property.  And `eo:bands` is also a type of asset specific metadata.

So the question is, what sorts of fields are allowed in Assets?  Is it just these specific fields we allow, or do we want some ability to define other properties at the Item level?   This has come up with various other fields, including EPSG/projection.

PR #786, opened 2020-04-27
Remove commons extension

matthewhanson:
**Related Issue(s):** #
#762

**Proposed Changes:**

With eo:bands moved to the collection level there is not much need for a Commons extension at all, which is confusing and also difficult to implement server-side (most implementations will end up de-normalizing the database anyway).

This removes the commons extension entirely. Items should contain all properties rather than relying on the collection to define some of them.

**PR Checklist:**


matthewhanson:
Still need to update all the examples to remove `properties` from collections - they should be converted to summaries.

This PR already includes changes from #785, so that should be merged first.

Also, PR #784 should be merged before finalizing this PR, otherwise there is going to be a lot of merge conflicts.

m-mohr:
I'm **very** happy with these changes, but I'm not 100% sure about eo:bands in #785. So I'm not sure if it's a good idea to base this proposal on another proposal...

matthewhanson:
This is ready to go, but I'll leave it in draft form for now since it already contains the changes in #785 , so that should be merged first

extensions/eo/README.md
```diff
@@ -78,6 +76,26 @@ numbers of several popular instruments.
...
 The difference between the `nir`, `nir08`, and `nir09` bands are that the `nir` band is a wider band that covers most of the spectral range of 0.75μm to 1.0μm. `nir08` and `nir09` are narrow bands centered 0.85μm and 0.95μm respectively. The same goes for the difference between `lwir`, `lwir11` and `lwir12`. 
 
+## Item fields
+
+| Field Name       | Type                     | Description |
+| ---------------- | ------------------------ | ----------- |
+| eo:gsd           | number                   | **REQUIRED.** Ground Sample Distance at the sensor. |
+| eo:cloud_cover   | number                   | Estimate of cloud cover as a percentage (0-100) of the entire scene. If not available the field should not be provided. |
+
+### Ground Sampling Distance
```
m-mohr, line 86: Shouldn't the gsd be moved to core? Also, for Items just cloud_cover is left, that's not a lot actually and in principle there should be something required for Items, otherwise validation will basically break ;-)

extensions/eo/README.md
```diff
@@ -87,59 +105,58 @@ Asset definitions that contain band data should reference the band index. Each a
 | ---------- | --------- | -------------------------------------------- |
 | eo:bands   | \[number] | Lists the band names available in the asset. |
 
-See [example-landsat8.json](examples/example-landsat8.json) for a full example.
+See the [landsat8 Item example](examples/landsat8-item.json) for a full example.
 
 ```js
```
m-mohr, line 110: This example is strange. Is that meant to be a collection? If yes, why's there the Item view extension mentioned?

extensions/eo/README.md
```diff
@@ -18,18 +18,16 @@ natural focus, and encourage other sensors to make their own extensions. Once th
...
-It is not necessary, but recommended to use the [Commons extension](../commons/README.md)
-(see chapter "Placing common fields in Collections").
-
 If the data has been collected by a satellite, it is strongly recommended to use the [`sat` extension](../sat/README.md), which in turn requires the [Instrument Fields](../../item-spec/common-metadata.md#instrument). If the data has been collected on an airborne platform it is strongly recommended to use the [Instrument Fields](../../item-spec/common-metadata.md#instrument).
 
 For defining view geometry of data, it is strongly recommended to use the [`view` extension](../view/README.md).
 
-- [Example (Landsat 8)](examples/example-landsat8.json)
+- Examples:
+  - [Landsat-8 Collection](examples/landsat8-collection.json)
```
m-mohr, line 26: Why's the collection link gone? The file seems to exist...

extensions/eo/examples/landsat8-collection.json
```diff
@@ -0,0 +1,229 @@
...
+            "type": "text/html",
+            "title": "HTML index page"
+        },
+        "thumbnail": {
+            "title": "Thumbnail image",
+            "type": "image/jpeg"
+        },
+        "B1": {
+            "type": "image/x.geotiff",
+            "eo:bands": [
```
m-mohr, line 140: Now that I'm seeing this, I'm really wondering why we don't combine assets and eo extensions to make something that's more expressive. Like wouldn't it be charming to have the actual band objects here and then the Items could inherit that? Would avoid all those array indices.

extensions/eo/examples/landsat8-item.json
```diff
@@ -41,14 +41,17 @@
...
+        "instrument": ["oli", "tirs"],
         "datetime": "2018-10-01T01:08:32.033Z",
-        "eo:cloud_cover": 78,
         "view:sun_azimuth": 168.8989761,
         "view:sun_elevation": 26.32596431,
+        "view:off_nadir": 0,
         "landsat:path": 107,
-        "landsat:row": 18
+        "landsat:row": 18,
+        "gsd": 30,
```
m-mohr, line 53: Minor, but maybe move gsd up so that it's basically grouped with the other core fields?

extensions/eo/json-schema/schema.json
```diff
@@ -3,12 +3,26 @@
...
-  "allOf": [
+  "oneOf": [
     {
-      "$ref": "../../../item-spec/json-schema/item.json"
+      "allOf": [
+        {
+          "$ref": "../../../item-spec/json-schema/item.json"
+        },
+        {
+          "$ref": "#/definitions/eo"
```
m-mohr, line 13: I don't get how this schema works. Why are Item and Collection refer both to the same eo definition?

extensions/eo/README.md
matthewhanson: Hmmm, gsd must have made it's way back in there due to resolving merge conflicts.

Well, I suppose we could move cloud_cover also to the common Item metadata.

extensions/eo/examples/landsat8-collection.json
matthewhanson: Yes! I'm all for that.   I think I had brought this up at some point. I'd like to see the Asset definition made core (not an extension), then the eo:bands appears under each asset and is an array of Band objects directly rather than indices.

Would be so much more manageable and readable. There might be some redundancy if there are multiple assets that share a band, but it's not much.

eo:bands would only be used in the asset definition, so in order to find out band ordering for an asset in an item, you would need to get the collection.  I'm fine with that, I think it's better than having indices in the item that need to be looked up in the collection.

extensions/eo/json-schema/schema.json
matthewhanson: I have no idea - I was using another json-schema to figure out how define it and very likely got it all wrong.

extensions/eo/examples/landsat8-collection.json
m-mohr: I'd like that. I also feel assets (with a new name, see #779) should be core. As with summaries, the utility is high and just makes Collections really useful.

extensions/eo/json-schema/schema.json
m-mohr: I can give it a try?!

extensions/eo/README.md
m-mohr: It feels like cloud_cover is actually generic enough to be core, yes. Although... that would basically leave us with a "spectral bands extension" ;-)

extensions/eo/json-schema/schema.json
matthewhanson: Well, let's wait until we sort out what we're doing with eo:bands and cloud_cover then come back to this

extensions/eo/examples/landsat8-collection.json
matthewhanson: No, that's the nice thing about calling this an "asset definition" rather than assets, it can have different fields than assets in items.

We could also move other fields to the asset definition as well...we can discuss, we wouldn't want assets to just be reduced to an href.

I wouldn't want to have to go look up `type` in the asset definition for instance, but something like `roles` might be better off there.

m-mohr: It was not clear to me that you are allowed to add more than actually available in the item assets. I thought it's a strict copy of the available assets excluding the href.

What fields are you thinking about to move?

matthewhanson: Well you can't do that right now - the asset extension is simply just a copy of all the same fields in the assets, but we could define the asset definition differently.

The point of the asset definition is to indicate what assets are going to be available in a member item, so I could actually see all of the fields getting included in the asset definition. It would be good to know the title, description, type, and roles for each of the assets.

But at the same time I wouldn't want all that stuff removed from the Item assets...nor would I want to have to repeat all that same information so I'm not sure there's a straightforward answer here.

matthewhanson:
This PR presumed the change of eo:bands to collection (and incorporated that PR #785), but that was closed without merging as we are doing something different with that (see #685 and #788).

It's easier to close this and redo the PR from the dev branch.

Issue #760, opened 2020-04-09
Asset specific metadata


matthewhanson:
A few of us had a quick discussion today relating to asset specific metadata: the ability to specify fields that are Item properties on a per asset basis.

It's discussed here in the context of the proposed "grids" extension:
https://github.com/radiantearth/stac-spec/pull/688#issuecomment-609910006
but could apply to other fields as well.

We concluded this was a good feature to have, with the following notes:
- properties defined at the asset level should always be first defined at the Item level. The asset field, if present, overrides the Item field for that asset
- make clear that fields defined per asset will not be searchable....users will search on the Item properties so that should be some sort of "nominal" value
- recommendation should include a short list of fields this may make sense for, e.g., grids extension, gsd
- otherwise, providers should think carefully on whether or not they should be doing this, or if what they really need is separate Items for this data

did I miss anything @m-mohr @cholmes @anayeaye ?

m-mohr:
This seems mostly complete, although it doesn't clearly state how the related between asset-level metadata and common metadata + extension fields is. There's a potential conflict between the "title" fields, should be mentioned at least. Also, what's the take on things like eo:bands that differs between properties and asset level? I think that needs some thoughts and explanation. And then we probably need to clarify that the asset extension can also inherit the "added" fields.

m-mohr:
We just had a use case where we have to put the "datetime" field into the assets (i.e. export of a single band time series), which IMHO is reasonable to do although STAC Items are more modeled for single datetimes. And that makes me think of #613 again. ;-)

matthewhanson:
closed with #788

PR #788, opened 2020-04-29
Item asset properties

matthewhanson:
**Related Issue(s):** #
#760 

**Proposed Changes:**

Add the ability to specify Item properties under Assets. This is a very powerful ability for the usage of STAC metadata, with the drawback that it can hinder discoverability if it is used without consideration.  It allows for greater flexibility in creating collections and Items, especially if they tend to be non-homogeneous, such as data grouped for a specific mission (e.g., post disaster catalog of data).  It also allows us to remove some oddities in existing STAC core and extensions, such as sar:polarizations being defined for both an Item and Asset, and removes the need to reference an `eo:bands` array from assets.

@m-mohr discussed this morning and thought this really made sense as a core feature to STAC rather than an extension. It adds no fields, but instead adds what should be a core behavior getting us closer to a common metadata model.  It also is likely to be used by multiple other extensions.

**PR Checklist:**


item-spec/item-spec.md
```diff
@@ -215,6 +216,105 @@ Deprecation notice: GeoTiff previously used the media type `image/vnd.stac.geoti
...
+
+For `eo:bands`, it could be put in Item properties as an array of all the bands, but in this case it's not. Instead, the assets each define an array containing the spectral band information for that asset (in the order the bands appear in the file).
+
+##### Common Use Cases for Item Properties in Assets
+
+Overriding or providing Item Properties only in the Assets makes discovery more difficult and should generally be avoided. However, there are some core and extension fields for which providing them at at the Asset level can prove to be very useful for using the data.
+
+- `gsd`: Specify some assets with different spatial resolution than the overall best resolution.
+- `updated`: Provide individual timestamps on when an asset was last updated. Since the Item property only tells when any of the Item was updated, this can provide additional info for users to determine which asset(s) might have been updated.
```
m-mohr, line 311: Updated is for metadata only and can't be used for such a use-case, unfortunately.

m-mohr: What I think is more useful could be a datetime for timeseries.

item-spec/item-spec.md
```diff
@@ -165,6 +165,7 @@ or streamed. It is allowed to add additional fields.
 | description | string    | A description of the Asset providing additional details, such as how it was processed or created. [CommonMark 0.29](http://commonmark.org/) syntax MAY be used for rich text representation. |
 | type        | string    | [Media type](#media-types) of the asset. |
 | roles       | \[string] | The [semantic roles](#asset-role-types) of the asset, similar to the use of `rel` in links. |
+| *Item Property* | varies | Any Item property can be defined for each asset |
```
m-mohr, line 168: I think I'd remove this row and replace it as follows:

```suggestion

It is allowed to add **additional fields**:
* [STAC Common Metadata](common-metadata.md#stac-common-metadata) - A list of fields commonly used throughout all domains.
* [Content Extensions](../extensions/README.md#list-of-content-extensions) - Domain-specific fields 
such as EO, SAR and point clouds.
* [Custom Extensions](../extensions/README.md#extending-stac) - It is generally allowed to add custom 
fields. [See above for more information.](#additional-fields).
```

Also, remove the "It is allowed to add additional fields." above.

item-spec/item-spec.md
```diff
@@ -215,6 +216,105 @@ Deprecation notice: GeoTiff previously used the media type `image/vnd.stac.geoti
 Cloud Optimized GeoTiffs used `image/vnd.stac.geotiff; profile=cloud-optimized`.
 Both can still appear in old catalogues, but are deprecated and should be replaced.
 
+#### Item Properties for Assets
+
+As detailed above, Items contain properties, which are the main source of metadata for searching across Items. Many content extensions can add further property fields as well. Any property that can be specified for an Item can also be specified for a specific asset. This can be used to override a property defined in the Item, or to specify fields for which there is no single value for all assets.
+
+**It is important to note that the STAC API does not faciliate searching across Asset properties in this way, and this should be used sparingly.** It is primarily used to define properties at the Asset level that may be used during use of the data instead of for searching.
```
m-mohr, line 223: Maybe also add some words about how repetitive or "larger" chunks can be put in asset definition using the extension?

item-spec/item-spec.md
matthewhanson: My thought for updated is that the link might be changed, in which case you have updated the metadata 

I'll add datetime

item-spec/item-spec.md
matthewhanson: I wasn't quite sure how to present it, I like the suggestion, will update

item-spec/item-spec.md
matthewhanson: I was going to tackle the idea of moving out parts of the assets into the Item Asset definition as a separate PR

item-spec/item-spec.md
m-mohr: Well, I think updated on assets would be quite confusing. People would think it's the time for the data, but the spec clearly says it's the metadata. Also, wouldn't updated in the item properties change also? Related: https://github.com/radiantearth/stac-spec/pull/769#issuecomment-621494830 - Maybe we need updated for data...

item-spec/item-spec.md
m-mohr: Okay, fine for me if there's a tracking issue?!

matthewhanson: Not yet - will create one

item-spec/item-spec.md
matthewhanson: Yes, if you updated an asset URL you'd update the Item properties.   But imagine that all the assets have an `updated` field, you'd be able to tell what the last one was that was updated.
Anyway, I'll remove it from this list.

matthewhanson:
@m-mohr Pushed update - not sure if this was what you meant but I added the additional fields directly in the table, but not sure it's really presenting the meaning of it well.

m-mohr:
The comment included a commitable suggestion. ;) I imagined it more like the list for Item fields where the list originated from. I think only concrete fields should be in the table. General stuff is better in text/lists...

item-spec/item-spec.md
m-mohr: I think there can never be multiple updated fields as it refers to the date the STAC metadata file was written. You'd need a separate field that is the asset update time.

matthewhanson:
Got it, pushed an update

item-spec/item-spec.md
```diff
@@ -166,6 +166,12 @@ or streamed. It is allowed to add additional fields.
 | type        | string    | [Media type](#media-types) of the asset. |
 | roles       | \[string] | The [semantic roles](#asset-role-types) of the asset, similar to the use of `rel` in links. |
 
+In addition to the fields above, [additional fields](#additional-fields) may be added to the assets. (See [Additional Fields for Assets](#additional-fields-for-assets))
+
+* [STAC Common Metadata](common-metadata.md#stac-common-metadata) - A list of fields commonly used throughout all domains.
+* [Content Extensions](../extensions/README.md#list-of-content-extensions) - Domain-specific fields such as EO, SAR and point clouds.
+* [Custom Extensions](../extensions/README.md#extending-stac) - It is generally allowed to add custom fields.
```
cholmes, line 173: I'd move these lists of links (common, content extensions, custom extensions) down to the section on 'additional fields for assets', like only present these links *after* we emphasize you should be really selective and why it's not the best idea to use these. Right now it reads like 'use all of these', and only later says 'probably don't use them'.

item-spec/item-spec.md
```diff
@@ -215,6 +221,105 @@ Deprecation notice: GeoTiff previously used the media type `image/vnd.stac.geoti
...
+#### Additional Fields for Assets
+
+As detailed above, Items contain properties, which are the main source of metadata for searching across Items. Many content extensions can add further property fields as well. Any property that can be specified for an Item can also be specified for a specific asset. This can be used to override a property defined in the Item, or to specify fields for which there is no single value for all assets.
+
+**It is important to note that the STAC API does not faciliate searching across Asset properties in this way, and this should be used sparingly.** It is primarily used to define properties at the Asset level that may be used during use of the data instead of for searching. 
+
+For example, `gsd` defined for an Item represents the best Ground Sample Distance (resolution) for the data within the Item. However, some assets may be lower resolution and thus have a higher `gsd`. The `eo:bands` field from the EO extension defines an array of spectral bands. However, it may be useful instead to specify the bands that are used in a particular asset.
+
+Here is a partial Sentinel-2 Item:
+
```
cholmes, line 233: I'd say move this to be an actual example? I think we've moved away from including longer json examples directly in the spec - I just cleaned up catalog in this way. Just make an example, and then link to it and talk about it.

item-spec/item-spec.md
```diff
@@ -215,6 +221,105 @@ Deprecation notice: GeoTiff previously used the media type `image/vnd.stac.geoti
...
+        }
+    }
+}
+```
+
+The Sentinel-2 overall `gsd` is 10m, because this is the best spatial resolution among all the bands and is defined in Item properties so it can be searched on. However, Band 5 has a `gsd` of 20m, so that asset specifies the `gsd` as well, which overrides the Item `gsd` for this one asset. Reduced resolution versions of files could also be included as assets and the `gsd` correectly represented.
+
+For `eo:bands`, it could be put in Item properties as an array of all the bands, but in this case it's not. Instead, the assets each define an array containing the spectral band information for that asset (in the order the bands appear in the file).
+
+##### Common Use Cases of Additional Fields for Assets
```
cholmes, line 311: I think this type of stuff probably belongs more in the 'best practices' document in the current organization. Core spec should say 'you generally don't want to use these, but see the best practices section for the cases where it often makes sense'. Keep the core spec focused on what is allowed / not, and then do more explanation in best practices (or the new 'overview' I'm working on, but I think this one makes sense in best practices).

item-spec/item-spec.md
m-mohr: Well, maybe in this case just link to the "Addtional Fields" section anyway as it's already there for Item Properties.

cholmes: Ah yeah, I like that approach.

Issue #789, opened 2020-05-01
Cleanup Item asset specific metadata, add to best practices


matthewhanson:
The recent PR #788 that adds in the ability to define Item properties for individual assets includes potential use cases and language that is better off under Best Practices.

m-mohr:
Should also tackle https://github.com/radiantearth/stac-spec/pull/788#discussion_r417642024

cholmes:
Ok, just made PR #790 - but I wasn't quite clear on what the action is from the comment on 788 that you mentioned @m-mohr ? If you explain it a bit more in the ticket I can add it in, I just felt like I didn't have the full context and wasn't sure if the stuff to link to has landed?

m-mohr:
Nevermind, I think it will be part of a separate PR from @matthewhanson that will establish behavior in the asset defintion extension to export "larger" properties like eo:bands into the asset definition of a collection.

Issue #445, opened 2019-05-02
Validate extension examples with circle ci


cholmes:
We are using CircleCI for checking the main examples, but it'd be great to also validate the examples in the 'extensions', making sure they are valid stac, and ideally valid to the extension fields.

cholmes:
I don't think we'll get to this for 0.8.0, moving out.

m-mohr:
See PR #834

Issue #298, opened 2018-10-16
 Format all examples using the same style


cholmes:
From #205 - this didn't quite get accomplished: 'Format all examples using the same style, e.g. by using https://jsonformatter.org/'

Would be good to do this before 0.6.0 final release. Is fine to do as it is no change to the spec. Would be great to also get a formatter for the examples and maybe json schemas as well in to continuous integration.

m-mohr:
I think we should enable JSON linting for the examples and schemas in the CI rather than let the CI reformat the code automatically. There should be a script that lets contributors check it. This could be combined with validation in a single script for CI and standalone usage before PRing things.

cholmes:
Cool, that approach makes good sense.

scisco:
Will do this after the content and api are separated to their own repos.

cholmes:
@scisco - we're split now, any chance you could tackle this in the next few weeks? Is in line with #704 - would be good to get everything clean for beta1

m-mohr:
That's probably something we could easily do for 1.0.0 final, it's just polishing things up.

scisco:
I can tackle it in the next few weeks. Although, Matthias suggests we should not do it at this point.

m-mohr:
I'm just thinking that doing it now would lead to duplicate work as we'd need to check it now and would probably check it again for final.

scisco:
yeah, it totally make sense.

cholmes:
Cool, sounds good. I'll set this for 1.0.0-RC1, as that seems like the right time to do it.

m-mohr:
I guess the best idea to do it together with #821? Maybe I can also include a small switch into stac-node-validator so that it's also linted during validation...

cholmes:
Yeah, definitely makes sense to do this after or as part of #821, so we're not fixing a bunch of examples we may throw out.

cholmes:
And yes, any tooling to help format and validate would be great to have.

m-mohr:
Re styling, I'd say the default should probably be what JavaScripts `JSON.stringify(value, null, 2)` returns?! Then I could quickly implement it in node-validator and then we could probably also just do it whenever we want as it's done automatically. Give me a week-end... ;-)

cholmes:
Yeah, that sounds good. I have no real preference on the style, just want something consistent, that is easy to apply. No rush on my end, I've got lots of tickets to take...

m-mohr:
See PR #941

[...]

PR #790, opened 2020-05-01
Asset metadata cleanup

cholmes:
**Related Issue(s):** #789 

**Proposed Changes:**

1. Moved section on common asset fields to best-practices
2. Moved example in item-spec.md to be in the s2 example
3. Cut the 60m assets from the s2 example, since they don't add much value imho and made the gsd stuff less clear. (Possibly controversial - the example is no longer aligns as closely with actual sentinel stac catalogs, but I think that's ok, as the examples in stac should be instructive imho. People can go
to the actual public catalogs to see the full example.

I don't believe this warrants a changelog entry, but happy to do one if others feel it should.

**PR Checklist:**

### DIFF
--- a/item-spec/examples/sentinel2-sample.json
+++ b/item-spec/examples/sentinel2-sample.json
@@ -55,6 +55,15 @@
       ]
     ]
   },
+  "properties": {
+    "collection": "s2-l2a",
+    "datetime": "2018-06-05T10:00:27.464Z",
+    "platform": "sentinel-2",
+    "gsd": 10,
+    "proj:epsg": 32635,
+    "eo:cloud_cover": 88.459539,
+    "view:sun_azimuth" : 176.091667178268
+  },
   "assets": {
     "metadata": {
       "href": "s3://sentinel-s2-l2a/tiles/35/V/MK/2018/6/5/0/metadata.xml",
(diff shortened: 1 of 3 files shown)


item-spec/examples/sentinel2-sample.json
```diff
@@ -55,6 +55,15 @@
       ]
     ]
   },
+  "properties": {
+    "collection": "s2-l2a",
+    "datetime": "2018-06-05T10:00:27.464Z",
```
m-mohr, line 60: That datetime (=> microseconds) is invalid, isn't it?

item-spec/examples/sentinel2-sample.json
```diff
@@ -55,6 +55,15 @@
...
+    "platform": "sentinel-2",
+    "gsd": 10,
+    "proj:epsg": 32635,
+    "eo:cloud_cover": 88.459539,
```
m-mohr, line 64: Some fields are required, but not listed (e.g. eo:gsd)

item-spec/examples/sentinel2-sample.json
cholmes: You mean that it's .464Z? And should be .46Z? I have no idea, looking at https://tools.ietf.org/html/rfc3339#section-5.6 it says time-secfrac    = "." 1*DIGIT which I think means one or more?

item-spec/examples/sentinel2-sample.json
cholmes: ? It looks like eo:gsd moved to just be 'gsd' in the [instrument section] of common metadata. https://github.com/radiantearth/stac-spec/blob/dev/item-spec/common-metadata.md#instrument 

This does highlight though that it's no longer required - could be good for EO extension to make it required. That should probably be another issue.

cholmes:
Cool. I didn't really make the example, just took the existing one and tweaked it. Happy to update it, but I'm not sure about the two changes you wanted. 

Do you have a recommended way to check the json schema of the examples? I was going to add info to 'contributing' on how to do the markdown checking locally.

item-spec/examples/sentinel2-sample.json
m-mohr: While microseconds is theoretically allowed, I see some tools in practice not liking them, so for interoperability and simplicity I generally try to avoid them if possible. I'd propose to change to `2018-06-05T10:00:27Z`.

item-spec/examples/sentinel2-sample.json
m-mohr: It seems I shouldn't make reviews at 1AM. You are right, eo:gsd has been moved to core, but then still eo:bands is required (at least at the moment of writing this, will likely change soon).

item-spec/examples/sentinel2-sample.json
```diff
@@ -55,6 +55,15 @@
       ]
     ]
   },
+  "properties": {
+    "collection": "s2-l2a",
```
m-mohr, line 59: This is a top-level field.

item-spec/examples/sentinel2-sample.json
cholmes: Cool, happy to change it. Should we put that recommendation somewhere in the spec? Could just be in best practices. Though I could see mention in the item spec. Just something like 'only using two decimal places is recommended, since while rfc3339 technically allows microseconds many tools don't understand them.

item-spec/examples/sentinel2-sample.json
cholmes: So what's my action on this one then?

item-spec/examples/sentinel2-sample.json
cholmes: Cool, will change it. This does make me think we probably need to review all the examples? I only did minor tweaks on this one, most of your catches were on things I did not change (I did move properties up for more visibility, but didn't really change anything with it).

item-spec/examples/sentinel2-sample.json
m-mohr: I guess that microsecond thing is quite a bit subjective. I saw several projects trying to get around datetimes with regular expressions and stuff that just doesn't catch microseconds, but I believe most datetime libraries handle it just fine. I think I'd just remove it from examples for simplicity and then see how it goes. We can still add something to BP later if there's really a problem with it.

item-spec/examples/sentinel2-sample.json
m-mohr: Add a `eo:bands` object to properties and make sure the `eo:bands` references in the assets are valid.

item-spec/examples/sentinel2-sample.json
m-mohr: Yes, before 1.0 we should definitely run all examples through validation (#298  / #445)

item-spec/examples/sentinel2-sample.json
cholmes: Ok, sounds good.

item-spec/examples/sentinel2-sample.json
cholmes: Ok, I looked into if I could do this easily, but I think it's beyond the scope of this PR for me. It looks like landsat also doesn't define its eo:bands references. And it seems like it'll change soon. I can add a ticket that says 'update item spec examples to have valid bands' or something.

m-mohr: In this case I'd simply remove all eo fields altogether, but at least the eo:bands in the assets, because that's really broken and confusing without having the referenced band array.

cholmes:
Ok, just fixed two of those. I think making things valid to our schema should be its own issue, though I did all the quick fixes on this. I was just trying to make the example work for the item asset text, instead of having a snippet directly in the text. So the bands definition is still not right / valid with our schema. But I'd prefer to merge this and ticket the remaining bits.

item-spec/examples/sentinel2-sample.json
m-mohr: Not sure, but why did you just remove one number from the decimals?

item-spec/examples/sentinel2-sample.json
```diff
@@ -55,6 +55,15 @@
       ]
     ]
   },
+  "properties": {
+    "datetime": "2018-06-05T10:00:27.46Z",
```
m-mohr, line 59:
```suggestion
    "datetime": "2018-06-05T10:00:27Z",
```

item-spec/examples/sentinel2-sample.json
m-mohr: See #797 for a "fix".

Issue #773, opened 2020-04-13
Document CI checks for local use


m-mohr:
Document all the command you can run on the command line to check the markdown, links, API etc.

Should be done for both the stac spec and the api spec.

PR #794, opened 2020-05-02
Contributing.md improvements - instructions to run check-markdown locally

cholmes:
**Related Issue(s):** #773 and a bit of #793 

**Proposed Changes:**

1. Reworked the contributing.md document to reflect maturity of spec
2. Added instructions on installing and running check-markdown locally

**PR Checklist:**

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,22 +1,56 @@
 ## Contributing
 
-Anyone building software that catalogs imagery or other geospatial assets is welcome to collaborate.
-Our goal is to be a collaboration of developers, not [architecture astronauts](http://www.joelonsoftware.com/articles/fog0000000018.html).
-A number of developers met in person and [sprinted on specs](https://github.com/radiantearth/boulder-sprint/)
-and made a number of key decisions. 
+Pull Requests are the primary method of contributing to the spec itself, and everyone is welcome to submit 
+changes. Suggestions for changes to the core will be taken with higher priority if a clear implementation 
+of STAC has been built that can highlight the problem. If the changes can be done as an [extension](extensions/) 
+instead of modifying the core files then that route is recommended first, and once there is uptake for the 
+extension it will be considered for core.
 
-The best way to contribute is to try to implement the specification and give feedback on what worked
-well and what could be improved. Currently there are a few major things to still work out, so if you
-are interested in implementing you should jump on our [gitter channel](https://gitter.im/SpatioTemporal-Asset-Catalog/Lobby)
-and check in on how ready the spec is and how you can support it. For the next bit of time you'll
-probably be encouraged to just do what feels best in some areas, so we can capture best practices
-and standardize.
+We consider everyone using the specification to catalog their data to be a 'contributor', as STAC is
+really about the end result of more interoperable data, not just creating a spec for the sake of it.
+So if you want to help then the best thing you can do is make new catalogs or build software that uses
+the spec. And please do give us feedback. The best place to do so is on our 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1,22 +1,56 @@
...
+
+Creating a Pull Request will show our PR template, which includes checkbox reminders for a number
+of things.
+
+* Adding an entry the [CHANGELOG](CHANGELOG.md). If the change is more editorial and minor then this 
+is not required, but any change to the actual specification should definitely have one.
+* Base the PR against dev, as mentioned above - even if the branch was made off of dev this reminds
+you to change the base in GitHub's PR creation page.
+* Make a ticket in the STAC API repo if anything here affects there.
+* Make sure the PR makes no breaking changes to the specification.
```
m-mohr, line 29: That sounds like you are not allowed to do breaking changes at all, although we still do.

CONTRIBUTING.md
```diff
@@ -1,22 +1,56 @@
...
+
+All pull requests additionally require a review of two STAC core team members. Releases are cut
+from dev to master (and require 3 approvals), see the [process](process.md) document for more details.
+
+There are also two additional checks done by the continuous integration system. The one most likely
+to catch you the markdown checker, which ensures we have clean markdown throughout the specification.
+You can click through to see what it caught. If you do not want to wait till you push to see if things
```
m-mohr, line 36: I think the preferable way it to check it first and don't push it first. So I think I would phrase it the other way round: Please check it with npm run ... or if you don't have time/tools, you could also let the CI check it...

CONTRIBUTING.md
```diff
@@ -1,22 +1,56 @@
...
+failed then you can use the same tool locally.
+
+### Using check-markdown locally
+
+The same check-markdown program that runs as a check on PR's is part of the repo and can be run locally. 
+To install you'll need npm, which is a standard part of any [node.js installation](https://nodejs.org/en/download/).
```
m-mohr, line 42:
```suggestion
To install you'll need npm, which is a standard part of any [node.js installation](https://nodejs.org/en/download/). Alternatively, you can also use [yarn](https://yarnpkg.com/) instead of npm. In this case replace all occurrences of `npm` with `yarn` below.
```

CONTRIBUTING.md
```diff
@@ -1,22 +1,56 @@
...
+
+First you'll need to install everything with npm. Just navigate to the root of the stac-spec repo and on
```
m-mohr, line 44:
```suggestion
First you'll need to install everything with npm once. Just navigate to the root of the stac-spec repo and on 
```

CONTRIBUTING.md
cholmes: Yeah, I got a little confused on that one. We are close to 1.0, and at that time we won't allow breaking changes. So went back and forth if it should be aspirational. But I'll tune it down for now.

CONTRIBUTING.md
cholmes: Ah, that's a great idea. Will fix it.

CONTRIBUTING.md
cholmes: First I've seen of the commitable suggestions. That's cool, will have to try them out.

cholmes:
Awesome, thanks for the review @m-mohr. I'll get those in and merge when I get another approval.

CONTRIBUTING.md
m-mohr: Really useful, would love to see more people use it.

cholmes:
Ok, changes made, thanks!

PR #772, opened 2020-04-13
Catalog spec break out and improvements

cholmes:
**Related Issue(s):** #696 #766 #667 

**Proposed Changes:**

1. Reorganized catalog spec so the spec is just the fields, and an overview contains the more expository info
2. Updated various links, as some were out of data
3. Added an example and updated the existing one, and removed the 'examples' section in the spec
4. Promoted 'versioning' to a higher level section, as it was in the middle of the 'use of links' section which didn't make much sense.
5. Removed warning, as this has been stable for awhile and we are moving to 1.0-beta1 #766 
6. Created overview.md in the root directory that talks through items and collections as well
7. Added section to explain catalogs vs collections in overview #catalog #667 
8. Refinements to the 'best practices' document, including a new section about putting vector data in STAC (recommending against it)

**PR Checklist:**

- [x] This PR is made against the dev branch (all proposed changes except releases should be against dev, not master).
- [x] This PR has **no** breaking changes.
- [x] I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-spec/blob/dev/CHANGELOG.md) **or** a CHANGELOG entry is not required.

PR #797, opened 2020-05-05
Fix absolute link +  example +  item assets

m-mohr:
**Related Issue(s):** Just a minor fix to PR #772 and #790

**Proposed Changes:**

1. Make the link relative
2. Remove eo from the example as we don't have the band information in there, making the band references invalid. Without eo:bands, we can't specify eo:cloud_cover, so removing, too. Feel free to add back the bands if you have suitable data.
3. Fix item assets (the key in the text was not updated) and the required fields were not marked. Also, it's better to have a distinct name in the schema definition to avoid conflicts.

**PR Checklist:**


m-mohr:
I'd see that as a separate PR.

Issue #792, opened 2020-05-01
Make datetime nullable for stac items for which a nominal datatime makes no sense


cholmes:
As discussed on the call on april 27, we want STAC to be able to represent data cubes and Zarr (#781), so we should relax our requirement that a datetime is needed - a time range should be used instead. The solution we liked was to make it required, but allow a 'null' value. 

I believe we said if a null is there then the start_datetime and end_datetime are required.

I'll take this on to start, but will need some help with the schema.

m-mohr:
I'll take on the schema.

m-mohr:
Just wanted to make the schema for it, but there's a bit of an issue as datetime is in the item spec and start/end_datetime are in the common metadata. Could we move datetime to common metadata or the other timestamps to the Item spec? As we are about to have a dependency between these fields, it seems also better to list them together in the documentation. I guess we prefer moving start/end_datetime to the item spec?

cholmes:
Hrm... Yeah, actually putting them in the item-spec.md at the exact same level is right over the line for me - I want it to be clear that 'datetime' is the core way. I'm ok to mention and link to them in the item-spec. And I'm ok if we put it in the schema because there's no other way to validate it. I don't think we actually need a one to one from item-spec.md to the item spec schema. To me the item-spec/ directory is 'the item spec', so common metadata is part of the 'spec'. But I really don't want to present two options in the core item-spec.md table.

I guess what I'm saying is I don't think it 'seems also better to list them together in the documentation'. That treats them as equal level. I was going to write the text to link from datetime to the fields as what you do if the datetime is null, but not to list them both.

m-mohr:
This separation sounds quite confusing. You require datetime, but if it's set to null go somewhere else to figure out the other fields required in this case? With adding null as accepted value you already show a second option in the item spec, but it's then written somewhere else what you need to use such an option. You can still say what is the recommended way of doing it. You can't just hide the "null" option in the item spec, that's really bad spec writing from a dev perspective. So strictly -1 on your proposal, sorry.

cholmes:
I'm not saying hide the null option. I should probably write the text and then we can debate it. But what I was thinking for the line would be:

| Field Name | Type   | Description                                                  |
| ---------- | ------ | ------------------------------------------------------------ |
[...]

PR #798, opened 2020-05-05
Make datetime nullable for stac items for which a nominal datatime makes no sense

m-mohr:
**Related Issue(s):** #792

**Proposed Changes:**

1. Make datetime nullable for stac items for which a nominal datatime makes no sense, but require start/end_datetime then.

**PR Checklist:**


item-spec/common-metadata.md
```diff
@@ -29,17 +29,21 @@ Descriptive fields to give a basic overview of a STAC Item.
...
 ## Date and Time
 
+- [JSON Schema](json-schema/datetime.json)
+
 Fields to provide additional temporal information such as ranges with a start and an end datetime stamp.
 
-### Date and Time Range
+| Field Name | Type         | Description |
+| ---------- | ------------ | ----------- |
+| datetime   | string\|null | See the [Item Spec Fields](item-spec.md#properties-object) for more information. |
```
m-mohr, line 38: I think it's valuable to mention this here so that we have a full overview here. Otherwise you have one field in the Item Spec and all the other here, which is a bit confusing.

cholmes: +1

Issue #778, opened 2020-04-16
Thumbnails for Collections?


m-mohr:
Anyone has a good idea how we could incorporate thumbnails for collections? We have them for items, but we have some cases where we want to directly give some examples for collections. 

Two way I could see here:
1. Add a link to a thumbnail with rel type [`preview`](https://www.iana.org/assignments/link-relations/link-relations.xhtml) and image type `image/jpeg` (or gif/png). Currently clients don't recognize this so I'd propose to add it to the Collection [Relation Types](https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md#relation-types).
    **Pro**: This is possible without a real spec change.
    **Con**: It's different from how items do previews.
2. Use the asset extension and add a `href`. This is also how we try to incorporate collection-level assets in #713. Based on this, define thumbnails as in the item spec.
    **Pro**: Is aligned with Item spec.
    **Con**: Needs a spec change (asset extension) and basically introduces collection-level assets.

Other options? Opinions?

matthewhanson:
I think if it were only for thumbnails then having it as a link would be good, but if we have a use for other collection level assets, such as using the Zarr format, then thumbnails should be specified as assets to be like collections.

PR #800, opened 2020-05-05
Collection-level Assets

m-mohr:
**Related Issue(s):** #778 #779

**Proposed Changes:**

1. Introduce a collection-level asset extension

**PR Checklist:**


cholmes:
Looks like a great draft. Will review the final one after the mini-sprint.

m-mohr:
Seems to be useful, would appreciate reviews now.

extensions/collection-assets/examples/example-esm.json
```diff
@@ -0,0 +1,97 @@
...
+  "license": "proprietary",
+  "links": [
+    {
+      "href": "https://pcmdi.llnl.gov/CMIP6/TermsOfUse/TermsOfUse6-1.html",
+      "type": "text/html",
+      "rel": "license",
+      "title": "CMIP6: Terms of Use"
+    }
+  ],
+  "assets": {
```
jisantuc, line 44: Could this key be something else to avoid collision with the existing `asset` extension that modifies collections? Or follow the `extensionName:key` pattern? As is it's impossible for a collection to implement both extensions

m-mohr: The existing asset definition extension has been renamed, partly because we want collection and item asset behave the same way.

jisantuc: I see, I got confused by the `dev` vs. `master` thing again.

m-mohr:
FYI: This will be part of the upcoming ESM extension and is now part of the openEO API, too. So there will be numerous implementations available in the next months.

extensions/collection-assets/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+A Collection extension to provide a way to specify assets available on the collection-level.
+
+- [Example](examples/example-esm.json)
+- [JSON Schema](json-schema/schema.json)
+
+This extension introduces a single new field, `assets` at the top level of a collection.
+An Asset Object defined at the Collection level is the same as the [Asset Object in Items](../../item-spec/item-spec.md#asset-object).
+
+Collection-level assets MUST NOT list any files also available in items.
```
jisantuc, line 17: Why not? This sounds hard to enforce and like it won't hurt much if someone lists an asset in two places

extensions/collection-assets/json-schema/schema.json
```diff
@@ -0,0 +1,22 @@
...
+      "$ref": "../../../collection-spec/json-schema/collection.json"
+    },
+    {
+      "type": "object",
+      "required": [
+        "assets"
+      ],
+      "properties": {
+        "assets": {
+          "$ref": "../../../item-spec/json-schema/item.json#/definitions/assets"
```
jisantuc, line 17: Since assets would no longer be an entity definition owned by items, would it make sense to migrate the "Asset Object" section into its own specification? Or is the rationale that, since this is an extension, asset objects actually are owned by items still, unless this extension is present on the collection?

m-mohr: Well, we have quite a lot of things that are shared across spec (Provider Object, Asset Object, Link Object, ...). I think we could move if it's used in core specs, but extensions shouldn't influence where we put things in the core spec. So would leave it as it is for now.

extensions/collection-assets/README.md
m-mohr: We want to avoid people put all their Item assets into collections. Indeed that's hard to enforce.

jisantuc: I have a preference against *MUST*s that can't be expressed in the JSON schema and that require a whole catalog of context to validate. I think this should be demoted to a best practice.

m-mohr: Well, we have quite a lot of things in the spec that can't be validated with JSON Schema though.

Tagging @cholmes as he was the main driver for this... I personally don't care so much about this rule.

cholmes: I'm fine with a 'should' / best practice here.

m-mohr:
```suggestion
Collection-level assets SHOULD NOT list any files also available in items.
```

m-mohr: Done.

PR #832, opened 2020-06-03
Fix JSON Schemas

m-mohr:
**Related Issue(s):** None

**Proposed Changes:**

1. Fix several issues found in JSON Schemas

**PR Checklist:**


cholmes:
Seems like we'll need a beta.2 soon, so people can use schemas that work?

m-mohr:
Now that I have validation working I'm going to check all examples and schemas, that will take some time.

cholmes:
Yeah, we feel a bit cursed lately. I'd say we should be checking everything before, but I think the point of the beta1 release is actually to really start on that. But we should build up more checks / infrastructure. Validate every example, and indeed I think we should shift to making sure the validator and core libraries are updated during the 'release candidate' process, even if that means longer release candidates. 

Thanks for catching it early and working on it @m-mohr

m-mohr:
Well, tooling and some more reviewers are missing to check beforehand, I think. Most tooling unfortunately isn't (logically, it's Open Source) always keeping up with the pace. I should be getting better with the 1.0.0 releases. I guess I could release my validator and easily put it in CI for now. It already checks all examples + extension schemas... I'll issue another PR with examples fixes.

matthewhanson:
I'd say we give beta1 a little time, see if there are any other issues that arise with schemas or examples, and add a bunch of examples.

PR #842, opened 2020-06-09
Add label:color field to item properties in label extension

jisantuc:
**Related Issue(s):** none

Why do this: frequently labels are generated by some external source. That source may have information about what color things should be. When that information is available, it's helpful to have that information available to downstream consumers, for instance, to programmatically generate a [`mapbox-gl-js` `"get"` expression](https://docs.mapbox.com/mapbox-gl-js/style-spec/expressions/#get). Also in that case, it's dramatically simpler than to build up a `match` with ad hoc color values for the available classes.

Also, mostly the changes are formatting because VSCode did it, but the important parts are:

- adding to the fields table: https://github.com/radiantearth/stac-spec/pull/842/files#diff-1f42ba0e9303b14ef5e45816cc39f70aR67
- adding to the schema: https://github.com/radiantearth/stac-spec/pull/842/files#diff-9f20905e8f9bba05c8e0cc1fe551f5b6R149-R152

**Proposed Changes:**

1. Add `label:color` field (not required) to label extension item properties

**PR Checklist:**

- [x] This PR is made against the dev branch (all proposed changes except releases should be against dev, not master). (🙈)
- ~This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec), and I have opened issue/PR #XXX to track the change.~


jisantuc:
I don't think a pattern should be added to the json schema -- the `label:color` field contains a pointer to the field in the `Feature`s of the `FeatureCollection`, it does not contain a color itself. As such it can be any non-empty string. I did however add an example value in a9f3303 to explain the sorts of values the field should point to.

m-mohr:
Oh, understood. Mis-read the description of the field. Is it for deduplication purposes that you link to a field instead of putting the color directly? Can't really be for space as the color is about the size of the field name.

jisantuc:
It's because the asset is a feature collection, so it can have many features, and there's no prior reason to believe that every feature should be the same color.

m-mohr:
Okay, I think for such behavior there should be an example, either by changing an existing one or adding a new one. Also, wouldn't it be more self-explanatory to call the field color_field or so? color implies I'd find an actual color in there.

jisantuc:
Sure, I can see how that would be misleading. Example incoming.

jisantuc:
I'm gonna close this for now in favor of a different extension, and track creating an extension for this purpose in Franklin. If it's useful, I'll at some point PR a real extension back, instead of overloading the label extension.

Issue #843, opened 2020-06-09
Empty Geometries?


m-mohr:
@abarton-airbus wrote on Gitter:

> Hello. My organization has implemented a STAC compliant catalog. We've come across an edge case that feels unclear in regards to compliance: STAC Items with an empty geometry don't have a clear definition for bbox.
>
> As noted by the spec (and its reference, https://tools.ietf.org/html/rfc7946#section-5), "The value of the bbox member MUST be an array of length 2*n where n is the number of dimensions represented in the contained geometries,.." An empty geometry would have 0 dimensions or bbox = []. However, this does not meet the schema definition (https://raw.githubusercontent.com/radiantearth/stac-spec/v1.0.0-beta.1/item-spec/json-schema/item.json).
> 
> It feels like this edge case may not be addressed? Suggestions?

My answer was:
> I'm not 100% sure we enforce this in the spec, but I'm pretty sure it was meant to be that STAC Items are not allowed to have an empty geometries.

We should check that and adopt wording if necessary.

Related: https://github.com/radiantearth/stac-spec/issues/367

cholmes:
+1 - to me geometry required is key. But yes, let's make sure it's clear in spec / schema.

abarton-airbus:
I'd just clarify that technically this use case is produced when a geometry is provided, it's just an empty geometry. So perhaps the spec/schema needs to not only require geometries, but require they are not empty.

cholmes:
Gotcha, thanks @abarton-airbus - that's a great clarification to be sure we have in the spec. Thanks for flagging it.

cholmes:
So in my mind the 'contract' of STAC is to ensure that everything has space and time. So I lean towards saying that an Item without a geometry (or set to empty) is not actually STAC compliant, but that we should allow non-geometry 'Items' in a catalog, but a STAC validator should only check those that have geometries. 

But that was more in response to people who want STAC to be their catalog for everything. I do think this use case is a bit different, it is things that are geospatial but aren't yet processed for their exact location. 

In the case of  near real time / level 1 products that aren't yet ground locked it does seem like it'd be valuable for a user to have the bounding box that reflects the satellite ephemeris if possible. But I can see that sometimes that isn't possible.

I think I still lean towards how @m-mohr tackled this. Matthias - do you use the same STAC structure but just change the media type? Or is it just an alternate JSON representation? And do you think that is ok, or would it be better for you if that's STAC with an empty geometry?

[...]

PR #849, opened 2020-06-16
WIP: Added anonymized location extension

kbgg:
**Related Issue(s):** #843

**Proposed Changes:**

1. Adding an "anonymized location" extension

**PR Checklist:**

At Radiant we've come across instances where we can not provide or do not have precise locations of labeled data. A solution we've come to is defining a grid across the globe and assigning each item to the grid square which it falls within. The geometry of the STAC item will then be set to that grid square instead of its actual location. Most of the time we just create 1x1 degree grid squares and that provides enough anonymization and geographic accuracy to still be able to train models.

There should still be some way for the user to know that the geometry present in the STAC item is not the actual geometry of the label item but is instead a grid square of where the item *could* be. The anonymized location extension being present would be an indication that the geometry should not be taken as true and the anon:precision property would indicate how large the grid squares are in terms of decimal degrees.

This could also solve the problem of data which has no geometry (#843). Instead of setting the geometry property to null you can define the grid as having only one grid square which covers the entire globe and the item falls within that one Earth-sized grid square.

This is still a WIP so any feedback on this solution is welcomed.


extensions/anonymized-location/README.md
```diff
@@ -0,0 +1,33 @@
...
+- Examples: [Collection](examples/collection.json), [Item](examples/item.json)
+- [JSON Schema](json-schema/schema.json)
+
+## Item and Collection Fields
+
+For Items, the fields are placed in the `properties`. For Collections, the fields are placed on the top level of the Collection.
+
+| Field Name       | Type                 | Description |
+| ---------------- | -------------------- | ----------- |
+| anon:precision   | number               | The maximum precision of the latitude and longitude coordinates present in the bbox and geometry properties. |
```
cholmes, line 27: Should probably mark this as required. Also I think it'd be good to have a bit more detail of what the number actually means. 1 means 1 degree square grid? 2 means 2 degrees? How does that relate to the number of significant digits? Like if my box is 148.25, 59.25, 152.25, 60.25 is that '1' because it's a 1x1 sized area? I think any answer is fine here, just should have more details to help people figure out the value they should put here.

Also this field should probably be marked as 'required'.

extensions/anonymized-location/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+*At least one of the fields must be specified.*
```
cholmes, line 29: This line is a bit odd if there's only one field.

extensions/anonymized-location/examples/collection.json
```diff
@@ -0,0 +1,34 @@
+{
+  "stac_version": "1.0.0-beta.1",
```
cholmes, line 2: should update to beta2

extensions/anonymized-location/examples/item.json
```diff
@@ -0,0 +1,241 @@
...
+        "anon:precision": 1
+    },
+    "assets": {
+        "ANG": {
+            "href": "https://landsat-pds.s3.amazonaws.com/c1/L8/107/018/LC08_L1TP_107018_20181001_20181001_01_RT/LC08_L1TP_107018_20181001_20181001_01_RT_ANG.txt",
+            "title": "Angle coefficients file",
+            "type": "text/plain"
+        },
+        "B1": {
+            "href": "https://landsat-pds.s3.amazonaws.com/c1/L8/107/018/LC08_L1TP_107018_20181001_20181001_01_RT/LC08_L1TP_107018_20181001_20181001_01_RT_B1.TIF",
```
cholmes, line 64: This example confuses me a bit, since landsat data from the public bucket pretty easily lets you get its location. I think it'd be better to use a more 'real' example of some image that is commercial / private.

cholmes: Like the assets linked should probably not be a geotiff, since that would have the actual location in it.

extensions/anonymized-location/examples/collection.json
m-mohr: Also, the JSON Schema should be updated to reflect recent changes we did to extension schemas.

kbgg:
> I put in a few individual comments. Overall I think the approach makes sense. Though I do worry that people may not notice that the data they're using is anonymized and may get confused. They'll need to know to look for 'anon:precision', and if they've not come across this extension before it's not obvious what that will be.
> 
> I find myself wanting some warning directly in the item / collection. Like a requirement in the description that says something like 'the geometry is not accurate since data has been anonymized.' Or a required 'anon:warning' field that has a value "this data has been anonymized - the geometry is only an approximation of the exact location'. Another route could be to only use the bbox and not the geometry, now that we can define null geometries. Or I'm open to other possibilities, but I think we have to give some clearer sign to people who may just be searching for general STAC data and be confused by it being not the actual footprint.

Thank you for all your feedback!

I've update the anon:precision field to anon:size and added a new description to better reflect what the field is supposed to contain. I've also added an anon:warning field which should let users know that the location is only approximate. Examples are also updated to be a sample of the dataset we intend to use this extension with.

matthewhanson:
As extensions now live in the stac-extensions organization, this PR is being closed. Follow instructions here to create a repository for the extension:
https://github.com/stac-extensions/stac-extensions.github.io#adding-a-new-extension

PR #854, opened 2020-06-23
Null geometry clarification

cholmes:
**Related Issue(s):** #843 

**Proposed Changes:**

1. Tweaked item spec to not require bbox if geometry is null
2. Updated best practices to explain cases when this should be used.

**PR Checklist:**


cholmes:
@m-mohr - could you update the schema to make the bbox not required if geometry is null? And a review would be great.

m-mohr:
Should we also add an example? And may it only be for checking that the JSON Schema works...

best-practices.md
```diff
@@ -5,6 +5,7 @@
 * [Field and ID formatting](#field-and-id-formatting)
 * [Field selection and Metadata Linking](#field-selection-and-metadata-linking)
 * [Datetime selection](#datetime-selection)
+* [Null Geometries](#null-geometries)
```
m-mohr, line 8: I think this section name is a bit technical. Would think "Unlocated Items" or so is more intuitive.

best-practices.md
```diff
@@ -69,6 +70,46 @@ might choose to have `datetime` be the start. The key is to put in a date and ti
...
+and use the same broad bounds for the geometry (or leaving it null) until there is precise ground lock. This estimation is 
+usually done by onboard equipment, like GPS or star trackers, but can be off by kilometers or more. But it is very useful for 
+STAC users to be able to at least find approximate area in their searches. A commonly used field for communicating ground lock 
+is not yet established, but likely should be (an extension proposal would be appreciated).  If there is no way to provide an 
+estimate then the data then a null geometry with no `bbox` can be used, as described above. But the data will likely not
+show up in STAC API searches, as most will at least implicitly use a geometry. Though this section is written with 
+satellite data in mind, one can easily imagine other data types that start with a less precise geometry but have it 
+refined after processing.
+
+### Non-STAC Items
```
m-mohr, line 103: I don't like this section. I think this section should not be in the 1.0 spec as it's more a roadmap thing.

item-spec/item-spec.md
```diff
@@ -34,7 +34,7 @@ inherited from GeoJSON.
...
 | geometry   | [GeoJSON Geometry Object](https://tools.ietf.org/html/rfc7946#section-3.1) | **REQUIRED.** Defines the full footprint of the asset represented by this item, formatted according to [RFC 7946, section 3.1](https://tools.ietf.org/html/rfc7946#section-3.1). The footprint should be the default GeoJSON geometry, though additional geometries can be included. Coordinates are specified in Longitude/Latitude or Longitude/Latitude/Elevation based on [WGS 84](http://www.opengis.net/def/crs/OGC/1.3/CRS84). |
-| bbox       | \[number]                                                                  | **REQUIRED.** Bounding Box of the asset represented by this item, formatted according to [RFC 7946, section 5](https://tools.ietf.org/html/rfc7946#section-5). |
+| bbox       | \[number]                                                                  | **REQUIRED\*.** Bounding Box of the asset represented by this item, formatted according to [RFC 7946, section 5](https://tools.ietf.org/html/rfc7946#section-5). \* Note that GeoJSON allows null geometries - if a null geometry is used then the bbox is not required. |
```
m-mohr, line 37:
```suggestion
| bbox       | \[number]                                                                  | **REQUIRED if `geometry` is not `null`.** Bounding Box of the asset represented by this item, formatted according to [RFC 7946, section 5](https://tools.ietf.org/html/rfc7946#section-5). |
```

best-practices.md
cholmes: Good call, will change.

best-practices.md
cholmes: I'd like a section to indicate that this use case isn't supported by STAC. 

But I can just say that, instead of hypothetical roadmap stuff.

m-mohr: Yeah, a short version is okay, I think. Could also refer to collection assets, which I think is the best place to put those "non STAC things"

item-spec/item-spec.md
```diff
@@ -34,7 +34,7 @@ inherited from GeoJSON.
 | id         | string                                                                     | **REQUIRED.** Provider identifier. As most geospatial assets are already defined by some identification scheme by the data provider it is recommended to simply use that ID. Data providers are advised to include sufficient information to make their IDs globally unique, including things like unique satellite IDs. |
 | type       | string                                                                     | **REQUIRED.** Type of the GeoJSON Object. MUST be set to `Feature`. |
 | geometry   | [GeoJSON Geometry Object](https://tools.ietf.org/html/rfc7946#section-3.1) | **REQUIRED.** Defines the full footprint of the asset represented by this item, formatted according to [RFC 7946, section 3.1](https://tools.ietf.org/html/rfc7946#section-3.1). The footprint should be the default GeoJSON geometry, though additional geometries can be included. Coordinates are specified in Longitude/Latitude or Longitude/Latitude/Elevation based on [WGS 84](http://www.opengis.net/def/crs/OGC/1.3/CRS84). |
```
m-mohr, line 36:
```suggestion
| geometry   | [GeoJSON Geometry Object](https://tools.ietf.org/html/rfc7946#section-3.1) \| [null](https://tools.ietf.org/html/rfc7946#section-3.2) | **REQUIRED.** Defines the full footprint of the asset represented by this item, formatted according to [RFC 7946, section 3.1](https://tools.ietf.org/html/rfc7946#section-3.1). The footprint should be the default GeoJSON geometry, though additional geometries can be included. Coordinates are specified in Longitude/Latitude or Longitude/Latitude/Elevation based on [WGS 84](http://www.opengis.net/def/crs/OGC/1.3/CRS84). |
```

best-practices.md
cholmes: Ah yeah, forgot about that. Will mention it.

cholmes: Ok, got it in.

cholmes:
As for an example, I'm of two minds. 1) be able to make sure the schema works is definitely a good thing. 2) don't really want to highlight this so much. And the plan is to slim down our examples more, so if one of just a handful of examples has a null then it would almost draw attention to it. 

If you want to I'd say go for it, as I'm conflicted on it.

m-mohr:
Let's see whether someone provides a good example: https://github.com/radiantearth/stac-spec/issues/843#issuecomment-648374035

m-mohr:
I added the JSON Schema changes.

Issue #391, opened 2019-02-05
Add JSON Schema for Point Cloud extension


m-mohr:
A JSON schema for the Point Cloud extension is missing and should be added.

adamsteer:
I’m having a go - but feel free to get in first :D

m-mohr:
We forgot to close this. There's now a JSON Schema for the PC extension.

PR #847, opened 2020-06-15
JSON Schemas, Validation and CI

m-mohr:
**Related Issue(s):** #835

**Proposed Changes:**

1. Add unique/resolvable JSON Schema ids
2. Allow schema urls in stac_extenions
3. Remove validation instructions
4. Adapt CI to host JSON Schemas

Open questions:
1. ~Schemas for the dev branch only work when we replace all version numbers in the schemas with dev and then on each release replace dev with the version number again. Is that worth the hassle?~ Yes, we do replace dev with the version number, added to process.md
2. ~Is it confusing that I removed json-schema from the paths? Feels a bit cleaner.~ added json-schema back to the path
3. ~Due to 2, the $refs are now all absolute instead of relative. Is that okay?~ relative again
4. ~It seems I couldn't make the deploy key working properly. Can anyone assist? It works locally with username and password...~ Solved
5. ~Why is the CI job running for PRs?~ Solved

~ToDo for @cholmes: Setup schemas.stacspec.org and add the CNAME file to the repo.~

**PR Checklist:**

- [x] This PR is made against the dev branch (all proposed changes except releases should be against dev, not master).
- [x] This PR has **no** breaking changes.
- [x] I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-spec/blob/dev/CHANGELOG.md) **or** a CHANGELOG entry is not required.
- [ ] This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec), and I have opened issue/PR #XXX to track the change.

PR #861, opened 2020-06-27
adding JSON schema for pointcloud extension

mccarthyryanc:
Noticed there hadn't been any updates on this for a few months and thought I would give it a try (Sorry @adamsteer. If you already have something put together, I'm happy removing this PR.) I'm new to JSON schema so happy to get feedback.

One thing that I wasn't quite sure about was adding Property Dependencies on Statistics. So if a `stats` object is given, there should also be a `schema` object with the same name.

**Related Issue(s):** #391 

**Proposed Changes:**

1. Adding schema file for pointcloud extension.

**PR Checklist:**

### DIFF
--- a/extensions/pointcloud/json-schema/schema.json
+++ b/extensions/pointcloud/json-schema/schema.json
@@ -1,0 +1,82 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://schemas.stacspec.org/dev/extensions/pointcloud/json-schema/schema.json#",
+  "title": "Point Cloud Extension",
+  "description": "STAC Point Cloud Extension to a STAC Item",
+  "allOf": [
+    {
+      "$ref": "../../../item-spec/json-schema/item.json"
+    },
+    {
+      "$ref": "#/definitions/stac_extensions"
+    },
+    {
+      "$ref": "#/definitions/pointcloud"
+    }
+  ],
+  "definitions": {
+    "pointcloud": {
+      "type": "object",
+      "required": [
+        "stac_extensions",
+        "properties"
+      ],
+      "properties": {
+        "stac_extensions": {
+          "type": "array",
+          "contains": {
+            "const": "pointcloud"
+          }
+        },
+        "properties": {
+          "type": "object",
+          "required": [
+            "pc:count",
+            "pc:type",
+            "pc:encoding",
+            "pc:schemas"
+          ],
+          "properties": {
+            "pc:count": {
+              "type": "number",
+              "multipleOf": 1.0
+            },
+            "pc:type": { "type": "string" },
+            "pc:encoding": { "type": "string" },
+            "pc:schemas": {
+              "type": "array",
+              "items": { "$ref": "#/definitions/schema" }
+            },
+            "pc:density": { "type": "number"},
+            "pc:statistics": {
+              "type": "array",
+              "items": { "$ref": "#/definitions/stats" }
+            }
+          }
+        }
+      }
+    },
+    "schema":{
+      "type": "object",
+      "properties": {
+        "name": {"type": "string" },
+        "size": {"type": "number", "multipleOf": 1.0 },
+        "type": {"type": "string" },
+        "required": ["name", "size", "type"]
+      }
+    },
+    "stats":{
+      "type": "object",
+      "properties": {
+        "average":  {"type": "number"},
+        "count":    {"type": "number", "multipleOf": 1.0 },
+        "maximum":  {"type": "number"},
+        "minimum":  {"type": "number"},
+        "name":     {"type": "string"},
+        "position": {"type": "number", "multipleOf": 1.0 },
+        "stddev":   {"type": "number"},
+        "variance": {"type": "number"}
+      }
+    }
+  }
+}
--- a/extensions/pointcloud/json-schema/schema.json
+++ b/extensions/pointcloud/json-schema/schema.json
@@ -5,7 +5,7 @@
   "description": "STAC Point Cloud Extension to a STAC Item",
   "allOf": [
     {
-      "$ref": "#/definitions/stac_extensions"
+      "$ref": "../../../item-spec/json-schema/item.json"
     },
     {
       "$ref": "#/definitions/pointcloud"
--- a/extensions/pointcloud/json-schema/schema.json
+++ b/extensions/pointcloud/json-schema/schema.json
@@ -5,9 +5,6 @@
   "description": "STAC Point Cloud Extension to a STAC Item",
   "allOf": [
     {
-      "$ref": "../../../item-spec/json-schema/item.json"
-    },
-    {
       "$ref": "#/definitions/stac_extensions"
     },
     {
@@ -25,7 +22,10 @@
         "stac_extensions": {
           "type": "array",
           "contains": {
-            "const": "pointcloud"
+            "enum": [
+              "pointcloud",
+              "https://schemas.stacspec.org/dev/extensions/pointcloud/json-schema/schema.json"
+            ]
           }
         },
         "properties": {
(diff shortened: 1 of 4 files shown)


adamsteer:
No worries @mccarthyryanc - my time to work on this has been less than zero, so it’s great to see someone else jump in!

package.json
```diff
@@ -14,6 +14,7 @@
     "remark-preset-lint-consistent": "^3.0.0",
     "remark-preset-lint-markdown-style-guide": "^3.0.0",
     "remark-preset-lint-recommended": "^4.0.0",
-    "remark-validate-links": "^10.0.0"
+    "remark-validate-links": "^10.0.0",
+    "stac-node-validator": "^0.2.1"
```
m-mohr, line 18: Please remove this line, not sure how that got in here...

extensions/pointcloud/json-schema/schema.json
```diff
@@ -0,0 +1,82 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://schemas.stacspec.org/dev/extensions/pointcloud/json-schema/schema.json#",
+  "title": "Point Cloud Extension",
+  "description": "STAC Point Cloud Extension to a STAC Item",
+  "allOf": [
+    {
+      "$ref": "../../../item-spec/json-schema/item.json"
+    },
+    {
```
m-mohr, line 10: This $ref Object is not defined anywhere and should be removed.

extensions/pointcloud/json-schema/schema.json
```diff
@@ -0,0 +1,82 @@
...
+      "type": "object",
+      "required": [
+        "stac_extensions",
+        "properties"
+      ],
+      "properties": {
+        "stac_extensions": {
+          "type": "array",
+          "contains": {
+            "const": "pointcloud"
```
m-mohr, line 28: To be aligned with #847, please change to:

```suggestion
            "enum": [
              "pointcloud",
              "https://schemas.stacspec.org/dev/extensions/pointcloud/json-schema/schema.json"
            ]
```

extensions/pointcloud/json-schema/schema.json
```diff
@@ -0,0 +1,82 @@
...
+              "type": "number",
+              "multipleOf": 1.0
+            },
+            "pc:type": { "type": "string" },
+            "pc:encoding": { "type": "string" },
+            "pc:schemas": {
+              "type": "array",
+              "items": { "$ref": "#/definitions/schema" }
+            },
+            "pc:density": { "type": "number"},
```
m-mohr, line 50: I guess we could add a `minimum: 0` here, right?

extensions/pointcloud/json-schema/schema.json
```diff
@@ -0,0 +1,82 @@
...
+          "type": "object",
+          "required": [
+            "pc:count",
+            "pc:type",
+            "pc:encoding",
+            "pc:schemas"
+          ],
+          "properties": {
+            "pc:count": {
+              "type": "number",
```
m-mohr, line 41: These two lines can be simplified to
```
              "type": "integer",
```

extensions/pointcloud/json-schema/schema.json
```diff
@@ -0,0 +1,82 @@
...
+              "multipleOf": 1.0
+            },
+            "pc:type": { "type": "string" },
+            "pc:encoding": { "type": "string" },
```
m-mohr, line 45: Not PR related, but just wondering what this field is about. Is this a media/mime type and better be part of the assets?

extensions/pointcloud/json-schema/schema.json
```diff
@@ -0,0 +1,82 @@
...
+            "pc:schemas": {
+              "type": "array",
+              "items": { "$ref": "#/definitions/schema" }
```
m-mohr, line 48: Not clear from the docs, but a required array should usually also just have at least one element (i.e. `minItems: 1`. Is this the case here and should that be clarified in the text, too?

extensions/pointcloud/json-schema/schema.json
```diff
@@ -0,0 +1,82 @@
...
+        }
+      }
+    },
+    "schema":{
+      "type": "object",
+      "properties": {
+        "name": {"type": "string" },
+        "size": {"type": "number", "multipleOf": 1.0 },
+        "type": {"type": "string" },
+        "required": ["name", "size", "type"]
```
m-mohr, line 65: "required" is on the wrong level and needs to be moved up to the "properties"/"type" fields.

extensions/pointcloud/json-schema/schema.json
```diff
@@ -0,0 +1,82 @@
...
+            }
+          }
+        }
+      }
+    },
+    "schema":{
+      "type": "object",
+      "properties": {
+        "name": {"type": "string" },
+        "size": {"type": "number", "multipleOf": 1.0 },
```
m-mohr, line 63:
```suggestion
        "size": {"type": "integer" },
```

extensions/pointcloud/json-schema/schema.json
```diff
@@ -0,0 +1,82 @@
...
+        "type": {"type": "string" },
```
m-mohr, line 64: Are "floating", "unsigned" and "signed" they only valid values?

extensions/pointcloud/json-schema/schema.json
```diff
@@ -0,0 +1,82 @@
...
+        "required": ["name", "size", "type"]
+      }
+    },
+    "stats":{
+      "type": "object",
```
m-mohr, line 69: Not really written down in the extension, but I think a "minProperties: 1" should be added here and in the docs. Having an empty object doesn't make much sense, I guess.

By the way, how does the mapping work? Is it on a name basis or on a array index basis? Should be clarified... If it's on a name basis, the field should be required.

extensions/pointcloud/json-schema/schema.json
```diff
@@ -0,0 +1,82 @@
...
+      "properties": {
+        "average":  {"type": "number"},
+        "count":    {"type": "number", "multipleOf": 1.0 },
```
m-mohr, line 72:
```suggestion
        "count":    {"type": "integer" },
```

extensions/pointcloud/json-schema/schema.json
```diff
@@ -0,0 +1,82 @@
...
+        "maximum":  {"type": "number"},
+        "minimum":  {"type": "number"},
+        "name":     {"type": "string"},
+        "position": {"type": "number", "multipleOf": 1.0 },
```
m-mohr, line 76:
```suggestion
        "position": {"type": "integer" },
```

extensions/pointcloud/json-schema/schema.json
mccarthyryanc: I was originally using `integer` but the json-schema [docs](http://json-schema.org/understanding-json-schema/reference/numeric.html#id4) had a warning about differences in validators and used `multipleOf` as a work around.

extensions/pointcloud/json-schema/schema.json
mccarthyryanc: I think `pc:type` is meant to specify how the pointcloud was generated: eopc, lidar, radar, or sonar. And `pc:encoding` would be the format: `LASzip`, `pbf`, `ept`, etc. However, the readme and `pdal-to-stac.py` differ slightly (python script only sets whether compression was used).

Based on the `example-autzen,json` I don't think these were meant to be assets. Perhaps renaming them:
`pc:type` -> `pc:device`
`pc:encoding` -> `pc:driver`
would avoid confusion?

extensions/pointcloud/json-schema/schema.json
mccarthyryanc: Yes, I'll add an `enum` here.

extensions/pointcloud/json-schema/schema.json
m-mohr: Never heard of these limitations before and we are always using integer here, I'm not aware of any issues yet.

extensions/pointcloud/json-schema/schema.json
m-mohr: If that's the case it's only the enum values, then we should also make this more clear in the docs. There it's not clear whether its exclusive...

extensions/pointcloud/json-schema/schema.json
m-mohr: Okay, thanks. For compatibility, I'm fine with leaving it as it is and could just be improved a little in the documentation, but also not really related to this PR.

extensions/pointcloud/json-schema/schema.json
mccarthyryanc: Sounds like it only applies to python based validators. I'll change to integer.

extensions/pointcloud/json-schema/schema.json
mccarthyryanc: Required `name` and set `minProperties: 2` to require at least one statistic.

extensions/pointcloud/json-schema/schema.json
mccarthyryanc: I'll leave the renaming to another issue/PR.

extensions/pointcloud/json-schema/schema.json
m-mohr: I think
```> | type       | string  | **REQUIRED.** Dimension type. Valid values include `floating`, `unsigned`, and `signed` |```

should be:

```| type       | string  | **REQUIRED.** Dimension type. Valid values are `floating`, `unsigned`, and `signed` |```

File: extensions/pointcloud/json-schema/schema.json [hunk 14]
```diff
@@ -0,0 +1,130 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://schemas.stacspec.org/dev/extensions/pointcloud/json-schema/schema.json#",
+  "title": "Point Cloud Extension",
+  "description": "STAC Point Cloud Extension to a STAC Item",
+  "allOf": [
+    {
+      "$ref": "#/definitions/stac_extensions"
```
m-mohr, line 8: Something went wrong here. You removed the STAC Item Ref instead of the stac_extensions Ref.

PR #872, opened 2020-07-17
Remove duplicate "collection" from example

blacha:
While technically valid JSON, having two "collection" keys generally means the first value is ignored, and is likely to cause confusion.

**Related Issue(s):** #
No related issue, figured a PR would be better than a issue

**Proposed Changes:**

1. Removes a duplicate "collection" key from a example

**PR Checklist:**

SInce this was a minor change, I have not done the following:


cholmes:
Thanks for the contribution!

PR #878, opened 2020-08-17
Including gsd field for assets with resolution distinct from the valu…

fredliporace:
…e defined at properties level.

**Related Issue(s):** #NA

**Proposed Changes:**

1. Including gsd value back in S2 example for assets with resolution distinct from the 10m value defined at properties. 

**PR Checklist:**

- [x] This PR does not affect the [STAC API spec](https://github.com/radiantearth/stac-api-spec)


m-mohr:
If it's 20m for all, we could also just put gsd in properties, right? It's not wrong the way it is, just a bit verbose.

m-mohr (approved):

Issue #807, opened 2020-05-19
STAC + COG and dynamic tiling


vincentsarago:
## Min/Max zoom (high priority)
I'm trying to support as much as possible STAC in other dynamic tiling projects and IMO having the min/max zoom for asset would be really useful (right now I have to read the metadata or to get the max zoom from the `gsd` is present).

## Datatype (low priority)
Having access to the datatype of the asset would also be useful, so we know if we need to apply any kind of rescaling 

## One band per asset (medium priority) 
STAC is useful to host metadata for dataset that are stored as a `one band per file` (e.g landsat/sentinel) but from my knowledge, I think there is not a field in the specs that make it clear enough. 

I'm sorry if this has already been discussed somewhere, or if those already exists :D 

I'm happy to start a PR but I wasn't sure if this would be a new extension or some addition in other extensions.

m-mohr:
1. Could https://github.com/radiantearth/stac-spec/tree/dev/extensions/tiled-assets help or be extended?
2. Probably a (new?) extension as the datatype doesn't seem search related. I'm also interested in this field. What exactly do you mean by data type though?
3. I don't understand this enough. Isn't that what eo:bands in the assets makes clear?

vincentsarago:
Hi @m-mohr thanks for your answer. 

> 1. Could https://github.com/radiantearth/stac-spec/tree/dev/extensions/tiled-assets help or be extended?

Why not, to be honest it will make some senses because Zooms are tied to a TMS 

> 2. Probably a (new?) extension as the datatype doesn't seem search related. I'm also interested in this field. What exactly do you mean by data type though?

datatype to me is https://numpy.org/doc/stable/user/basics.types.html. Basically I'd love to know if each asset is in `byte`, `float32`.... the problem is then what convention to follow (C Type ?)

> 3. I don't understand this enough. Isn't that what eo:bands in the assets makes clear?

Sorry for not being clear, the problem I have is when I need to create tiles from an Item, it's not straight forward if I can merge assets together (if each assets are separate bands) or if each assets should be treated as separate.

In fact I got this. Because the specs are loose, you can have both `band`+`product` in the asset list. So as you mentioned, maybe really on item `eo:bands` is the best way.

geospatial-jeff:
Let me start off by saying I'm a huge fan of STAC + COG.  I think there is a lot to gain by creating interoperability between the two.

[...]

PR #879, opened 2020-08-18
Rendering Hints extension (WIP)

cholmes:
**Related Issue(s):** #807 

**Proposed Changes:**

1. Added rendering hints extension

**PR Checklist:**

### DIFF
--- a/extensions/rendering-hints/README.md
+++ b/extensions/rendering-hints/README.md
@@ -29,19 +29,22 @@ the values are assumed to apply to all non-thumbnail Assets in that Item.
 
 | Field Name       | Type                     | Description |
 | ---------------- | ------------------------ | ----------- |
-| render:max_zoom  | integer | The maximum [zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels), expressed as an integer, that corresponds with the highest resolution of the data. Specified as [Web Mercator](https://en.wikipedia.org/wiki/Web_Mercator_projection) zoom levels. |
-| render:min_zoom  | integer  | The minimum [zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels), expressed as an integer, that tells renderers where the overviews of the data stop. |
+| render:overview_max_gsd  | number | The maximum Ground Sample Distance represented in an overview. This should be the GSD of the highest level overview, generally of a [Cloud Optimized GeoTIFF](http://cogeo.org), but should work with any format. |
 | render:data_type    | string | The data `type` (float, int, complex, etc) to let the renderer apply any needed rescaling up front. The full set of options is listed below. |
 
-**render:max_zoom**: This field informs renderers of the resolution of the data, using 
-[Web Mercator](https://en.wikipedia.org/wiki/Web_Mercator_projection) zoom levels (since it is used in the majority
-of web tile maps). Some renderers will not use web mercator, but most should be able to use the hint in web mercator 
-and translate to the appropriate zoom level. The easiest way to get the zoom levels for data is to use [rio-cogeo](https://pypi.org/project/rio-cogeo/): `[get_zooms()](https://github.com/cogeotiff/rio-cogeo/blob/master/rio_cogeo/utils.py#L69)` in python or using `[rio info](https://github.com/cogeotiff/rio-cogeo/#examples)` on the command line. Should be between
-0 and 30. 
+**render:overview_max_gsd**: This field helps renderers of understand what zoom levels they can efficiently show. It is 
+generally used in conjunction with `gsd` (from [common metadata](https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#instrument)). `overview_max_gsd` enables the calculuation of the 'minimum' zoom level that a renderer
+would want to show, and then the maximum zoom level is calculated from the `gsd` - the resolution of the image. The former
+is based on the highest level of overview (also known as a [pyramid](https://en.wikipedia.org/wiki/Pyramid_(image_processing)))
+contained in the asset. 
 
-**render:min_zoom** - This field informs renderers of the minimum level supported by the data, generally corresponding
-to the resolution for which overviews were generated. It works just like max zoom: specified in web mercator and 
-rio-cogeo can be used to easily get the value. If not specified then renderers will assume the minimum zoom is 0.
+![Image Pyramid](https://en.wikipedia.org/wiki/Pyramid_(image_processing)#/media/File:Image_pyramid.svg)
+
+So in the above image it would be the ground sample distance of 'level 4', which will be a much higher gsd than the image,
+as each pixel is greatly down-sampled. Dynamic tile servers (like [titiler](https://github.com/developmentseed/titiler)) will
+generally convert the gsd to [zoom 
+levels](https://wiki.openstreetmap.org/wiki/Zoom_levels), [Web Mercator](https://en.wikipedia.org/wiki/Web_Mercator_projection) 
+or others, which is easily done (example python [to webmercator](https://github.com/cogeotiff/rio-cogeo/blob/b9b57301c2b7a4be560c887176c282e68ca63c27/rio_cogeo/utils.py#L62-L66) or arbitrary [TileMatrixSet](https://github.com/cogeotiff/rio-tiler-crs/blob/834bcf3d39cdc555b3ce930439ab186d00fd5fc5/rio_tiler_crs/cogeo.py#L98-L105))
 
 **render:data_type** - Specifies the data storage `type` of the assets. Can be used at the asset level if the assets are of
 different types. This is used to let the renderer know if any type of rescaling is needed up front. The possible values
@@ -50,9 +53,10 @@ for the type in STAC are specified in the table below.
 | Type Name | Description |
 |-----------|-------------|
 | `unknown` | Not known   |
-| `byte`    | An unsigned 8-bit integer (common for 8-bit rgb png's) |
+| `int8`    | 8-bit integer  |
 | `int16`   | 16-bit integer |
 | `int32`   | 32-bit integer |
+| `uint8`   | unsigned 8-bit integer (common for 8-bit rgb png's)
 | `unit16`  | unsigned 16-bit integer |
 | `uint32`  | unsigned 32-bit integer |
 | `float32` | 32-bit float |
--- a/extensions/rendering-hints/README.md
+++ b/extensions/rendering-hints/README.md
@@ -38,7 +38,7 @@ would want to show, and then the maximum zoom level is calculated from the `gsd`
 is based on the highest level of overview (also known as a [pyramid](https://en.wikipedia.org/wiki/Pyramid_(image_processing)))
 contained in the asset. 
 
-![Image Pyramid](https://user-images.githubusercontent.com/407017/90821250-75ce5280-e2e7-11ea-9008-6c073e083be0.png)
+<img src="https://user-images.githubusercontent.com/407017/90821250-75ce5280-e2e7-11ea-9008-6c073e083be0.png" alt="image pyramid" width="300">
 
 So in the above image it would be the ground sample distance of 'level 4', which will be a much higher gsd than the image,
 as each pixel is greatly down-sampled. Dynamic tile servers (like [titiler](https://github.com/developmentseed/titiler)) will
--- a/extensions/rendering-hints/README.md
+++ b/extensions/rendering-hints/README.md
@@ -38,7 +38,7 @@ would want to show, and then the maximum zoom level is calculated from the `gsd`
 is based on the highest level of overview (also known as a [pyramid](https://en.wikipedia.org/wiki/Pyramid_(image_processing)))
 contained in the asset. 
 
-![Image Pyramid](https://en.wikipedia.org/wiki/Pyramid_(image_processing)#/media/File:Image_pyramid.svg)
+![Image Pyramid](https://user-images.githubusercontent.com/407017/90821250-75ce5280-e2e7-11ea-9008-6c073e083be0.png)
 
 So in the above image it would be the ground sample distance of 'level 4', which will be a much higher gsd than the image,
 as each pixel is greatly down-sampled. Dynamic tile servers (like [titiler](https://github.com/developmentseed/titiler)) will
--- a/extensions/rendering-hints/README.md
+++ b/extensions/rendering-hints/README.md
@@ -1,0 +1,67 @@
+# Rendering Hints Extension Specification
+
+- **Title: Rendering Hints**
+- **Identifier: rendering_hints**
+- **Field Name Prefix: render**
+- **Scope: Item**
+- **Extension [Maturity Classification](../README.md#extension-maturity): Proposal**
+
+One of the things many people want to do with data cataloged in STAC is to render it, often doing so directly
+from the assets (usually [Cloud Optimized GeoTIFF's](http://cogeo.org)) with dynamic web tile services. To 
+perform well these services benefit from a few 'hints' that can help the renderer perform more effectively. 
+
+This document explains the fields of the STAC Rendering Hints (`render`) Extension to a STAC Item. They are meant
+to be 'hints' - things that could be discovered by inspecting the data, but provided in STAC to help out tools
+that can take advantage of them. It is written with geospatial raster data in mind, but could apply to other data types.
+
+The field names defined herein should be added as fields in the Item Properties object. When specified on an Item, 
+the values are assumed to apply to all non-thumbnail Assets in that Item. 
+
+## Examples
+
+- TODO
+
+## Schema
+
+- TODO
+
+## Item Properties fields
+
+| Field Name       | Type                     | Description |
+| ---------------- | ------------------------ | ----------- |
+| render:max_zoom  | integer | **REQUIRED.** The maximum [zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels), expressed as an integer, that corresponds with the highest resolution of the data. Specified as
+[Web Mercator](https://en.wikipedia.org/wiki/Web_Mercator_projection) zoom levels.|
+| render:min_zoom  | integer  | The minimum [zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels), expressed as an integer, that tells renderers where the overviews of the data stop. |
+| render:type    | string | The data `type` (float, int, complex, etc) to let the renderer apply any needed rescaling up front. The full set of options is listed below. |
+
+**render:max_zoom**: This field informs renderers of the resolution of the data, using 
+[Web Mercator](https://en.wikipedia.org/wiki/Web_Mercator_projection) zoom levels (since it is used in the majority
+of web tile maps). Some renderers will not use web mercator, but most should be able to use the hint in web mercator 
+and translate to the appropriate zoom level. The easiest way to get the zoom levels for data is to use [rio-cogeo](https://pypi.org/project/rio-cogeo/): `[get_zooms()](https://github.com/cogeotiff/rio-cogeo/blob/master/rio_cogeo/utils.py#L69)` in python or using `[rio info](https://github.com/cogeotiff/rio-cogeo/#examples)` on the command line. Generally will
+be between 
+
+**render:min_zoom** - This field informs renderers of the minimum level supported by the data, generally corresponding
+to the resolution for which overviews were generated. It works just like max zoom: specified in web mercator and 
+rio-cogeo can be used to easily get the value. If not specified then renderers will assume the minimum zoom is 0.
+
+**render:type** - Specifies the data storage `type` of the assets. Can be used at the asset level if the assets are of
+different types. This is used to let the renderer know if any type of rescaling is needed up front. The possible values
+for the type in STAC are specified in the table below.
+
+| Type Name | Description |
+|-----------|-------------|
+| `unknown` | Not known   |
+| `byte`    | An unsigned 8-bit integer (common for 8-bit rgb png's) |
+| `int16`   | 16-bit integer |
+| `int32`   | 32-bit integer |
+| `unit16`  | unsigned 16-bit integer |
+| `uint32`  | unsigned 32-bit integer |
+| `float32` | 32-bit float |
+| `float64` | 64-big float |
+| `cint16`  | 16-bit complex integer |
+| `cint32`  | 32-bit complex integer |
+| `cfloat32` | 32-bit complex float |
+| `cfloat64` | 64-bit complex float |
+
+
+
--- a/extensions/rendering-hints/README.md
+++ b/extensions/rendering-hints/README.md
@@ -29,21 +29,21 @@ the values are assumed to apply to all non-thumbnail Assets in that Item.
 
 | Field Name       | Type                     | Description |
 | ---------------- | ------------------------ | ----------- |
-| render:max_zoom  | integer | **REQUIRED.** The maximum [zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels), expressed as an integer, that corresponds with the highest resolution of the data. Specified as [Web Mercator](https://en.wikipedia.org/wiki/Web_Mercator_projection) zoom levels. |
+| render:max_zoom  | integer | The maximum [zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels), expressed as an integer, that corresponds with the highest resolution of the data. Specified as [Web Mercator](https://en.wikipedia.org/wiki/Web_Mercator_projection) zoom levels. |
 | render:min_zoom  | integer  | The minimum [zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels), expressed as an integer, that tells renderers where the overviews of the data stop. |
-| render:type    | string | The data `type` (float, int, complex, etc) to let the renderer apply any needed rescaling up front. The full set of options is listed below. |
+| render:data_type    | string | The data `type` (float, int, complex, etc) to let the renderer apply any needed rescaling up front. The full set of options is listed below. |
 
 **render:max_zoom**: This field informs renderers of the resolution of the data, using 
 [Web Mercator](https://en.wikipedia.org/wiki/Web_Mercator_projection) zoom levels (since it is used in the majority
 of web tile maps). Some renderers will not use web mercator, but most should be able to use the hint in web mercator 
-and translate to the appropriate zoom level. The easiest way to get the zoom levels for data is to use [rio-cogeo](https://pypi.org/project/rio-cogeo/): `[get_zooms()](https://github.com/cogeotiff/rio-cogeo/blob/master/rio_cogeo/utils.py#L69)` in python or using `[rio info](https://github.com/cogeotiff/rio-cogeo/#examples)` on the command line. Generally will
-be between 
+and translate to the appropriate zoom level. The easiest way to get the zoom levels for data is to use [rio-cogeo](https://pypi.org/project/rio-cogeo/): `[get_zooms()](https://github.com/cogeotiff/rio-cogeo/blob/master/rio_cogeo/utils.py#L69)` in python or using `[rio info](https://github.com/cogeotiff/rio-cogeo/#examples)` on the command line. Should be between
+0 and 30. 
 
 **render:min_zoom** - This field informs renderers of the minimum level supported by the data, generally corresponding
 to the resolution for which overviews were generated. It works just like max zoom: specified in web mercator and 
 rio-cogeo can be used to easily get the value. If not specified then renderers will assume the minimum zoom is 0.
 
-**render:type** - Specifies the data storage `type` of the assets. Can be used at the asset level if the assets are of
+**render:data_type** - Specifies the data storage `type` of the assets. Can be used at the asset level if the assets are of
 different types. This is used to let the renderer know if any type of rescaling is needed up front. The possible values
 for the type in STAC are specified in the table below.
 
--- a/extensions/rendering-hints/README.md
+++ b/extensions/rendering-hints/README.md
@@ -56,7 +56,7 @@ for the type in STAC are specified in the table below.
 | `int8`    | 8-bit integer  |
 | `int16`   | 16-bit integer |
 | `int32`   | 32-bit integer |
-| `uint8`   | unsigned 8-bit integer (common for 8-bit rgb png's)
+| `uint8`   | unsigned 8-bit integer (common for 8-bit rgb png's) |
 | `unit16`  | unsigned 16-bit integer |
 | `uint32`  | unsigned 32-bit integer |
 | `float32` | 32-bit float |
--- a/extensions/rendering-hints/README.md
+++ b/extensions/rendering-hints/README.md
@@ -29,8 +29,7 @@ the values are assumed to apply to all non-thumbnail Assets in that Item.
 
 | Field Name       | Type                     | Description |
 | ---------------- | ------------------------ | ----------- |
-| render:max_zoom  | integer | **REQUIRED.** The maximum [zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels), expressed as an integer, that corresponds with the highest resolution of the data. Specified as
-[Web Mercator](https://en.wikipedia.org/wiki/Web_Mercator_projection) zoom levels.|
+| render:max_zoom  | integer | **REQUIRED.** The maximum [zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels), expressed as an integer, that corresponds with the highest resolution of the data. Specified as [Web Mercator](https://en.wikipedia.org/wiki/Web_Mercator_projection) zoom levels. |
 | render:min_zoom  | integer  | The minimum [zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels), expressed as an integer, that tells renderers where the overviews of the data stop. |
 | render:type    | string | The data `type` (float, int, complex, etc) to let the renderer apply any needed rescaling up front. The full set of options is listed below. |
 
@@ -62,6 +61,3 @@ for the type in STAC are specified in the table below.
 | `cint32`  | 32-bit complex integer |
 | `cfloat32` | 32-bit complex float |
 | `cfloat64` | 64-bit complex float |
-
-
-
--- a/extensions/rendering-hints/README.md
+++ b/extensions/rendering-hints/README.md
@@ -1,7 +1,7 @@
 # Rendering Hints Extension Specification
 
 - **Title: Rendering Hints**
-- **Identifier: rendering_hints**
+- **Identifier: rendering-hints**
 - **Field Name Prefix: render**
 - **Scope: Item**
 - **Extension [Maturity Classification](../README.md#extension-maturity): Proposal**


cholmes:
Ok, first draft is up. Detailed review is appreciated since I am not sure that I explained everything right. Any more details on the 'why' would also be good to add.

Still need to make an example and provide a schema. But feedback on the core ideas would be great. 

Also am thinking about making the schema check zoom levels between 0 and 25? I welcome ideas on what to put for the max, but would be good to check to make sure users aren't putting in like 500 for the zoom levels.

extensions/rendering-hints/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+- TODO
+
+## Item Properties fields
+
+| Field Name       | Type                     | Description |
+| ---------------- | ------------------------ | ----------- |
+| render:max_zoom  | integer | **REQUIRED.** The maximum [zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels), expressed as an integer, that corresponds with the highest resolution of the data. Specified as [Web Mercator](https://en.wikipedia.org/wiki/Web_Mercator_projection) zoom levels. |
+| render:min_zoom  | integer  | The minimum [zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels), expressed as an integer, that tells renderers where the overviews of the data stop. |
+| render:type    | string | The data `type` (float, int, complex, etc) to let the renderer apply any needed rescaling up front. The full set of options is listed below. |
```
geospatial-jeff, line 34: I think `dtype` would be more explicit.

extensions/rendering-hints/README.md
```diff
@@ -0,0 +1,63 @@
...
+rio-cogeo can be used to easily get the value. If not specified then renderers will assume the minimum zoom is 0.
+
+**render:type** - Specifies the data storage `type` of the assets. Can be used at the asset level if the assets are of
+different types. This is used to let the renderer know if any type of rescaling is needed up front. The possible values
+for the type in STAC are specified in the table below.
+
+| Type Name | Description |
+|-----------|-------------|
+| `unknown` | Not known   |
+| `byte`    | An unsigned 8-bit integer (common for 8-bit rgb png's) |
```
jisantuc, line 53: as far as type names go, `int8` is numpy's reference to the type:

```
In [1]: import numpy as np

In [2]: np.int8
Out[2]: numpy.int8

In [3]: np.byte
Out[3]: numpy.int8
```

And `rasterio` has a similar idea:

```
In [1]: import rasterio

In [2]: src = rasterio.open("<a tiff i have>")

In [3]: src.dtypes
Out[3]: ('uint8', 'uint8', 'uint8')
```

I don't think it's super important but it's not an uncommon name for the type

extensions/rendering-hints/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+- TODO
+
+## Schema
+
+- TODO
+
+## Item Properties fields
+
+| Field Name       | Type                     | Description |
```
jisantuc, line 30: (throwing this here because it seems like the most appropriate place to ask about included fields)

I don't know what the appropriate scope is for this extension, but I think the main hint I would want, if it's available, is a colormap from the tif. I'm a little bit confused about what I'd do with the min and max zoom hints.

extensions/rendering-hints/README.md
jisantuc: Oh no I scrolled up in gitter and now see the numpy / rasterio discussion 🙃

extensions/rendering-hints/README.md
cholmes: I might go full-on data_type in that case (as I'm guessing that's what the d is, but not sure, which is why I'd like to make it explicit). I was going back and forth between the two.

extensions/rendering-hints/README.md
cholmes: I've been interested in having that type of stuff in here, though I'm far from an expert on it. I think it can be in scope, and I was anticipating it'd grow to include similar things. But I'm more than happy to include it now. Just give me more details of what that actually looks like - what is the field, what's in it, how do you explain it. Or (preferred) feel free to add a 'commitable suggestion' (or whatever it is called) on the PR and I'll add it in.

extensions/rendering-hints/README.md
cholmes: I really know very little about these things, so I'm happy for whatever you all think is best. I just want to have a clear list for people to pick from.

kylebarron: Oh sorry; I thought `byte` was an alias for `uint8`, not `int8`. I see Numpy has `ubyte` as the alias for `uint8`. `uint8` is the more common one for imagery as you saw

extensions/rendering-hints/README.md
kylebarron: `dtype` is pretty common in the Python world (that's the term Numpy and Pandas use) but I think `data_type` would be most clear to non-Python users.

kylebarron:
> Also am thinking about making the schema check zoom levels between 0 and 25?

@vincentsarago uses 30 as the max zoom in his mosaic code. 
https://github.com/developmentseed/cogeo-mosaic/blob/157d31972398d8f1d8e0a13e8290e8ddb6f1a33a/cogeo_mosaic/mosaic.py#L61-L62

extensions/rendering-hints/README.md
```diff
@@ -0,0 +1,63 @@
...
+One of the things many people want to do with data cataloged in STAC is to render it, often doing so directly
+from the assets (usually [Cloud Optimized GeoTIFF's](http://cogeo.org)) with dynamic web tile services. To 
+perform well these services benefit from a few 'hints' that can help the renderer perform more effectively. 
+
+This document explains the fields of the STAC Rendering Hints (`render`) Extension to a STAC Item. They are meant
+to be 'hints' - things that could be discovered by inspecting the data, but provided in STAC to help out tools
+that can take advantage of them. It is written with geospatial raster data in mind, but could apply to other data types.
+
+The field names defined herein should be added as fields in the Item Properties object. When specified on an Item, 
+the values are assumed to apply to all non-thumbnail Assets in that Item.
```
kylebarron, line 18: > the values are assumed to apply to all non-thumbnail Assets in that Item

I think this is ok, but for e.g. Sentinel 2 with varied-resolution bands, some bands will have a lower `max_zoom`

extensions/rendering-hints/README.md
```diff
@@ -0,0 +1,63 @@
...
+## Schema
+
+- TODO
+
+## Item Properties fields
+
+| Field Name       | Type                     | Description |
+| ---------------- | ------------------------ | ----------- |
+| render:max_zoom  | integer | **REQUIRED.** The maximum [zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels), expressed as an integer, that corresponds with the highest resolution of the data. Specified as [Web Mercator](https://en.wikipedia.org/wiki/Web_Mercator_projection) zoom levels. |
+| render:min_zoom  | integer  | The minimum [zoom level](https://wiki.openstreetmap.org/wiki/Zoom_levels), expressed as an integer, that tells renderers where the overviews of the data stop. |
```
kylebarron, line 33: I think it would be good to specify whether `min_zoom` and `max_zoom` are inclusive of the given integer.

I think it would be good to define more clearly how `min_zoom` and `max_zoom` are defined:

- At least in the sense of COGs where you're generally creating a raster, what pixel resolution underlies these min/max zoom estimations? Most raster map data is served as either 256x256px tiles or 512x512px tiles. That difference changes the `zoom` level by 1.
- Specify that these are generated by considering the pixel resolution per meter and that for Web Mercator this is technically only valid at the equator.
- If min/max zoom must be an integer, how does rounding happen? Is `max_zoom` the integer zoom beyond which you can't get any more resolution and `min_zoom` the integer zoom before which you can't get any less resolution from overviews? So if these correspond to fractional zooms of 7.5 to 10.5, the min/max zooms would be stored as 7 and 11, respectively? I'm guessing that storing these numbers as integers will be more accessible to more people.

kylebarron: - If min/max zooms are a single number for all bands, does the number represent the most detailed band?
- If a less-detailed band had the same number of overviews as a more-detailed band, the less-detailed band would have a smaller `min_zoom` than the more-detailed band

extensions/rendering-hints/README.md
kylebarron: > I'm a little bit confused about what I'd do with the min and max zoom hints

Min and max zoom hints are helpful for on-demand rendering. The `max_zoom` is generally discoverable from the `gsd` separately.

That gets me thinking... what if a different way to describe min/max zoom is the `gsd` of the full-resolution data vs the `gsd` of the smallest overview? the main `gsd` is of course already stored per band. Storing the `min_overview_gsd` [of each band?] would allow the dynamic tiler to derive the min/max zoom levels, and would additionally support non-mercator rendering by being able to derive those zoom levels from the `gsd` values

m-mohr:
@DanielJDufour I think this one could be interesting for you and I think you have some expertise to share from your work on geotiff-stats, geotiff.io, georaster etc. Could you have a look, please?

extensions/rendering-hints/README.md
m-mohr: Additionally, for visualization it's sometimes important to know the min/max values (not zoom, actual values) for the data. Sometimes that can't be well determined from the data_type. For example, normalized_differences are usually between -1 and 1, but non of the data types really caters for that.

Additionally, I would like to use some of the fields from this extension, but don't always care about zoom levels. So having that required seems to make this less useful. But I'm not exactly sure what to require. Maybe just one of the properties? (Schema: `minProperties: 1`)

m-mohr:
@cholmes Let me know whenever you are finished and need a JSON Schema ;-)

DanielJDufour:
Hello.  I apologize if my suggestions are already covered by other extensions (I'm still learning about what is out there).  I submit the following items for your consideration:

### minimum and maximum pixel values for each band of each asset
This is useful for rendering single-band images without a color palette.  It could also help render a single-band from a Landsat scene.  I should note that GDAL can embed this information as XML into a GeoTIFF's metadata.  However, I'm not sure if you can do this for other types of gridded data (e.g., JPG2000).  Maybe someone else can provide information on that.  And JSON is easier to work with if you are doing anything client-side in the browser.

### more projection information
It's probably because of my lack of knowledge about STAC that I suggest this.  But is there currently a way to grab the projection information for an asset?  Ideally we could have the wkt and maybe proj4js string easily available.  In order to render a GeoTIFF asset in a projection other than the one it is in would require this information.  However, my understanding is that this has been solved in Python already via GDAL/RasterIO, but the JS ecosystem doesn't yet have a solution for converting a GeoTIFFs key to a proj4js or wkt string.  I started a little of this work [here](https://github.com/GeoTIFF/geokeys-to-proj4js), but there's a long way to go.  This is also currently (imho) the biggest blocker to creating an OpenLayers plugin ([ol-geotiff](https://github.com/GeoTIFF/ol-geotiff)) that can handle and display any GeoTIFF (and doesn't rely on external services).  It's also the most common issue that comes up with [GeoRasterLayer](https://github.com/geotiff/georaster-layer-for-leaflet)

### color palette
(I think someone mentioned this already).  The color palette is basically the mapping of a pixel value to an RGBA Color.  GeoTIFFs often store color palette imagery in a somewhat compressed format where you don't store all the mappings, but instead the scaled step between pixel values. [geotiff-palette](https://github.com/GeoTIFF/geotiff-palette) basically expands this info into a simple array of RGBA values where the index number in the array refers to its pixel value.  For 8-bit imagery, there would be 256 `[R,G,B,A]` arrays, but for imagery with a larger range of values (e.g. float/decimal values), you could have a very large palette, so it probably makes sense to store this information in the compressed format.  Open to correction because I don't use palettes that much.

### histograms
Similar to palettes the size of the histogram depends on the number of uniques values found in the raster.  GDAL will bin/group this data for you, which will shrink the space required to store this information.  However, it's often useful to have direct access to the raw un-binned counts for each pixel value, especially if the renderer would like to bin it with different bin sizes.  The raw data could get very large if used on raster with more than 8-bits per pixel.

### tile-level information
It would be awesome if there was a standard around storing not just file-level statistics, but the stats for each tile.  It could be helpful for customizing/stretching the rendering relative to tiles that are actually being displayed.  Sometimes people will want to apply a threshold (i.e., only show pixel values over 300), so having access to the tile-level minimum, could help us avoid extra calls for tiles without this information.  Here's an example of displaying a GeoTIFF that displays the areas where tropical fruit is grown in Puerto Rico: https://geotiff.github.io/georaster-layer-for-leaflet-example/examples/thresholding.html

### mask
This might have been mentioned earlier and I've seen it used in some of the awesome tilers out there.  It might be useful in some situations to have access to a polygon representing how the data should be masked (i.e. where the no-data values are).  It can help avoid extra calls for pixels in an area that is only no-data values.

### range of pixel values for each band
This could be optional or not included in the extension because it's basically derived by subtracting the minimum pixel value from the maximum pixel value.  However, it could be useful and remove one (albeit simple) step.  Here's an example of the range (along with min and max) being used: https://github.com/GeoTIFF/georaster-layer-for-leaflet/blob/master/georaster-layer-for-leaflet.js#L377

### official rendering functions
I'm not sure about this one, but thought I should share it in case it sparks conversation.  I see it's usefulness but also its drawbacks.  It could be interesting if there was a way that data holders can provide validated band arithmetic functions.  I've personally found it difficult to come up these equations because of all the caveats.  For example, although NDVI is a rather simple equation on the face of it, in practice, one often has to add in caveats like making sure water displays correctly.  We could also consider using a generic language like the one proposed by [stac-expr](http://github.com/geotiff/stac-expr), but I've heard its more useful for people to have code in the language they would actually run the arithmetic in, like Python or JavaScript.
  For example, we could have a JSON structure look like:
```json
"expressions": {
    "ndvi": {
        "js": "results = (nir - red) / (nir + red); return result <= 0.1 ? 'blue' : result >= 0.8 ? 'black' : result",
        "python": ...
   }
}
```
Although these expressions can get long and complicated, that's precisely why I think it could be useful for people who simply want to display NDVI without having to do the Math themselves.  However, it would increase the maintenance cost and what happens when an equation gets updated?  Would this change any applications depending on the previous rendering equation?  Is that a good or bad thing?  What do you think @m-mohr ? Would love your thoughts on this.

### Supervised Classification and AI model output
This is probably out of scope, but if we really wanted to go overboard, we could include include the results of supervised classification.  For example, we could store the range for water values:
```
"classification_results": {
    "corn": "0.4 < b01 < 0.9",
    "water": "(nir - red) / (nir + red) <= 0.1"
}
```

I'd also like to invite @rowanwins to offer some feedback.  He's done a lot of good work with GeoTIFFs and might also have some good suggestions to add :-)  He's mentioned standard deviation and quantiles being useful before ([source](https://github.com/GeoTIFF/geotiff-stats/issues/3).

Apologies again for the lengthy post.  I think there's a clear use case for min/max, projection info, color palette, histogram, tile-level info, and mask.  However, some of this might already be covered by other extensions.  I'm also unsure about whether it makes sense to include range, band arithmetic, and classification output in the STAC extension because it would increase the scope of it and might make sense to separate concerns more.

Looking forward to your thoughts and feedback.

extensions/rendering-hints/README.md
cholmes: Cool. I think the audience for this is beyond python users, so will go for data_type.

extensions/rendering-hints/README.md
cholmes: So I should change this to just be unit8 ?

extensions/rendering-hints/README.md
cholmes: min_overview_gsd does seem a bit cleaner. Is there an easy way to calculate that? Would want to give people some reference of how to get it. 

min/max values sounds good to add. I'm for it. What should the field name be? 

And `minProperties: 1` sounds good - I only knew the one use case, but if you have a use case that involves not using them then I'm happy to not make that one required.

extensions/rendering-hints/README.md
cholmes: The new asset fields should cover that - you can apply the max_zoom to the asset if you want more granularity.

cholmes:
Thanks for all the extensive comments @DanielJDufour! Really appreciate it. A few responses:

> minimum and maximum pixel values for each band of each asset

@m-mohr suggested this too. Let's get it in. `min_pixel_value` and `max_pixel_value` as two fields? And what's the range of values we should accept here, and what type? For the json schema, and to explain. 

> more projection information

https://github.com/radiantearth/stac-spec/tree/master/extensions/projection is pretty extensive - let me know if there's anything that's needed for you, but I think it should be sufficient. We also recently added transform & shape, to enable things like VRT's without having to open the files.

> color palette

This sounds like more of an asset than a property? Like it's a small file you reference (or even be embedded in the geotiff?), not something you'd embed in JSON? Or do I have it wrong? I think that'd just be an asset. I don't know that other extensions yet specify assets, but it seems like it would make sense to me.

> histogram

This also sounds interesting to me, and also sounds like more of an asset to reference? I'd love an example with a color palette and a histogram, so we can show directly what it would look like for people, and provide a best practice for this stutff.

> tile-level information

This one feels outside the scope of STAC to me? Unless maybe if you're using the [tiled asset extension](https://github.com/radiantearth/stac-spec/tree/master/extensions/tiled-assets)

> mask

Feels like another asset? Again, if someone can get me an example mask I can probably make an example with it, and recommend it.

> range of pixel values for each band

I'm inclined to just leave this for people to get by subtracting min/max. So that there's not a situation where the values don't agree with one another, like someone changes one but doesn't update the other. 

> official rendering functions

I'm into this idea in general - like getting to standard rendering functions. At Planet we're experimenting with this, and I would have loved a standard to point at. In the interest of 'small pieces loosely coupled' I'm inclined to aim for a standard on this that stands alone / isn't tied to STAC. But that could be referenced by STAC. Since I think there are people who would want to use this without having to understand STAC.

> Supervised Classification and AI model output

Similar feelings to the official rendering functions - cool to have, let's do in its own spec.

kylebarron:
A few thoughts

### Min/Max `gsd` instead of Min/Max `zoom`

After some thought and discussion with @geospatial-jeff and @vincentsarago, I'd like to propose that instead of `minzoom` and `maxzoom`, we have `maxgsd` and `mingsd`, where `gsd` already exists in STAC and is defined as "Ground Sample Distance at the sensor", and the other is the `gsd` that corresponds to the smallest overview level.

Using `gsd` instead of `zoom` is more applicable to a wider range of applications because `gsd` is essentially using the "local projection" (i.e. UTM in meters) rather than Web Mercator at the equator. This means that for applications that want to visualize the STAC item _not in Web Mercator_, `gsd` will be more helpful.

Additionally, converting `gsd` to `zoom` is very simple (5 lines of code) in a range of projections. [Web Mercator](https://github.com/cogeotiff/rio-cogeo/blob/b9b57301c2b7a4be560c887176c282e68ca63c27/rio_cogeo/utils.py#L62-L66); [Arbitrary TileMatrixSet](https://github.com/cogeotiff/rio-tiler-crs/blob/834bcf3d39cdc555b3ce930439ab186d00fd5fc5/rio_tiler_crs/cogeo.py#L98-L105).

> ### minimum and maximum pixel values for each band of each asset

If you know the `dtype` is `uint16`; then you know the max value of each band is `65535`, no? Are there real-world examples of the `dtype` set to `uint16` but not all of the range is used?

> ### more projection information

This is already covered with the [`proj` extension](https://github.com/radiantearth/stac-spec/blob/dev/extensions/projection/README.md#item-properties-fields) I believe. In Sentinel 2 COG STAC this exists per band:

```json
"proj:shape": [1830, 1830],
"proj:transform": [
  60.0,
  0.0,
  600000.0,
  0.0,
  -60.0,
  1800000.0,
  0.0,
  0.0,
  1.0
]
```

> ### mask

At least with Sentinel 2 COG STAC, the primary STAC geometry already includes only the valid portion of the image.

> ### official rendering functions
> ### Supervised Classification and AI model output

IMO a black hole and out of scope for STAC. I would argue STAC should only be for objective data and these both seem very subjective. Hence this should be a client choice. Additionally, it seems to be collection-level information that doesn't vary at the item level.

> ### tile-level information

If I understand you correctly, you're suggesting info _for each internal tile of each overview_? This seems like a ton of data and would make each STAC item _huge_. For a 640MB NAIP COG, it has 908 tiles, so storing tile-level information for each one would make the STAC json at least an order of magnitude larger. And it would make it much harder to store an entire collection of STACs with this info per tile.

kylebarron:
> > color palette
> 
> This sounds like more of an asset than a property? Like it's a small file you reference (or even be embedded in the geotiff?), not something you'd embed in JSON? Or do I have it wrong? I think that'd just be an asset. I don't know that other extensions yet specify assets, but it seems like it would make sense to me.

I think these are generally embedded within the GeoTIFF, but I could be wrong.

cholmes:
> Should we include raster statistics (min/mean/max) as optional fields in this extension?

@geospatial-jeff - I'm into this in theory, but I struggle with how to implement. I'm guessing you mean like the results of gdal with -stats:

    STATISTICS_MAXIMUM=255
    STATISTICS_MEAN=28.054506440187
    STATISTICS_MINIMUM=0
    STATISTICS_STDDEV=42.88762832552
    STATISTICS_VALID_PERCENT=100

I could see it working if all data distributed one band per asset, since then you can just have those fields on each asset. But many have one asset with several bands, so we'd need like some nested json object to give the values for each band. And that just seems to get pretty messy.

Seems like we should just get this written up as an 'extension' on the COG spec. Like it feels much cleaner at the tiff tag level. I've been meaning to do that since @mojodna came up with the idea.

cholmes:
> > ###  Min/Max gsd instead of Min/Max zoom
> After some thought and discussion with @geospatial-jeff and @vincentsarago, I'd like to propose that instead of minzoom and maxzoom, we have maxgsd and mingsd, where gsd already exists in STAC and is defined as "Ground Sample Distance at the sensor", and the other is the gsd that corresponds to the smallest overview level.

I'm on board. Seems like we just need min_gsd for this extension then? And a perhaps a recommendation on how to use gsd to calculate the max? I'll take a crack at it.

> > ### minimum and maximum pixel values for each band of each asset
> If you know the dtype is uint16; then you know the max value of each band is 65535, no? Are there real-world examples of the dtype set to uint16 but not all of the range is used?

I had thought someone mentioned this in the thread, but now I can't seem to find it. But I think the real-world example is NDVI output, that is often just -1 to 1. Unless I'm off on that one. That'd obviously be a float, so I don't know an uint16.

> > ### mask
> At least with Sentinel 2 COG STAC, the primary STAC geometry already includes only the valid portion of the image.

Ah, true. We should probably call this out specifically. I think in the label extension we do, since often it just uses a subset of an overall valid image. But perhaps we call it out for EO as well, or even make it a general recommendation - the polygon you use should be the _valid_ portion, not including black fill, etc. I think we'd want extensions to be able to change that behavior if there's some reason to, but could be good to make clear that should be the default. Definitely seems more useful.

kylebarron:
> > > ### Min/Max gsd instead of Min/Max zoom
> > 
> > After some thought and discussion with @geospatial-jeff and @vincentsarago, I'd like to propose that instead of minzoom and maxzoom, we have maxgsd and mingsd, where gsd already exists in STAC and is defined as "Ground Sample Distance at the sensor", and the other is the gsd that corresponds to the smallest overview level.
> 
> I'm on board. Seems like we just need min_gsd for this extension then? And a perhaps a recommendation on how to use gsd to calculate the max? I'll take a crack at it.

Overviews have associated _decimations_, which I believe is defined as the number of times fewer pixels that overview has.

Running `rio cogeo info` on a GeoTIFF from the `naip-analytic` S3 bucket shows:

```
    Id      Size           BlockSize     Decimation
    0       10478x12642    512x512       0
    1       5239x6321      128x128       2
    2       2620x3161      128x128       4
    3       1310x1581      128x128       8
    4       655x791        128x128       16
    5       328x396        128x128       32
```

Then given that the `gsd` for the full-resolution image is `0.6` meters, the `gsd` of the level-5 overview I presume would be `19.2` meters. `min` might not be the best prefix since the value is necessarily larger than `gsd`, because each pixel of the overview covers a larger area.

> > > ### minimum and maximum pixel values for each band of each asset
> > 
> > If you know the dtype is uint16; then you know the max value of each band is 65535, no? Are there real-world examples of the dtype set to uint16 but not all of the range is used?
> 
> I had thought someone mentioned this in the thread, but now I can't seem to find it. But I think the real-world example is NDVI output, that is often just -1 to 1. Unless I'm off on that one. That'd obviously be a float, so I don't know an uint16.

This seems strange to me. I understand that NDVI output ranges from -1 to 1, but why would you store the literal values in a GeoTIFF as -1 to 1? You'd be either losing precision or making a larger file size. If the source data comes as `uint16`, it would seem much better to store the NDVI also as `uint16`, where you wouldn't lose any precision and still store each pixel with 16 bits. Then `0` in the file maps to a logical `-1` and `65535` in the file maps to a logical `1`.

In any case, this isn't something I feel strongly about either way, and it's not too much space to store min and max.

extensions/rendering-hints/README.md
kylebarron: Responded in main thread.

extensions/rendering-hints/README.md
kylebarron: It doesn't hurt to include both I guess

| Type Name | Description |
|-----------|-------------|
| `uint8` | 8-bit unsigned integer |
| `int8` | 8-bit signed integer |

cholmes:
>> min might not be the best prefix since the value is necessarily larger than gsd, because each pixel of the overview covers a larger area.

Hrm good point. `overview_gsd` or something that? `max_overview_gsd`?

geospatial-jeff:
> Are there real-world examples of the `dtype` set to `uint16` but not all of the range is used?

Another good real world example is satellite imagery is often captured between 11-14 bits but stored in a 16 bit image.

cholmes:
As I sat down to try to add the new things discussed it occurs to me that the min/max values per band is the same situation as the stats - we need some sort of per band construct. 

Oh wait, we just decided that the 'bands' information should be included in every item. So I think we could just add on fields to the band. That would be an extension of an extension, which I think is ok? But not sure if that should be a new extension or done in this one.

I think for now I'm going to consider those out of scope, but file new issues for them, so that we can get this extension out the door.

cholmes:
Ok, so I think this draft is getting close to ready. Somehow with all the great suggestions we ended up with less fields. I'll file a ticket on the min/max values, it'll just be trickier to integrate with eo bands.

@kylebarron @geospatial-jeff @vincentsarago - a final review would be great. Also if one of you could help me get together a real world example.  Like could you figure out the overview gsd of one [planet stac sample images](https://planet.stac.cloud/5k3UqPNLpDJMxoAfw1YUV9y9QsbZpgkBacBWwUJ9/3MxsQZbdxjScFVpNqiHrDSMjKgPQo9Uq1JYtn2CAwxwSj9F?t=items)? And the data type? 

@m-mohr - a schema would be great at this point too, hopefully not much will change.

m-mohr:
@cholmes In general this looks good, but I'm not sure I follow you on min/max values. What's the exact reson to not include them here? It feels like it exactly fits into this extension. min/max are likely closely bound to eo:bands, but could also apply to data without bands, like SAR or so. So I'd actually define them on the same level as data type, but also allow them in bands, I guess. We may re-use the Stats Objects from Collections, which is extensible for mean, median or whatever...

extensions/rendering-hints/README.md
```diff
@@ -0,0 +1,67 @@
...
+- TODO
+
+## Item Properties fields
+
+| Field Name       | Type                     | Description |
+| ---------------- | ------------------------ | ----------- |
+| render:overview_max_gsd  | number | The maximum Ground Sample Distance represented in an overview. This should be the GSD of the highest level overview, generally of a [Cloud Optimized GeoTIFF](http://cogeo.org), but should work with any format. |
+| render:data_type    | string | The data `type` (float, int, complex, etc) to let the renderer apply any needed rescaling up front. The full set of options is listed below. |
+
+**render:overview_max_gsd**: This field helps renderers of understand what zoom levels they can efficiently show. It is
```
geospatial-jeff, line 35:
```suggestion
**render:overview_max_gsd**: This field helps renderers understand what zoom levels they can efficiently show. It is 
```

davidraleigh:
Sorry I've been absent after striking up this whole conversation in gitter. Is there any notion of whether it'd be nice to support this earlier request:
> Quadkey is perfect. Having the full list of quadkey might be a bit expensive (in processing/storage/response) so maybe the list of quakdey at the lowest resolution.
Is there any support for storing the lowest resolution quad_key(s) for a stac item?

kylebarron:
> Like could you figure out the overview gsd of one [planet stac sample images](https://planet.stac.cloud/5k3UqPNLpDJMxoAfw1YUV9y9QsbZpgkBacBWwUJ9/3MxsQZbdxjScFVpNqiHrDSMjKgPQo9Uq1JYtn2CAwxwSj9F?t=items)? And the data type?

`rio cogeo info` is helpful for this. You can pass it a URL to a COG or a path to a local COG. For example, as you can see below, the `Dtype` is `uint16`. The IFD section lists the overviews this image contains. The "decimation" represents the level of downsampling for each overview. The [STAC item](https://planet.stac.cloud/item/5k3UqPNLpDJMxoAfw1YUV9y9QsbZpgkBacBWwUJ9/3MxsQZbdxjScFVpNqiHrDSMjKgPQo9Uq1JYtn2CAwxwSj9F/sMSJpYrw6qjYkCm1EJhRCK1hhCMRyJhV8spzrYVRwuZmjssZuCJ9hGo9QriS4uMo?si=2&t=preview#11/29.567842/-95.911077) says that the full-resolution image has a GSD of 2.6 meters. Therefore the smallest overview would have a GSD of `2.6 * 27 = 70.2` meters.

```
> pip install 'rio-cogeo>=2.0a6'
> rio cogeo info https://storage.googleapis.com/pdd-stac/disasters/hurricane-harvey/0831/20170831_172754_101c_3B_AnalyticMS.tif
Driver: GTiff
File: https://storage.googleapis.com/pdd-stac/disasters/hurricane-harvey/0831/20170831_172754_101c_3B_AnalyticMS.tif
COG: True
Compression: LZW
ColorSpace: None

Profile
    Width:            8310
    Height:           3919
    Bands:            4
    Tiled:            True
    Dtype:            uint16
    NoData:           0.0
    Alpha Band:       False
    Internal Mask:    False
    Interleave:       PIXEL
    ColorMap:         False
    ColorInterp:      ('red', 'green', 'blue', 'undefined')
    Scales:           (1.0, 1.0, 1.0, 1.0)
    Offsets:          (0.0, 0.0, 0.0, 0.0)

Image Metadata
    AREA_OR_POINT:         Area
    TIFFTAG_DATETIME:      2017:08:31 17:27:54

Geo
    Crs:              EPSG:32615
    Origin:           (205503.0, 3280287.0)
    Resolution:       (3.0, -3.0)
    BoundingBox:      (205503.0, 3268530.0, 230433.0, 3280287.0)
    MinZoom:          10
    MaxZoom:          15

IFD
    Id      Size           BlockSize     Decimation
    0       8310x3919      256x256       0
    1       2770x1307      128x128       3
    2       924x436        128x128       9
    3       308x146        128x128       27
```

> Is there any support for storing the lowest resolution quad_key(s) for a stac item?

In above discussion I suggested moving to min/max _ground sample distance_ instead of minzoom/maxzoom. In my opinion this is broader because it isn't tied to Web Mercator. Similarly quadkeys are tied to Web Mercator and can be derived from existing information pretty easily:

1. Find zoom level from overview's GSD. As discussed above, it's only a few lines of code, or use rio-cogeo directly: `rio_cogeo.utils.zoom_for_pixelsize(70.2)` -> `11`.
2. Use item's geometry to find intersecting web mercator tiles. For example with [`mercantile`](https://mercantile.readthedocs.io/), it's just: `mercantile.tiles(*geometry.bounds, 11)`.

Both steps should also be straightforward in a language other than Python. Given this I'd argue that storing quadkeys in the STAC is unnecessary.

cholmes:
@m-mohr: 1) I'll get an example in - wanted to get a real example, and just saw @kylebarron provided the info I need to do it. Will try to get it in within a week (on vacation and have a ton of sprint organizing to do). 2) Midway through writing I realized that it was not as hard as I thought, since we use the bands object. But I think the one thing I'm less clear on is 'extending an extension', since bands isn't in core. Just not sure if we do it all in one extension, where some of the fields don't need to extend eo.

@kylebarron  - thanks! This is helpful. 

It does feel like we should have a good 'best practices' section on this stuff, with how to figure out quadkeys, zoom levels, etc. And ideally those link to code in various languages eventually.

m-mohr:
I think it shouldn't only live in bands, because there's more data then EO.

cholmes:
@m-mohr - how do we specify it then? Just at the asset level? And then note that assets can use it in their bands objects if desired?

m-mohr:
There's likely not an easy answer to this. It all depends on the underlying data structure / file format. I guess it needs to be available at eo:bands and in assets, assets being the primary place where it lives in, but in future extensions it could also live in other places. Like in data cubes for example it would also be useful and there's likely more...

Maybe we should just make this extension work for assets and then say in the EO and data cube extension that the rendering extension can be used in their Band / Dimension Objects?

cholmes:
> Maybe we should just make this extension work for assets and then say in the EO and data cube extension that the rendering extension can be used in their Band / Dimension Objects?

Yeah, that's probably a good way to do it. Specify here, and then in EO it can mention that rendering hints extension can be used at the bands level. 

Also should we specify at the item property level and say that it's usually just used at the asset level? So that items with only one asset can use it? Or is it easier to just keep it at the asset level?

m-mohr:
I'd go with the properties level, I guess. For summary reasons and we did that for most fields recently and just allow everything also in assets. But no strong preferrence from my side.

m-mohr:
Should we add the min/max values per data type to the table? Then these would be default values for the min/max fields, if not provided, right?

cholmes:
> Should we add the min/max values per data type to the table? Then these would be default values for the min/max fields, if not provided, right?

Yes, great idea.

m-mohr:
Any progress on this? I'd like to re-use it in CARD4L...

m-mohr:
#934 also proposes to add the data type field, so we may remove it here.

cholmes:
Yeah, just noticed the data type field in the file extension. I agree it makes more sense there, it's more general. So let's remove it here. 

As for progress, I'm hoping to work on it sometime soon - paused all my stac core spec work in favor of API.

extensions/rendering-hints/README.md
```diff
@@ -0,0 +1,67 @@
+# Rendering Hints Extension Specification
+
+- **Title: Rendering Hints**
+- **Identifier: rendering_hints**
```
m-mohr, line 4:
```suggestion
- **Identifier: rendering-hints**
```

kylebarron:
Anything I can do to help move this forward? We're removing the data type from this ext since it's now in the file extension, and adding data range? 

One question about data range: would it represent the "global" minimum and maximum across the dataset or the "local" minimum and maximum within that specific scene? For example, Sentinel 2 data that's 12 bits would presumably have a global min/max of 0-4095, but if the specific scene isn't fully saturated, the scene's min and max might be a smaller range.

cholmes:
@kylebarron if you want to work on this that'd be awesome.

Though I think this particular extension ends up being pretty minimal. I think the path ahead is:

* data type is in file extension, so no need to have it here.
* min / max should go in 'stats' extension - #906 - as it has uses other than just rendering. No one has started that.
* This extension retains the `overview_max_gsd` field. And perhaps just that?

So the main work is probably on the 'stats' extension, getting it written up and into a PR, and making sure the structure it proposes works with the different ways people organize files.

> One question about data range: would it represent the "global" minimum and maximum across the dataset or the "local" minimum and maximum within that specific scene?

In my mind this is mostly per item, so yeah, it's the 'local' one. You might put the global in a 'summary' at the collection level.

matthewhanson:
As extensions now live in the stac-extensions organization, this PR is being closed. Follow instructions here to create a repository for the extension:
https://github.com/stac-extensions/stac-extensions.github.io#adding-a-new-extension

emmanuelmathot:
Progress on this at https://github.com/stac-extensions/raster

PR #880, opened 2020-08-19
Add clarification around collection-level assets in item-assets.

lossyrob:
This PR adds language to the `item-assets` spec to clarify that collection-level asset objects should not contain properties that are not explicitly stated in the item's asset objects themselves.

This comes out of a discussion during STAC Sprint 6. It was determined that while allowing for asset properties to be defined in _only_ the collection-level asset objects was a way to save space, it would be more in line with the spec to have items explicitly contain all asset property information.

If you're currently using the item-assets extension to consolidate item-level asset object properties in an API setting, consider storing the STAC items without those consolidated properties internally as 'invalid' STAC items, and merge in the desired collection-level asset object properties at serving time.

**PR Checklist:**

### DIFF
--- a/extensions/item-assets/README.md
+++ b/extensions/item-assets/README.md
@@ -42,7 +42,8 @@ However, it is recommended that the Asset Definition is a complete set of all as
 
 Other custom fields, or fields from other extensions may also be included in the Asset object.
 
-Any property that exists for a collection-level asset object must also exist in the correspond assets object in each item. If a collection's asset object contains properties that are not explicitly stated in the item's asset object then that property does not apply to the item's asset. Collection-level asset objects can describe any of the properties of an asset, but those assets properties and values must also reside in the item's asset object.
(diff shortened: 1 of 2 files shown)


extensions/item-assets/README.md
```diff
@@ -42,6 +42,8 @@ However, it is recommended that the Asset Definition is a complete set of all as
 
 Other custom fields, or fields from other extensions may also be included in the Asset object.
 
+Any property that exists for a collection-level asset object must also exist in the correspond assets object in each item. If a collection's asset object contains properties that are not explicitly stated in the item's asset object then that property does not apply to the item's asset. Collection-level asset objects can describe any of the properties of an asset, but those assets properties and values must also reside in the item's asset object.
```
m-mohr, line 45:
```suggestion
Any property that exists for a collection-level asset object must also exist in the corresponding assets object in each item. If a collection's asset object contains properties that are not explicitly stated in the item's asset object then that property does not apply to the item's asset. Item asset objects at the collection-level can describe any of the properties of an asset, but those assets properties and values must also reside in the item's asset object. To consolidate item-level asset object properties in an API setting, consider storing the STAC items without the larger properties internally as 'invalid' STAC items, and merge in the desired properties at serving time from the collection-level.
```

* I thin the correspond must be corresponding?
* Collection-level assets is a separate extension. To avoid confusion, changed the wording.
* I like the note on the PR description regarding APIs. Added that, too.

m-mohr (approved):

Issue #822, opened 2020-05-29
Use h4 instead of bold in table explanations?


cholmes:
One of my minor annoyances when using the spec is that I'll want to refer to a particular section but it's done in bold and there's no hyperlink. 

The latest was summaries - I have to say: 'go to https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md#collection-fields and scroll down'.

We could consider just using the smallest markdown heading, so we get an automatic hyperlink.

m-mohr:
Linking to the description below is also not ideal:
1. It only covers things with a description below the table
2. It doesn't cover the "formal" data type description

I feel like linking to the collection fields fine, but on my screen I see it directly with the link, not sure about smaller screens.

cholmes:
This is changed to must have as it's now in an inconsistent state, as h4's were added in #942 so it could link better.

PR #883, opened 2020-08-20
Unique ids

cholmes:
**Related Issue(s):** #1011 #822 

**Proposed Changes:**

1. Added more information to make sure ID's are unique within a collection.
2. Added h4's to the item spec.

**PR Checklist:**

- [ ] This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec), and I have opened issue/PR https://github.com/radiantearth/stac-api-spec/issues/103 to track the change.


cholmes:
Just put this up for discussion. Wondering if we should be more explicit about what makes a ID globally unique? Like is it collection ID plus the URL of one of the provider roles? Should that role be producer? Or host? Host is the one where we say explicitly there should just be one, so perhaps that makes the most sense.

lossyrob:
I think these constraints would solve this:
- Item IDs are unique within their collection. If there is no collection, item IDs must be unique within their parent catalog.
- Catalog and Collection IDs must be unique within the root catalog (ideally globally unique everywhere)

I'd advocate for not having to parse out Provider information in order to have collection IDs unique; providers aren't required and it would make the constraint more simple across catalogs and collections if the ID was required to have all information necessary for it to be globally unique (i.e. the STAC creator can insert the provider name into the IDs if that's needed to make them unique).

pomadchin:
@lossyrob It is a bit unclear (since that was just an oral conversation), but from what I understand it can be an actual `Collection` with links to items. 

It would definitely work if items would be identified by both collection ID and item ID in a such collection. :+1:

m-mohr:
I'm fine with @lossyrob proposal, except that I would say:

 * Catalog and Collection IDs must be unique within the ~root~ **parent** catalog

Re aggregations: I understood that aggregations could be Collections, but I don't think they should duplicate IDs. They are somewhat "virtual" anyway.

m-mohr:
Good point regarding APIs. But then I'd say they need to be unique for the parent collection, which is usually different to the root catalog. That works for APIs and makes combining several collections into whatever catalogs easier. You just can't always guarantee uniqueness if you combine different sources, like Matt does with Earth Search or I do with STAC Index (which itself are catalogs again).

cholmes:
Picking this up after a long delay... And I'm confused as to who is advocating what.

The core recommendation seems to be id's should be unique for the parent collection. And we strongly recommend that. If items don't have a parent collection then it seems like they do need to be unique in the 'root'. 

There is the case where some meta catalog wants to include a catalog that doesn't have collections, and thus can't guarantee that the id is unique. I'd say for that we just warn people that if they define a items without collections then their catalogs just won't be nearly as used.

cholmes:
Ah, now I see that I was confused, was mixing up collections and items.

I just committed an attempt at this. Basically said that id's in a collection need to be unique, and that collection id's should aim to be globally unique. And handled the 'no collection' use case by saying that Items should attempt to make their id's globally unique (and remind people we strongly recommend a catalog).

collection-spec/collection-spec.md
```diff
@@ -19,7 +19,7 @@ STAC Collections are meant to be compatible with *OGC API - Features* Collection
 | --------------- | ------------------------------------------------ | ------------------------------------------------ |
 | stac_version    | string                                           | **REQUIRED.** The STAC version the Collection implements. STAC versions can be mixed, but please keep the [recommended best practices](../best-practices.md#mixing-stac-versions) in mind. |
 | stac_extensions | \[string]                                        | A list of extension identifiers the Collection implements. |
-| id              | string                                           | **REQUIRED.** Identifier for the collection that is unique across the provider. |
+| id              | string                                           | **REQUIRED.** Identifier for the collection that is unique across the provider, ideally globally unique. |
```
m-mohr, line 22: I'm not sure this works in practice. I don't think providers often want to complicate their collection ids and have something like https://provider.com/collections/provider.com.sentinel-2 - instead they will usually just use sentinel-2, especially when the ids are re-used in user-oriented processing workflows like in openeo.

item-spec/item-spec.md
```diff
@@ -40,12 +40,33 @@ inherited from GeoJSON.
...
 
-**stac_version**: In general, STAC versions can be mixed, but please keep the [recommended best practices](../best-practices.md#mixing-stac-versions) in mind.
+### Additional Field Information
 
-**stac_extensions**: A list of extensions the Item implements. The list contains URLs to the JSON Schema files it can be validated against. For official [extensions](../extensions/README.md#list-of-stac-extensions), a "shortcut" can be used. This means you can specify the folder name of the extension, for example `pointcloud` for the Point Cloud extension. This does *not* apply for API extensions. If the versions of the extension and the item diverge, you can specify the URL of the JSON schema file.
+#### stac_version
+
+In general, STAC versions can be mixed, but please keep the [recommended best practices](../best-practices.md#mixing-stac-versions) in mind.
+
+#### id
```
m-mohr, line 49: I'd align with the order in the table and move this below the stac_extensions.

m-mohr:
I'm fine with unique Item ids in a collection.
I'm also fine with items having a unique id up until the root catalog they define if there's no collection.

I have issues with the "collection id" must be globally unique. I see that collection id's should be unique across a provider, but globally uniqueness no one can really guarantee. I don't think there will be broad adoption of it, especially as the PR doesn't say why this is useful. Additionally, most people have their collection IDs already defined before adopting STAC and I don't think many people will change them (at least I don't see anyone in openEO doing it, I doubt GEE would do that - the obvious reason is that we use them in processing workflows and different IDs just in STAC would break things and confuse users).

matthewhanson (approved):

item-spec/item-spec.md
```diff
@@ -40,12 +40,33 @@ inherited from GeoJSON.
...
+
+It is important that an Item identifier is unique within a collection, and that the 
+[Collection identifier](../collection-spec/collection-spec.md#id) in turn is unique globally. Then the two can be combined to
+give a globally unique identifier. Items are *[strongly recommended](#collections)* to have collections, and not having one makes
+it more difficult to be used in the wider STAC ecosystem. If an Item does not have a collection then it should strive to provide
+a globally unique identifier, with a long, unique prefix, to have a better chance of being globally unique.
```
lossyrob, line 55:
```suggestion
It is important that an Item identifier is unique within a collection, and that the 
[Collection identifier](../collection-spec/collection-spec.md#id) in turn is unique globally. Then the two can be combined to
give a globally unique identifier. Items are *[strongly recommended](#collections)* to have collections, and not having one makes
it more difficult to be used in the wider STAC ecosystem. Iff an Item does not have a Collection, then the Item identifier should be unique within it's root Catalog.
```

collection-spec/collection-spec.md
lossyrob, line 22: Agree after conversation in working session; suggested change removes globally unique as the ideal case.

```suggestion
| id              | string                                           | **REQUIRED.** Identifier for the collection that is unique across the provider. |
```

lossyrob (approved):
Added suggested changes, otherwise +1

cholmes:
Ok, merging this. Committed Rob's changes, as Matt, Rob and Matthias all agree on not having the 'globally unique' stuff for collections.

item-spec/item-spec.md
m-mohr: @cholmes I guess you've not seen this?

Issue #885, opened 2020-08-20
Add TileDB media type to item spec


normanb:
A TileDB array is a prefix / folder and does not have an extension. As such the use of the `type` is required when linking to this media type.

m-mohr:
We are waiting for a TileDB example in STAC. Closing for now until something has been added to the repo.

PR #886, opened 2020-08-20
Added TileDB media type

normanb:
**Related Issue(s):** #885

**Proposed Changes:**

1. Add TileDB as a media type in the item spec.

**PR Checklist:**

- [ ] I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-spec/blob/dev/CHANGELOG.md) **or** a CHANGELOG entry is not required. *Change log entry not required*
- [x] This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec), and I have opened issue/PR #885 to track the change.

### DIFF
--- a/item-spec/item-spec.md
+++ b/item-spec/item-spec.md
@@ -217,6 +217,7 @@ Common STAC Item Media Types:
 | `application/geopackage+sqlite3`                        | GeoPackage                                                   |
 | `application/x-hdf5`                                    | Hierarchical Data Format version 5                           |
 | `application/x-hdf`                                     | Hierarchical Data Format versions 4 and earlier.             |
+| `application/x-tiledb`                                  | TileDB                                                       |
 
 Deprecation notice: GeoTiff previously used the media type `image/vnd.stac.geotiff` and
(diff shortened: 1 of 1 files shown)


m-mohr:
I hope I'm not too picky here, but the table says "Common STAC Item Media Types", but I have not seen a single Catalog with this media type. Is this really common at the moment?

cholmes:
Actually, now that @m-mohr points it out, I do think we should have at least one reference STAC that uses TileDB before adding it here. Usually to add things to the core spec they start as extensions, and then when there's appropriate use they get in the main spec. This is perhaps a bit different, as it doesn't seem like any extension is really needed, outside having a media type in the spec. 

Though actually an extension could serve a good use, as I think TileDB arrays are a bit more flexible than COG's, so it could be good to have some recommendations on how to use them, and if there's any additional metadata fields to put in. @normanb - starting it as an extension would also let us get it in right away, as you can just mark it as 'proposal'. And then after there's reference STAC's to point at we can add the media type, and also figure out if we still need the extension.

Thanks!

m-mohr:
This seems to be stale, so closing for now until the comments have been adressed.

PR #895, opened 2020-09-09
wordsmith text

philvarner:
**Related Issue(s):** # n/a

**Proposed Changes:**

1. Clarify text
2. Clarify which fields are for Item Asset and Item Properties
3. remove "QA" as an example for a band
4. expand the common_name rededge wavelength range to be what Sentinel-2 considers rededge
5. add Sentinel-2 rededge band #s
6. add NAIP bands as an example

**PR Checklist:**

### DIFF
--- a/extensions/eo/README.md
+++ b/extensions/eo/README.md
@@ -6,8 +6,13 @@
 - **Scope: Item**
 - **Extension [Maturity Classification](../README.md#extension-maturity): Proposal**
 
-This document explains the fields of the STAC Electro-Optical (EO) Extension to a STAC Item. EO
-data is considered to be data that represents a snapshot of the earth for a single date and time. It
+This document explains the fields of the STAC Electro-Optical (EO) Extension to a STAC Item. 
+
+These fields defined by this extension follow the convention for 
+[additional fields for a STAC Item](../../item-spec/item-spec.md#additional-fields-for-assets) and are 
+allowed in either Item Properties or Item Assets.  
+
+EO data is considered to be data that represents a snapshot of the Earth for a single date and time. It
 could consist of multiple spectral bands in any part of the electromagnetic spectrum. Examples of EO
 data include sensors with visible, short-wave and mid-wave IR bands (e.g., the OLI instrument on
 Landsat-8), long-wave IR bands (e.g. TIRS aboard Landsat-8).
@@ -22,22 +27,16 @@ For defining view geometry of data, it is strongly recommended to use the [`view
   - [Landsat 8 with bands in Item Asset Definition and Collection Summaries](../item-assets/examples/example-landsat8.json)
 - [JSON Schema](json-schema/schema.json)
 
-## Item Asset fields
+## Item Properties or Item Asset fields
 
 | Field Name     | Type                           | Description |
 | -------------- | ------------------------------ | ----------- |
-| eo:bands       | \[[Band Object](#band-object)] | An array of available bands where each entity is a [Band Object](#band-object). |
+| eo:cloud_cover | number                         | Estimate of cloud cover |
+| eo:bands       | \[[Band Object](#band-object)] | An array of available bands where each object is a [Band Object](#band-object). |
 
-**eo:bands**: In STAC versions 0.9.x and prior, `eo:bands` was only allowed to be used in an Asset by referencing the Band Object definitions in Item Properties via array index. Starting with STAC 1.0.0-beta.1, an Asset must have an `eo:bands` array containing Band Object entities, as described in general in the [STAC Item](../../item-spec/item-spec.md#additional-fields-for-assets).  
+**eo:cloud_cover**: Estimate of cloud cover as a percentage (0-100) of the entire scene. If not available, the field should not be provided. Generally, this value should be used in Item Properties rather than Item Assets, as an Item from an electro-optical source is a single snapshot of the Earth, so the cloud cover value would apply to all assets. 
 
-## Item Properties fields
-
-| Field Name     | Type                           | Description |
-| -------------- | ------------------------------ | ----------- |
-| eo:cloud_cover | number                         | Estimate of cloud cover as a percentage (0-100) of the entire scene. If not available, the field should not be provided. |
-| eo:bands       | \[[Band Object](#band-object)] | An array of available bands where each entity is a [Band Object](#band-object). |
-
-**eo:bands**: This is the same as the attribute for Item Asset. As of STAC 1.0.0-beta.1, there is no way to reference these Band Objects from an Asset, so their primary use is simply to advertise a set of bands that may be available in some of the Assets in this Item. These can also be rolled up into Collection `summaries` object.
+**eo:bands**: In STAC versions 0.9.x and prior, `eo:bands` could only be used by an Asset putting the the Band Object definitions in an Item Properties and referencing these via array index. **Starting with STAC 1.0.0-beta.1, an Asset may only have an `eo:bands` array containing Band Object entities.** Since `eo:bands` definitions in Item Properties can no longer be referenced by array index from an Asset, their primary use now is only to advertise a set of bands that may be available in some of the Assets in this Item, which may be aggregated into Collection `summaries` object.
 
 ### Band Object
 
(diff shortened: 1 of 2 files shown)


extensions/eo/README.md
```diff
@@ -22,20 +22,28 @@ For defining view geometry of data, it is strongly recommended to use the [`view
   - [Landsat 8 with bands in Item Asset Definition and Collection Summaries](../item-assets/examples/example-landsat8.json)
 - [JSON Schema](json-schema/schema.json)
 
-## Item fields
+## Item Asset fields
```
philvarner, line 25: It wasn't clear from this text into which entities these fields were supposed to go.

extensions/eo/README.md
```diff
@@ -22,20 +22,28 @@ For defining view geometry of data, it is strongly recommended to use the [`view
...
+| eo:cloud_cover | number                         | Estimate of cloud cover as a percentage (0-100) of the entire scene. If not available, the field should not be provided. |
+| eo:bands       | \[[Band Object](#band-object)] | An array of available bands where each entity is a [Band Object](#band-object). |
+
+**eo:bands**: This is the same as the attribute for Item Asset. As of STAC 1.0.0-beta.1, there is no way to reference these Band Objects from an Asset, so their primary use is simply to advertise a set of bands that may be available in some of the Assets in this Item. These can also be rolled up into Collection `summaries` object.
 
 ### Band Object
 
 | Field Name          | Type   | Description |
 | ------------------- | ------ | ----------- |
-| name                | string | The name of the band (e.g., "B01", "B02", "B1", "B5", "QA"). |
```
philvarner, line 38: a "QA" band doesn't conform to the definition of what these are supposed to be.  Might be useful in the future to have a "layer" extension or something similar for image "bands" that are derived values, e.g., a QA mask, NDVI, Land Surface Temerature

extensions/eo/README.md
```diff
@@ -52,24 +60,24 @@ The band's common_name is the name that is commonly used to refer to that band's
...
 
-| Common Name | Band Range (μm) | Landsat 5/7 | Landsat 8 | Sentinel 2 | MODIS |
-| ----------- | --------------- | ----------- | --------- | ---------- | ----- |
-| coastal     | 0.40 - 0.45     |             | 1         | 1          |       |
-| blue        | 0.45 - 0.50     | 1           | 2         | 2          | 3     |
-| green       | 0.50 - 0.60     | 2           | 3         | 3          | 4     |
-| red         | 0.60 - 0.70     | 3           | 4         | 4          | 1     |
-| yellow      | 0.58 - 0.62     |             |           |            |       |
-| pan         | 0.50 - 0.70     | 8 (*L7 only*) | 8       |            |       |
-| rededge     | 0.70 - 0.75     |             |           |            |       |
```
philvarner, line 63: Sentinel-2 calls band 7 (which goes up to 0.782μm) a red edge band, so let's expand the range slightly.

item-spec/examples/sample-full.json
```diff
@@ -43,16 +43,20 @@
     "instruments": ["cool_sensor_v1"],
     "eo:bands": [
       {
-        "name": "band1"
+        "name": "band1",
```
philvarner, line 46: these are in Item Properties

item-spec/examples/sample-full.json
```diff
@@ -74,7 +78,25 @@
   "assets": {
     "analytic": {
       "href": "http://cool-sat.com/catalog/CS3-20160503_132130_04/analytic.tif",
-      "title": "4-Band Analytic"
+      "title": "4-Band Analytic",
+      "eo:bands": [
```
philvarner, line 82: add Asset eo:bands to this example.

extensions/eo/README.md
m-mohr: We are generally allowing any fields from Item properties in Item Assets, so no need to specify this explicitly. Otherwise you'd need to duplicate the description for all the other fields as well, e.g. eo:cloud_cover and all other extensions.

https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#asset-object says: 
> Beyond to the fields above, any of the additional fields may be added to the assets. But this is recommended only in special cases, see Additional Fields for Assets) for more information.

So I'd prefer to remove this duplication. What we better add is a general hint that eo:bands might be better suited in assets (refering back to what I quoted abvoe) in certain cases.

extensions/eo/README.md
m-mohr: I don't understand this. Why is QA not supposed to be a band if it's offered like the other bands?

extensions/eo/README.md
philvarner: Thanks for clarifying. I'll update this now that I understand what that means.

extensions/eo/README.md
philvarner: I had the same question, and I had previously been adding EO bands for QA (quality assurance) and derived data (e.g., NDVI, temperature). The intro to this extension "It
could consist of multiple spectral bands in any part of the electromagnetic spectrum. Examples of EO
data include sensors with visible, short-wave and mid-wave IR bands (e.g., the OLI instrument on
Landsat-8), long-wave IR bands (e.g. TIRS aboard Landsat-8)." -- which seems to me (apart from the "could" weasel word) to indicate it's only for data bands that represent wavelengths.

In the example for this extension, the "band" BQA doesn't have an eo:bands definition, whereas all the other data bands do have it. 

Also, since eo:bands have the wavelength and full_width_half_max, and all the common names are light spectra, it seems to be very specific to light from a sensor and not derived products like a QA band. 

I'm open to it either being specific to electromagnetic or any band/layer -- I was just trying to clarify the spec wrt to how I interpreted it.

m-mohr: I see the point. I guss we should leave it open to implementors whether it's useful for them to add such bands or not. I could see good reasons to add them, I'd certainly do. But your arguments make sense, too.

m-mohr: And thanks for clarifying!

extensions/eo/README.md
m-mohr: Maybe we need to improve how we describe the general behavior.

extensions/eo/README.md
m-mohr: Are there any "official"/scientific values we could relate to here?

extensions/eo/README.md
philvarner: I'll also look at clarifying the general behavior description.

extensions/eo/README.md
philvarner: There may be, but I the approach we've take is the right one -- look at what range the values that each sensor calls "red", etc and make that the range for common_name red.

extensions/eo/README.md
```diff
@@ -22,20 +27,22 @@ For defining view geometry of data, it is strongly recommended to use the [`view
   - [Landsat 8 with bands in Item Asset Definition and Collection Summaries](../item-assets/examples/example-landsat8.json)
 - [JSON Schema](json-schema/schema.json)
 
-## Item fields
+## Item Properties or Item Asset fields
```
philvarner, line 30: I would like to go through all of the extensions and make this consistent.  I found it confusing when this was "Item" because that could be interpreted to mean that it's a new attribute on Item, rather than it can be added to Item Properties or Item Assets. I looked a a few of the other extensions and some have "Item" and some "Item Properties" (and maybe those also apply to Assets?). Other extensions have this heading as the Object the properties go in, e.g., Collection or Catalog.

extensions/eo/README.md
```diff
@@ -22,20 +27,22 @@ For defining view geometry of data, it is strongly recommended to use the [`view
...
 
 | Field Name     | Type                           | Description |
 | -------------- | ------------------------------ | ----------- |
-| eo:bands       | \[[Band Object](#band-object)] | This is a list of the available bands where each item is a [Band Object](#band-object). |
-| eo:cloud_cover | number                         | Estimate of cloud cover as a percentage (0-100) of the entire scene. If not available the field should not be provided. |
+| eo:bands       | \[[Band Object](#band-object)] | An array of available bands where each object is a [Band Object](#band-object). |
```
philvarner, line 34: list -> array and item (an overloaded term) -> object

philvarner:
I've finished updating, ready for re-review.

extensions/eo/README.md
cholmes: +1 I'll put up a ticket to track. We only recently added the ability to flexibly put things on assets. I think technically people can put any property field on an asset, but we should call out in each extension where we think that's appropriate.

extensions/eo/README.md
cholmes: +1

cholmes (approved):

matthewhanson (approved):

PR #900, opened 2020-09-16
Clairify if fields are on Item Properties and/or Item Asset

philvarner:
I'm going to split out the minor wordsmithing into a separate PR, but I wanted some impressions on the text around extension fields in Properties or Assets. 

One issue with the Projection Extension is that it's not clear if all fields should be allowed on both Item Properties and Assets, or only the proj:shape and proj:transform should be -- they're the only two that mentioned being in Assets specifically. I feel any of them should be allowed in Assets. 

**Related Issue(s):** #

n/a

**Proposed Changes:**

1. Use 'object' instead of 'entity' and 'field' instead of 'attribute'
2. Explicitly state that most extensions on Item will apply to both Item Properties and Item Asset
3. Other clarifying text

**PR Checklist:**

- [ ] n/a This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec), and I have opened issue/PR #XXX to track the change.


extensions/README.md
```diff
@@ -1,22 +1,33 @@
-# Extensions
+# Content Extensions
```
m-mohr, line 1: Not sure whether this are all "content" extensions, e.g. collection assets or item assets seem to be more on the "behavior" side and don't add much "additional" content...

Also, it sounds like there are other extensions for STAC spec, but where are they? So I guess I'd just replace content extension with extension throughout the whole document.

m-mohr (approved):
Looks good. Is this complete though? Or just tacking a subset of extensions? Also, please note the comment about "content" extensions.

extensions/README.md
philvarner: "Content" was to distinguish it from "API" -- but I think it should be obvious by it being in this spec rather than the api spec. I'll remove it. 

I also have some clarifying text for the Extensions to clarify if the fields apply to Properties, Assets, or both.  I wanted to clarify this text because one of my co-workers recently started working with the Label Extension, and mistakenly thought the fields were supposed to be added to the Item rather than Item Properties.

cholmes: Agree that it's all 'content' extensions now in the core spec. 

And clarifying text on extensions applying to properties, assets or both is definitely appreciated. With the ability to put things at either level it's good to tell people where they make more sense, and the special situations that might warrant both or breaking from the norm.

extensions/README.md
```diff
@@ -40,9 +51,9 @@ the less breaking changes of the next level.
 A 'mature' classification level will likely be added once there are extensions that have been
 stable for over a year and are used in twenty or more implementations.
 
-## List of content extensions
```
philvarner, line 43: @m-mohr this is where I got the "content extensions" phrase

philvarner:
@cholmes @matthewhanson this is ready for review

cholmes (approved):

extensions/README.md
cholmes: Added #972 to track this.

Issue #897, opened 2020-09-14
sar:product_type to sat extension ?


ycespb:
Product_type (defined in the "sar" extension) is a useful property, not only for SAR granules.  Could it not be moved to a more general extension such as "sat" to use it for optical and other instruments ?  Alternatively, a similar "product_type" property could be added to the "sat" extension as well.

m-mohr:
On the one hand, I see that it can be useful for other products that are not SAR.
On the other hand, I think it's useful for search to either a specific enum or at least some common values for better alignment between catalogs and those values should be listed in the corresponding extensions. 
Another question is whether it's best suited in SAT or somewhere else? Product Types don't seem to be really SAT specific and could also be part of aerial imagery, UAVs etc. This might even go into Common Metadata as there's no better place yet.

ycespb:
Agree this is not specific for satellites.  Is the EO extension assumed to be more general than "SAT" and a better place then ?    We also have sensorType, operationalMode etc if you want to make precise searches and these do not only apply to "sar" instruments...  The mapping for these is currently not obvious in https://github.com/stac-utils/stac-crosswalks/blob/master/OGC_17-003r2/README.md. Would be nice to give these a proper place and allow for a mapping.

m-mohr:
EO (Electro-Optical! - not Earth Observation) is not really suitable for SAR, so then we wouldn't have the product_type for SAR any longer. I don't really know yet where to put it...

matthewhanson:
I think it might better belong in common metadata, but we would want recommendations on how to use it for different types of data.  Optical data would have product types for surface reflectance or TOA, while SAR we have enumerated.  

What other types of datasets would use products?

ycespb:
The OGC specs https://docs.opengeospatial.org/is/10-157r4/10-157r4.html and https://docs.opengeospatial.org/is/17-003r2/17-003r2.html were the result of consultations with many agencies and productType is in both specs  a property that can be used for any mission (optical (opt), radar (sar), altimetric (alt), synthesis, etc.)  It may have ground-segment or mission specific values.  E.g. the altimetric mission Cryosat-2 has about 20 possible values like SIR_GDR_2_, SIR_GOP_2_, SIR_GOPM_2, SIR_GOPM1B...  Note that in the layered representations of earth observations (shown below copied from OGC 10-157r4, including SAR), productType was a property at the "EOP" thus most general level.  

[...]

PR #881, opened 2020-08-19
Lineage best practice #863

m-mohr:
**Related Issue(s):** #863

**Proposed Changes:**

1. Moved the derived_from note to best practices, makes the item spec a bit slimmer and bundles details at a single place.
2. Added idea from #863 for lineage information to BP.

**PR Checklist:**

- [x] This PR is made against the dev branch (all proposed changes except releases should be against dev, not master).
- [x] This PR has **no** breaking changes.
- [ ] I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-spec/blob/dev/CHANGELOG.md) **or** a CHANGELOG entry is not required.
- [ ] This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec), and I have opened issue/PR #XXX to track the change.

PR #907, opened 2020-10-05
[Extension] Processing

emmanuelmathot:
**Proposed Changes:**

***Processing extension***
Processing metadata is considered to be data that indicate from which processing chain a data originates and how the data itself has been produced. Overall, it helps to increase traceability and search among processing levels and multiple algorithm versions.

**PR Checklist:**


m-mohr:
I have not looked into details yet, but some remarks:

Why not make this extend the providers array? There's a role "processor" and you may want to provide processing instructions for multiple providers... That way we would also get rid of `proc:organisation`, which duplicates content.

How does proc:datetime relate to https://github.com/radiantearth/stac-spec/tree/master/extensions/timestamps? Can we use a known timestamp instead of a separate property?

How does proc:name (any maybe proc:level) relate to the product_type fields? Currently in sar, but could also be in commons soon? See https://github.com/radiantearth/stac-spec/issues/897

Could we name this differently? I think "Processing" is a bit too generic, basically due to the fact that there's a different extension planned with the same name (so that both will need a more specific name).

emmanuelmathot:
> I have not looked into details yet, but some remarks:
> 
>     1. Why not make this extend the providers array? There's a role "processor" and you may want to provide processing instructions for multiple providers... That way we would also get rid of `proc:organisation`, which duplicates content.
> 
>     2. How does proc:datetime relate to https://github.com/radiantearth/stac-spec/tree/master/extensions/timestamps? Can we use a known timestamp instead of a separate property?
> 
>     3. How does proc:name (any maybe proc:level) relate to the product_type fields? Currently in sar, but could also be in commons soon? See #897
> 
>     4. Could we name this differently? I think "Processing" is a bit too generic, basically due to the fact that there's a different extension planned with the same name (so that both will need a more specific name).

1. Indeed, `stac:organisation` is a duplicate of the provider object. Let's remove it.
2. `proc_datetime` is very similar to `created`. A quick note to the production date could be added in the timestamps extension and get rid of `proc:datetime`
3. `proc_name` is only a label and should reference the processing name from the provider (e.g. processing section in the [manifest.safe of the Copenicus Sentinels products](https://sentinel.esa.int/web/sentinel/user-guides/sentinel-1-sar/data-formats/safe-specification)) and could be used for free text search. As described, `proc:level` is more formal even if it is not standardized. In the earth observation sector, the level is often linked of the product_type but is really not formalized. For instance, Sentinel-1 `GRD` is `L1`, Sentinel-2 `S2MSI1C` is `L1C`, Sentinel-3 `SR_2_LAN___` is `L2`, Landsat-8 `L1TP` is a specific `L1` known as `L1TP`. The processing level is a filter that is very used for searching spatio-temporal sensor items in a  catalog. Having the `proc:level` properties formalized as the proposed way allows to make query filtering over some first characters (e.g. `L1` or `L2A`) to find items or collections corresponding to a certain product processing.
4. Yes it could be named as `production` with `prod:` prefix. I proposed `processing` to avoid confusion as it is the common terminology when it comes to the data processing level:
   * [NASA's Earth Observing System Data and Information System](https://earthdata.nasa.gov/collaborate/open-data-services-and-software/data-information-policy/data-levels)
   * [The National Snow and Ice Data Center (NSIDC)](https://nsidc.org/the-drift/2013/08/is-it-1b-2-or-3-definitions-of-data-processing-levels/)
   * [Committee on Earth Observation Satellites (CEOS) Interoperability Handbook p.10](http://ceos.org/document_management/Working_Groups/WGISS/Documents/WGISS_CEOS-Interoperability-Handbook_Feb2008.pdf)

emmanuelmathot:
How to move forward? What is the procedure?

m-mohr:
Update the document according to the "review", e.g. removing the organisation field? I'll try to do a more in-depth review tomorrow or next week.

cholmes:
Hey @emmanuelmathot - I'm sorry I've been remiss in reviewing this PR, but I wanted to say thanks for putting all this work in. I'm excited to see this type of thing as an option for STAC implementors, and appreciate that it comes from a lot of direct experience. 

I've been slammed, and now I'm putting all my focus into getting to STAC API 1.0.0-beta.1, but I wanted to say that we have not forgot about you. 

The general procedure is to continue to update your branch as feedback comes in. It's easiest if you make those changes, but others can make PR's on your branch too I think.

extensions/processing/README.md
```diff
@@ -0,0 +1,64 @@
...
+- [Examples](examples/): [Sentinel-1 Item](examples/sentinel1-item.json),  [Sentinel-2 Level 1C Collection](examples/sentinel2-collection.json)
+- [JSON Schema](json-schema/schema.json)
+
+## Item and Collection fields
+
+For Items, the fields are placed in the properties. For Collections, the fields are placed in the summaries of the Collection.
+
+| Field Name                  | Type      | Description                                                  |
+| --------------------------- | --------- | ------------------------------------------------------------ |
+| proc:name         | string | The name of the processing of the data that is commonly used. For example, `GRD Post Processing` for "GRD" product of Sentinel-1 satellites. |
```
m-mohr, line 24: What's the difference between proc:name and for example sar:product_type? Could that be somehow aligned?

extensions/processing/README.md
```diff
@@ -0,0 +1,64 @@
...
+Processing metadata is considered to be data that indicate from which processing chain a data originates and how the data itself has been produced. Overall, it helps to increase traceability and search among processing levels and multiple algorithm versions.
+
+Often, data items are the result of one or more waterfall processing pipeline. Tracing information such as the processing facility , the algorithm version or the processing date helps in the data version management.
+
+This extension applies to STAC Items and STAC Collections. As these processing information are often closely bound to the collection level and therefore are shared across all items, it is recommended adding the fields to the corresponding STAC Collection.
+
+- [Examples](examples/): [Sentinel-1 Item](examples/sentinel1-item.json),  [Sentinel-2 Level 1C Collection](examples/sentinel2-collection.json)
+- [JSON Schema](json-schema/schema.json)
+
+## Item and Collection fields
```
m-mohr, line 18: We have recently also discussed with @prateekbudhwar and colleagues to add processing details to STAC (outdated PR #881). The idea we had was to simply add a `lineage` free-text field to the providers array in Collections so that you can track the processing for each entity that has processed date. Would you think that makes also sense for your fields?

Example:
```
{
  "id": "...",
  "description": "...",
  "license": "proprietary",
  "providers": [
    {
      "name": "European Union/ESA/Copernicus",
      "roles": [
        "producer",
        "licensor"
      ],
      "proc:lineage": "Atmospheric Correction..."
    },
    {
      "name": "SomeComp",
      "roles": [
        "processor"
      ],
      "proc:lineage": "Foo bar... NDVI processing..."
    },
    {
      "name": "Amazon",
      "roles": [
        "host"
      ],
      "proc:lineage": "Conversion to COG..."
    }
  ],
  "extent": {...}
}
```

I think for Collections it may make sense to push your suggested fields into the providers array like proc:lineage in this example. Could be overkill for items and we just put it into properties there. Thoughts?

Anyway, could we add a field proc:lineage to the proposal? Free-text string.

extensions/processing/README.md
```diff
@@ -0,0 +1,64 @@
...
+| --------------------------- | --------- | ------------------------------------------------------------ |
+| proc:name         | string | The name of the processing of the data that is commonly used. For example, `GRD Post Processing` for "GRD" product of Sentinel-1 satellites. |
+| proc:level        | string | **REQUIRED.** The name commonly used to refer to the processing level to make it easier to search for product level across collections or items. The short name must be used (only `L`, not `Level`). See the [list of suggested processing levels](#suggested-processing-levels). |
+| proc:facility     | string    | The name of the facility that produced the data. For example, `Copernicus S1 Core Ground Segment - DPA` for product of Sentinel-1 satellites. |
+| proc:software     | string    | The name of the software that produced the data. For example, `Sentinel-1 IPF` for the software that produces Sentinel-1 satellites data. |
+| proc:version      | string | The version of the software that produced the data. For example, `002.71` for the software that produces Sentinel-1 satellites data. |
```
m-mohr, line 28: I think I'd prefer proc:software_version to clarify the scope of the field.

Also, I'm wondering whether multiple software could be required? Then it might make sense to do something like:

```
proc:software: {
  "Sentinel-1 IPF": "002.71",
  "Other Tool": "1.1.1"
}
``` 
or something similar (could also be an array with objects).

extensions/processing/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Item and Collection fields
+
+For Items, the fields are placed in the properties. For Collections, the fields are placed in the summaries of the Collection.
+
+| Field Name                  | Type      | Description                                                  |
+| --------------------------- | --------- | ------------------------------------------------------------ |
+| proc:name         | string | The name of the processing of the data that is commonly used. For example, `GRD Post Processing` for "GRD" product of Sentinel-1 satellites. |
+| proc:level        | string | **REQUIRED.** The name commonly used to refer to the processing level to make it easier to search for product level across collections or items. The short name must be used (only `L`, not `Level`). See the [list of suggested processing levels](#suggested-processing-levels). |
+| proc:facility     | string    | The name of the facility that produced the data. For example, `Copernicus S1 Core Ground Segment - DPA` for product of Sentinel-1 satellites. |
```
m-mohr, line 26: What is the use case for this field?

extensions/processing/README.md
emmanuelmathot: Indeed, that's the common idea with this processing extension. So I totally see lineage could be part of this proposal. My only doubt is about having it at assets level. For us, an item is an atomic product even if split of several assets and thus should share the same mileage. For instance, a sentinel-2 product can have 13 bands that have all been processed during the same "lineage" and thus 13 assets all having the "proc:lineage": "Atmospheric Correction..." property would be overkill IMO.

extensions/processing/README.md
emmanuelmathot: Fine with the pair software "name:version"

extensions/processing/README.md
emmanuelmathot: If you mean the facility, it allows filtering product on their processing center origin. For instance, to select products produced by a specific ground segment components.

extensions/processing/README.md
emmanuelmathot: proc:name would be a label for the processing like an identifier but maybe a bit redundant if the `lineage` property is integrated. `sar:product_type`, which IMO should be a common metadata, is the family product identifier defined by the provider

m-mohr: Yes, we discussed whether product_type should be in common metadata. The reason not to include it is that we want to define a list of "common" values like in the SAR extension, but that is a bit hard in common metadata. I guess we'll have follow-up discussions.

Replacing proc:name with proc:lineage makes sense, I think.

extensions/processing/README.md
m-mohr: Your proposal was going to be mainly for properties in items, right? I think that is correct and I didn't meant to say something different here. You could put them in assets, like any other property, but I doubt providers will do that if not absolutely required. So I'm fine with it.

extensions/processing/README.md
m-mohr: Okay, so that's mainly for the "original" products, I guess? I must admit I wouldn't know what to put there if I process something at - for example - AWS or even locally. ;-)

extensions/processing/json-schema/schema.json
```diff
@@ -0,0 +1,90 @@
...
+    {
+      "allOf": [
+        {
+          "$ref": "../../../collection-spec/json-schema/collection.json"
+        },
+        {
+          "$ref": "#/definitions/stac_extensions"
+        },
+        {
+          "$ref": "#/definitions/scientific"
```
m-mohr, line 22: This seems to be a mistake...

extensions/processing/json-schema/schema.json
```diff
@@ -0,0 +1,90 @@
...
+      "required": [
+        "stac_extensions",
+        "properties"
+      ],
+      "properties": {
+        "stac_extensions": {
+          "type": "array",
+          "contains": {
+            "enum": [
+              "proc",
```
m-mohr, line 39: This must be `processing` (folder name)

extensions/processing/json-schema/schema.json
```diff
@@ -0,0 +1,90 @@
...
+              "https://schemas.stacspec.org/v1.0.0-beta.2/extensions/proc/json-schema/schema.json"
```
m-mohr, line 40:
```suggestion
              "https://schemas.stacspec.org/v1.0.0-beta.2/extensions/processing/json-schema/schema.json"
```

extensions/processing/json-schema/schema.json
```diff
@@ -0,0 +1,90 @@
...
+            ]
+          }
+        },
+        "properties": {
+          "type": "object",
+          "required": [
+            "proc:level"
```
m-mohr, line 47: Should proc:level be required or just any of the fields (i.e. `minProperties: 1`)?

m-mohr:
Once we found common ground, I'll likely also use this extension for #922

extensions/processing/README.md
cholmes: +1 for proc:lineage 

And defining it at the Item property level makes sense to me, and then people could choose to use it at the asset level if they really wanted.

Re-using proc:lineage at the provider level for Collections also makes sense to me.

extensions/processing/README.md
cholmes: +1 to software_version. I'm neutral on string vs array/object

extensions/processing/README.md
emmanuelmathot: I have replaced `processing:name` with processing:lineage` referring to https://wiki.earthdata.nasa.gov/display/NASAISO/Lineage+Information.
For the moment, I have left all processing properties at the same level extending to the asset. I don't know exactly how to specify the extension with providers.
Feel free to suggest other places where they could be specified. On my side, I am very attach to the formalism 1 item = 1 dataset but there are probably several other use cases.

extensions/processing/json-schema/schema.json
m-mohr: I see you also replaced the prefix in the field names. That was not my intention, but is also okay for me. The stac_extensions entry and the prefix don't need to match, but the stac_extensions entry needs to match the folder name.

extensions/processing/README.md
m-mohr: I can do the Collection work for you and also do the JSON Schema. :-)

m-mohr (approved):
@emmanuelmathot I've made a JSON Schema that includes the Collection providers and adapted the README and examples accordingly, also fixed some other inconsistencies. I hope it's fine for you that I directly made the changes. Could you please review whether the current state if okay for you? I'd be fine to merge it now.

matthewhanson:
This is great, thanks for doing this @emmanuelmathot , two comments:

I think this extension should at least mention using the `derived-from` and `via` links, they would complement the rest of the info.

I'm a little cautious about the the processing levels, because they are not standards, and as much as we've talked about it at the ARD conferences there's been no consensus.  I'm not opposed to STAC offering up some basic definitions, like we've done with spectral bands but in that case I think there might be too many levels in this table making it difficult to search on, since some data may not have some levels.   What about limiting table to the major processing levels, and not including the sub-levels that tend to be more provider specific?   i.e., L0, L1, L2, L3, L4.....I think most providers would at least agree on these.  To me the information contained in the sub-levels indicate detail about the radiometric and geometric processing which are probably better off as different fields.  For example, there are fields indicating terrain correction, atmospheric correction, etc that can be specified separately from gross processing level.

In fact, some of these fields may end up being better handled under an ARD extension describing what type of corrections have been applied to the data, whereas processing extension describes the software and versions used.

m-mohr:
@emmanuelmathot Do you have time to tackle Matts comments? I'd like to get this merged rather sooner than later ;-)

emmanuelmathot:
@matthewhanson thank you for the review. I added a paragraph recommending the usage of `derived-from` or `via` rel to trace the processing.
I also reduced the suggested processing levels to be more common. Nevertheless, I think the processing level field may be used to specify a finer level.

extensions/processing/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+*At least one of the fields must be specified.*
+
+### Processing Date Time
+
+The time of the processing is directly specified via the `created` properties of the target asset as specified in the [STAC Common metadata](https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#date-and-time)
+
+### Linking the items
+
+In items declaring this `processing` extension, it is recommended to add one or more [links](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#link-object) with `derived-fom` or `via` relationships to the eventual source items used in the processing. They could be used to trace back the processing history of the dataset.
```
m-mohr, line 38:
```suggestion
In items declaring this `processing` extension, it is recommended to add one or more [links](../../item-spec/item-spec.md#relation-types) with `derived_from` or `via` relationships to the eventual source items used in the processing. They could be used to trace back the processing history of the dataset.
```

m-mohr (approved):

extensions/processing/README.md
```diff
@@ -43,22 +47,17 @@ This list is not exhaustive and can be extended with the processing level specif
...
-| L2SP        | Surface reflectance and surface temperature science products | [Landsat Level 2 science products](https://www.usgs.gov/core-science-systems/nli/landsat/landsat-collection-2-level-2-science-products) |
 | L3          | Data or retrieved environmental variables which have been spatiallyand/or temporally re-sampled (i.e., derived from level 1 or 2 products). Such re-sampling may include averaging and compositing.  A wide variety of sub-level products are possible (see below). | [ENVISAT Level-3](http://envisat.esa.int/level3/), [Sentinel-2 L3](https://s2gm.sentinel-hub.com/) |
 | L4          | Model output or results from analyses of lower level data (i.e.,variables that are not directly measured by the instruments, but are derived fromthese measurements) |
 
+## Implementations
+
+[DotNetStac](https://github.com/Terradue/DotNetStac) uses this extension specification for the processing extension plugin with accessor to the value specified of this extension.
```
m-mohr, line 59: Depending on the outcomes in https://github.com/radiantearth/stac-spec/pull/920 we might ask you to (re)move this.

emmanuelmathot: Fine with me

cholmes: Though do you have catalogs produced with dotnetstac that use it? Even if they aren't public, you can describe it and say how large a catalog it is, or how many catalogs have been produced with DotNetStac. I think you could mention they are produced by DotNetStac and link to it, but the main implementation bullet point should be about the catalogs (also ok if none are produced yet / haven't yet been done with the latest changes, just add it in when they are).

cholmes (approved):

extensions/processing/README.md
emmanuelmathot: No public catalog so far. I will update as soon as we have more consistent examples. I remove the reference to DotNetStac to move forward with the merge. No problem with that.

m-mohr:
Merged, thank you @emmanuelmathot

cholmes:
Yes, thanks a lot @emmanuelmathot! This is a really great contributions. Apologies that we were slow on initially responding and getting it in, but thanks for sticking with it - I'm quite sure this will be a well used extension, and its addition will be a main highlight of our next release.

m-mohr:
I'm actually already planning to use it in my CARD4L stuff :-)

PR #917, opened 2020-11-14
SLD Examples

simonff:
**Related Issue(s):** 916

**Proposed Changes:**

Examples of porting SLD to a STAC extension.
The source of SLD examples under the "SourceUrl" key (which I don't intend to be used in the actual extension)

### DIFF
--- a/extensions/style/examples/mapbox-raster.json
+++ b/extensions/style/examples/mapbox-raster.json
@@ -1,0 +1,5 @@
+"mapbox:layers": [
+  {
+    "mapbox:type": "raster",
+  },
+], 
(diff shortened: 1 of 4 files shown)


m-mohr:
Thanks for going ahead and making a proposal. This seems to be basically a 1:1 mapping from XML to JSON, right? I'm wondering, why do we not just link to the XML?

simonff:
I'm okay with SLD being in an external XML file if everyone else is.

cholmes:
Yeah, I think if we're using SLD it'd make the most sense to just have it as an 'asset' in STAC and link to an external XML file that is actually SLD compliant, instead of porting to JSON. 

We may not even need an extension for this, could just add `application/vnd.ogc.sld+xml` to the table in https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#media-types

matthewhanson:
+1 linking to SLD

cholmes:
Closing this, as it should either be a new PR that includes SLD xml as an asset media type, or else a new extension.

PR #919, opened 2020-11-21
Modify valid range for sun_elevation to be -90-90.

### LINKED ISSUE #853: Negative view:sun_elevation value support (Night Shots)
The current spec has `view:sun_elevation` defined as:
The angle from the tangent of the scene center point to the sun. Measured from the horizon in degrees (0-90).

At Maxar Technologies negative values are used for night shots. Validation on the equivalent `view:sun_elevation` field is -90 to 90.

A negative sun elevation value helps users identify that the image is a night shot, which can help with filtering depending on use case, as well as potential radiometric algorithms that are sensitive to energy return intensity.

Based on the current `view:sun_elevation` validation, the current STAC spec doesn't support the idea of night shots without creating a custom extension and modifying the `view:sun_elevation` value in our STAC Items.

It seems like this would be a fairly common use case and would be worthy of a standardized way of handling it.

lossyrob:
Fixes #853

**Related Issue(s):** #853

**Proposed Changes:**

As mentioned in #853, a negative view angle can be validly recorded in remove sensing data. Besides the example listed by the issue author, I've hit up against this in ASTER L1T data, in which the [product specification](https://lpdaac.usgs.gov/documents/300/ASTER_L1T_Product_Specification.pdf) states values will range from -90 to 90, and a many of the images record such a value in the HDF metadata.

This change modifies the spec text and JSON schema to allow for sun_elevation values ranging from -90 to 90.

**PR Checklist:**

- [ ] ~~This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec), and I have opened issue/PR #XXX to track the change.~~

### DIFF
--- a/extensions/view/README.md
+++ b/extensions/view/README.md
@@ -20,7 +20,7 @@ This document explains the fields of the View Geometry Extension to a STAC Item.
 | view:incidence_angle | number               | The incidence angle is the angle between the vertical (normal) to the intercepting surface and the line of sight back to the satellite at the scene center. Measured in degrees (0-90). |
 | view:azimuth         | number               | Viewing azimuth angle. The angle measured from the sub-satellite point (point on the ground below the platform) between the scene center and true north. Measured clockwise from north in degrees (0-360). |
 | view:sun_azimuth     | number               | Sun azimuth angle. From the scene center point on the ground, this is the angle between truth north and the sun. Measured clockwise in degrees (0-360). |
-| view:sun_elevation   | number               | Sun elevation angle. The angle from the tangent of the scene center point to the sun. Measured from the horizon in degrees (-90-90). |
+| view:sun_elevation   | number               | Sun elevation angle. The angle from the tangent of the scene center point to the sun. Measured from the horizon in degrees (`-90`-`90`). Negative values indicate the sun is below the horizon, e.g. this data was captured at night. |
 
 *At least one of the fields must be specified.*
 
--- a/extensions/view/README.md
+++ b/extensions/view/README.md
@@ -20,7 +20,7 @@ This document explains the fields of the View Geometry Extension to a STAC Item.
 | view:incidence_angle | number               | The incidence angle is the angle between the vertical (normal) to the intercepting surface and the line of sight back to the satellite at the scene center. Measured in degrees (0-90). |
 | view:azimuth         | number               | Viewing azimuth angle. The angle measured from the sub-satellite point (point on the ground below the platform) between the scene center and true north. Measured clockwise from north in degrees (0-360). |
 | view:sun_azimuth     | number               | Sun azimuth angle. From the scene center point on the ground, this is the angle between truth north and the sun. Measured clockwise in degrees (0-360). |
-| view:sun_elevation   | number               | Sun elevation angle. The angle from the tangent of the scene center point to the sun. Measured from the horizon in degrees (0-90). |
+| view:sun_elevation   | number               | Sun elevation angle. The angle from the tangent of the scene center point to the sun. Measured from the horizon in degrees (-90-90). |
 
 *At least one of the fields must be specified.*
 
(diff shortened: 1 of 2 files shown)


matthewhanson:
+1 on this, it's come up several times. Most users will consider the absolute value, but the +/- will indicate port/starboard depending on platform.

matthewhanson (approved):

extensions/view/README.md
```diff
@@ -19,7 +19,7 @@ This document explains the fields of the View Geometry Extension to a STAC Item.
 | view:incidence_angle | number               | The incidence angle is the angle between the vertical (normal) to the intercepting surface and the line of sight back to the satellite at the scene center. Measured in degrees (0-90). |
 | view:azimuth         | number               | Viewing azimuth angle. The angle measured from the sub-satellite point (point on the ground below the platform) between the scene center and true north. Measured clockwise from north in degrees (0-360). |
 | view:sun_azimuth     | number               | Sun azimuth angle. From the scene center point on the ground, this is the angle between truth north and the sun. Measured clockwise in degrees (0-360). |
-| view:sun_elevation   | number               | Sun elevation angle. The angle from the tangent of the scene center point to the sun. Measured from the horizon in degrees (0-90). |
+| view:sun_elevation   | number               | Sun elevation angle. The angle from the tangent of the scene center point to the sun. Measured from the horizon in degrees (-90-90). |
```
m-mohr, line 22: The -90-90 looks a bit strange, maybe put spaces?

```suggestion
| view:sun_elevation   | number               | Sun elevation angle. The angle from the tangent of the scene center point to the sun. Measured from the horizon in degrees (-90 - 90). |
```

m-mohr:
I agree with @cholmes https://github.com/radiantearth/stac-spec/issues/853#issuecomment-645585134 that it feels like it is overloading the view angle as it may not be intuitively clear what a negative value could mean and maybe it should be considered go with an (additional?) and more general `view:night: true/false` or so. But as this seems to be used across multiple providers, it might also just be a best practice that has been established for night shots to just use negative values and then I think it's fine. 🤷‍♀️

m-mohr (changes requested):
A bit more explanation on what negative values mean should be added.

cholmes:
I'm ok with the negative value, since it looks like you're seeing it across multiple providers @lossyrob but I agree with Matthias that we should add a bit of explanation in the table that a negative value means it was taken at night.

matthewhanson:
Not sure why I didn't remember this before, but I ended up re-implementing the solar angle calculation in the processing of ASTER L1 for LPDAAC a few years ago in Cumulus, so I'm familiar with this particular case. It was using old FORTRAN code that was last updated in 1997, and I switched to using [Skyfield](https://rhodesmill.org/skyfield/).

But to the point...

A negative sun elevation does not necessarily mean "night", or at least it depends on your definition of "night".  The most accepted definitions are:
image


ASTER has a day/night flag, and it defines night as astronomical twilight, so 12 degrees below the horizon.

cholmes:
@lossyrob - can you add a bit more explanation on negative values? What they mean? Doesn't have to be a lot. If not I can take a crack at it, but it'd be good to get this merged in.

lossyrob:
@cholmes thanks for the ping. @m-mohr does the additional text work for additional explanation?

m-mohr (approved):
I'm neutral on this change and can't say whether it's a good solution or not, but I have to approve after requesting changes. So better get another opinion on this one. Was @matthewhanson's comment considered?

matthewhanson:
I think this is a necessary change but think we should expand on "this data was captured at night" because "night" isn't really a boolean by itself.  Suggest updating text to be:

Negative values indicate the sun is below the horizon, e.g. sun elevation of -10&deg; means the data was captured during [nautical twilight or dawn](https://www.timeanddate.com/astronomy/different-types-twilight.html).

Including the link provides users a simple and concise explanation of night vs day.

lossyrob:
@matthewhanson included your example and the link, if good by you then merge away!

cholmes (approved):

PR #920, opened 2020-11-22
Add implementation links to pystac for sat, scientific, and version.

schwehr:
**Related Issue(s):** #

None

**Proposed Changes:**

1. Added implementation links for the sat, scientific, and version extensions in pystac

**PR Checklist:**

  - I don't think this warrants a CHANGELOG entry, but I can add one if folks thing there should be one 
- I only added the extensions that I implemented.  I can add links for the rest of the pystac extensions if folks want.

### DIFF
--- a/extensions/sat/README.md
+++ b/extensions/sat/README.md
@@ -28,7 +28,7 @@ The Satellite extension requires the [Instrument Fields](../../item-spec/common-
 
 ## Implementations
 
-- No implementations yet
+- [pystac](https://github.com/stac-utils/pystac/blob/develop/pystac/extensions/sat.py)
 
 ## Extensions
 
--- a/extensions/scientific/README.md
+++ b/extensions/scientific/README.md
@@ -60,4 +60,4 @@ This extension adds the following types as applicable `rel` types for the [Link 
 
 ## Implementations
 
-None yet, still in proposal stage.
+- [pystac](https://github.com/stac-utils/pystac/blob/develop/pystac/extensions/scientific.py)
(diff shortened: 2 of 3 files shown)


m-mohr:
More a general question @cholmes @matthewhanson , but wasn't the list of implementations meant to include catalogs exposing the fields (i.e. example catalogs) so that we get details on the maturity? But it seems at the moment people only add software libraries so we may want to reconsider this. What is meant to be listed here? For example, the validators implement all extensions, but it doesn't seem to be very valuable to add them here.

lossyrob:
I agree a better use of the "Implementations" section would be to point to catalogs that use the extension, rather than tools that include a feature for that specific extension.

schwehr:
Thanks for the clarification

cholmes:
Yes, intent was to be the actual catalogs using it. Though I do think it would be good to also list the software that supports a given extension, so users can know that they can use X software to help with the extension, and it's also an indication of maturity.

matthewhanson:
Weren't we going to move Implementations out of the spec and onto the stac website? That gives us a way to keep a live version without having to make a release or look at the dev branch.

Also, maybe stacindex is the better place to store this info (for catalogs). Using the extensions advertised it could be automatically generated and used to generate implementation lists for each extension. That way it wouldn't be reliant on someone adding this information.

m-mohr:
Yes, I agree. We should try to leave this out of the spec to limit changes and friction here.

m-mohr:
Closing for now as I've started implementing that [API] extensions can be assigned to each ecosystem entry on stacindex.org. I still need to support modifying existing entries, but we'll get there another day. Until then I'm happy to change it manually, just send me an e-mail with a full list of supported STAC extensions / STAC API extensions.

Issue #921, opened 2020-11-23
[Item] Asset size property (content length)


emmanuelmathot:
## Summary

Having a property to indicate the size of the asset

## Context

It is sometimes useful to know the size of the assets referenced in the STAC item. For instance, to provision disk space or to avoid downloading the same asset doing a simple content length comparison.

## Proposal

2 options

1. An additional [STAC Common Metadata](https://github.com/radiantearth/stac-spec/blob/master/item-spec/common-metadata.md#stac-common-metadata) with typical name such as `size` or `content_length`
2. An dedicated or existing extension

m-mohr:
Sounds a bit related to the Checksum extension, I think it should start as an extension. I agree that this field can be useful for clients and users.

matthewhanson:
The `checksum` is more useful for determining if something is a duplicate, I think `size` can be useful though.

I don't think there's an existing extension that would fit here, unless we combine it with checksum and rename.

It's not quite Common Metadata either, because that implies it can be an Item property (or an asset field), but this would be exclusive to assets (and possibly links), like the checksum extension.

cholmes:
I agree on starting as an extension. It does seem to group well with checksum, in an 'file info' extension or something? But if checksum is in use and we don't want to change its extension then it could just be a small dedicated extension.

matthewhanson:
We could also create a new file-info extension that includes checksums and deprecate the existing checksum extension.

m-mohr:
I've not seen it used very widely in public, maybe 1 or 2 collections max. I should finally start crawling collections so that we have statistics for such things...

m-mohr:
Writing up the CARD4L extension, I'm wondering whether there's more than just size and checksum for a fileinfo extension.
For example, I could think of
* `file:size`
* `file:data_type` (currently proposed in #879 as render:data_type)
* `file:checksum` (currently `checksum:multihash`)
* `file:byte_order`
* maybe: `file:name` (avoid parsing it from the path?)
* maybe: `file:compressed` or `file:archive` (for asstes in ZIP, TAR and other archives as recently discussed on Gitter, but that can probably also be figured out by the media `type`)

m-mohr:
@emmanuelmathot (and all others reading this) Good that you like my comment ;-) Which fields would be useful for you? I'll probably write this up and I'm wondering what is actually useful for others. For example byte_order I could also define in card4l directly if that is too specific. Please name any of the fields mentioned or even additional ones. :-) Thank you!

[...]

PR #934, opened 2020-12-16
Add file info extension (remove checksum extension)

m-mohr:
**Related Issue(s):** #879 #921

**Proposed Changes:**

1. Add file info extension with 5 fields, see #921
2. Remove checksum extension (is now part of the file info extension)

cc for review: @geospatial-jeff @kylebarron @emmanuelmathot @vincentsarago

**PR Checklist:**


extensions/file/README.md
```diff
@@ -0,0 +1,63 @@
...
+- `uint8`: unsigned 8-bit integer (common for 8-bit RGB PNG's)
+- `unit16`: unsigned 16-bit integer
+- `uint32`: unsigned 32-bit integer
+- `uint64`: unsigned 64-bit integer
+- `float32`: 32-bit float
+- `float64`: 64-big float
+- `cint16`: 16-bit complex integer
+- `cint32`: 32-bit complex integer
+- `cfloat32`: 32-bit complex float
+- `cfloat64`: 64-bit complex float
```
jisantuc, line 48: Some background on where these come from could be useful -- they're obvious reference points for rasterio / gdal users, but if I'm working with a frontend-only client and don't tend to work much with the underlying data I think these would look kind of out-of-nowhere.

extensions/file/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Checksums
+
+`file:checksum` was previously defined in the `checksum` extension and the field name was `checksum:multihash` before STAC v1.0.0-beta.3. The specification of the field has not changed.
```
jisantuc, line 52: Since the checksum extension has been removed, should that specification be copied here?

extensions/file/README.md
m-mohr: Took it from #923, not sure where people there took it from. I added (u)int64 though.

extensions/file/README.md
m-mohr: It is the same spec ("The specification of the field has not changed."), so mostly everything from the old extension was ported over here. It's just a different field name. Not sure what else I should add?

jisantuc: Oh I didn't expand the diff on the deletion and assumed a link to multihash docs would be included there. I think a link to those docs would be helpful.

m-mohr: A link to multihash is above in the field description, but I can surely add another instance here and also link to the old tagged checksum extension.

jisantuc: I'm swearing off PR reviews before coffee #2. sorry for the noise.

jisantuc (approved):

extensions/file/README.md
m-mohr: Added additional links :-)

extensions/file/README.md
cholmes: Did you mean you took it from #879?

We had a discussion on gitter about what to do, generally its taken from rasterio/numpy. 

I agree we should give more background - I can try to take a crack at explaining it a bit more, but this is not my area of expertise, so will have to re-read the conversations to do so. And won't get to it super fast, so if anyone else wants to take a shot then go for it.

m-mohr: Yes, indeed. I felt it might be a bit too restrictive so I added int64/uin64/float16, but still there might be additional types, like you find references to int128 online for example and also additions to complex ints/floats. But I'm also no expert in this...

m-mohr: We could also go with `^(u|c)?(int|float)(8|16|32|64|128)$` with u = unsigned, c = complex and the number being the bits (but then there's likely no float8) ;-)

Anyway, I'm not sure what additional context you'd like to see here? There's likely no table that lists (standardizes) all of them. One could try to align with IEEE 754, but the names seem to be completely different than what is often used in tooling?!

cholmes: I think the context is just a couple lines on what these are used for - why a client might want this information, why providers should consider including it.

extensions/file/README.md
```diff
@@ -0,0 +1,66 @@
...
+### Data Types
+
+The allowed values for `file:data_type` are:
+
+- `int8`: 8-bit integer
+- `int16`: 16-bit integer
+- `int32`: 32-bit integer
+- `int64`: 64-bit integer
+- `uint8`: unsigned 8-bit integer (common for 8-bit RGB PNG's)
+- `unit16`: unsigned 16-bit integer
```
kylebarron, line 42:
```suggestion
- `uint16`: unsigned 16-bit integer
```

kylebarron (approved):

extensions/file/README.md
m-mohr: Thank you, included it in my last commit!

extensions/file/README.md
m-mohr: I've added some context I could think of. Also added `other` for any files that may include other data types (e.g. larger number, string, boolean/bit). Could you check that again?

extensions/file/README.md
```diff
@@ -32,14 +32,23 @@ Please be aware that the integer values (always unsigned) given for the sizes (e
 
 ### Data Types
 
+The data type gives information about the values in the file.
+This can be used to indicate the (maximum) range of numerical values expected.
+For example `unit8` indicates that the numbers are in a range between 0 and 255,
```
kylebarron, line 37:
```suggestion
For example `uint8` indicates that the numbers are in a range between 0 and 255, 
```

extensions/file/README.md
```diff
@@ -32,14 +32,23 @@ Please be aware that the integer values (always unsigned) given for the sizes (e
 
 ### Data Types
 
+The data type gives information about the values in the file.
+This can be used to indicate the (maximum) range of numerical values expected.
+For example `unit8` indicates that the numbers are in a range between 0 and 255, 
+they can never be smaller or larger. This can help to pick the optimal numerical
+data type when reading the files to keep memory consumption low.
+Nevertheless, it doesn't necessarily mean that the expected values fill the whole range.
+For example, there can be use cases for `unit8` that just use the numbers 0 to 10 for example.
```
kylebarron, line 41:
```suggestion
For example, there can be use cases for `uint8` that just use the numbers 0 to 10 for example.
```

kylebarron (approved):
:smile:

extensions/file/README.md
m-mohr: LOL - "unit" is what I type more often it seems. Good catch!

cholmes (approved):
Looks great!

m-mohr:
Great. Merging for now, but let me know if you find additional things to change/improve/add. Happy to make further PRs.

PR #935, opened 2020-12-18
clarify the working around Label requiring an asset named label

philvarner:
**Related Issue(s):** n/a

**Proposed Changes:**

1. clarify the wording the `labels` asset, to mean exactly one

**PR Checklist:**


extensions/label/README.md
```diff
@@ -132,7 +132,8 @@ Some additional notes are given here for some of the core STAC Item fields and w
 #### Assets
 
 ##### labels (required)
-The Label Extension requires at least one asset that uses the key "labels". The asset will contain a link to the actual label data. The asset has these requirements:
+
+The Label Extension requires an asset with the key "labels". The asset will contain a link to the actual label data. This asset has these requirements:
```
m-mohr, line 136: A general question: Why is this using the asset key instead of asset roles?

vpipkt:
A comment related to this PR, the spacenet roads example [does not follow this requirement](https://github.com/radiantearth/stac-spec/blob/master/extensions/label/examples/spacenet-roads/roads_item.json#L162-L163).

And a broader question, will the specification have some means to accomodate more than one `assets`. In particular, how to accomodate the pattern recommended [for a raster asset and associated geojson asset](https://github.com/radiantearth/stac-spec/blame/master/extensions/label/README.md#L142)?

m-mohr:
@vpipkt That's why I was asking about the asset roles above. It seems like the better way to make this work.

philvarner:
I agree with these issues. I filed https://github.com/radiantearth/stac-spec/issues/936 to push this change forward. 

Leaving this PR as simply clarifying the existing language. Maybe it doesn't matter now?

jisantuc (approved):
I think this is an improvement in clarity over what's currently there -- apparently duplicate keys in an object had an undefined behavior [as of 2014](https://stackoverflow.com/questions/21832701/does-json-syntax-allow-duplicate-keys-in-an-object)? TIL! Anyway, it's good not to encourage the duplicate keys -- we definitely haven't considered what should happen with them in `stac4s`, and what happens with them (keep the last one) seems less than ideal:

```scala
@ case class Foo(x: Int)
defined class Foo

@ decode[Foo]("""{"x": 3, "x": 4}""")
res10: Either[Error, Foo] = Right(Foo(4))
```

cholmes (approved):

philvarner:
> I think this is an improvement in clarity over what's currently there -- apparently duplicate keys in an object had an undefined behavior [as of 2014](https://stackoverflow.com/questions/21832701/does-json-syntax-allow-duplicate-keys-in-an-object)? TIL! Anyway, it's good not to encourage the duplicate keys -- we definitely haven't considered what should happen with them in `stac4s`, and what happens with them (keep the last one) seems less than ideal:
> 
> ```scala
> @ case class Foo(x: Int)
> defined class Foo
> 
> @ decode[Foo]("""{"x": 3, "x": 4}""")
> res10: Either[Error, Foo] = Right(Foo(4))
> ```

apparently, that's configurable in the jawn CirceSupportParser -- `allowDuplicateKeys = false` throws an exception if there are duplicates.

Issue #903, opened 2020-09-21
Make summaries available in catalog as well as collection


cholmes:
Having the summaries back in 'catalog' will allow it to be more used to aid in navigation, and since all collections are catalogs it will still work there.

m-mohr:
Wait, you mean summaries, right?

cholmes:
yes :(

cholmes:
I get confused because it's the 'stats object'. but the 'summary' field.

matthewhanson:
+1

PR #942, opened 2021-01-15
Summaries in catalogs

cholmes:
**Related Issue(s):** #903 #822

**Proposed Changes:**

1. Moved summaries from collection spec to catalog spec
2.  Had to do a minor bit on #822 so that I could link easily between the two. 

**PR Checklist:**


m-mohr:
I'm not a fan of this at all and won't approve. For me summaries is fundamental for collections and catalogs is just a "side" use case that no body has ever implemented yet. Thus, I'm not a fan to make it less visible in Collections and hide important details about summaries in Catalogs.

cholmes:
Ok. I don't feel strongly, just was going for not repeating things. I agree collections is the 'main' use case.

What about having the text in collections and linking to that from catalogs? So it appears in the primary place, but catalogs has a line that mentions summaries can be used there? Like just flip it around from how I proposed?

m-mohr:
That would work for me.

cholmes:
Ok, this should be ready to go. Please approve if it works.

matthewhanson (approved):

m-mohr (approved):
LGTM

PR #943, opened 2021-01-19
Avoid duplicate band in example

aaime:
The bands contain "band1" twice, seems like a copy/paste error?

**PR Checklist:**

- [ ] This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec), and I have opened issue/PR #XXX to track the change (did not yet, it's a minor change in an example file)

### DIFF
--- a/item-spec/examples/sample-full.json
+++ b/item-spec/examples/sample-full.json
@@ -46,9 +46,6 @@
         "name": "band1"
       },
       {
-        "name": "band1"
-      },
-      {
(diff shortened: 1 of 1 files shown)


m-mohr:
Thank you for the bug fix. Unfortunately, all changes must be made against dev instead of master.

aaime:
Ah, the bands have changed name there, and show a rgb + NIR case. Ok,never mind, closoing.

Issue #918, opened 2020-11-16
MIME types for STAC and for things like Earth Engine?


schwehr:
I'm wondering about mimetypes. Should there be one for STAC?  e.g. `application/stac+json`?

And what to do for EarthEngine FeatureCollections, Images and ImageCollections. These things don't return specific bytes.  The API is JSON, so something like  `application/vnd.ee.image+json`, `application/vnd.ee.imagecollection+json`, `application/vnd.ee.featurecollection+json` Although ImageCollection might be explicitly forbidden as a collection of things, but I don't think so as zip, tar, etc. are ok.

References:

- https://www.iana.org/assignments/media-types/media-types.xml has lots of things that are `+json`.
- https://en.wikipedia.org/wiki/Media_type 
- https://tools.ietf.org/html/rfc6838

m-mohr:
Is that for Items, Collections or Catalogs? Would we need multiple media types

schwehr:
I would think for all of the file types.  I'm not sure if they would need to be multiple types.  Some of the already registered mime types cover vast swaths of functionality. 

I reached out to [George Percivall](https://www.ogc.org/ogc/organization/staff/gpercivall) at OGC for thoughts (links added by me):

> [OGC API - Records group](https://www.ogc.org/projects/groups/apirecordsswg) handles much of that discussion.  They are meeting.  Peter Vretanos chairs that WG.
> Also, OGC is registered as a MIME authority and that discussion happens in the [OGC Naming Authority group](  https://www.ogc.org/ogcna).

cholmes:
+1 to doing this - I just hit a potential need for it myself. From gitter @ycespb said:

 > For your info, we also need a dedicate MIME type to distinguish from other OGC representations. We are currently using "type": "application/geo+json;profile=\"https://stacspec.org\"" for STAC representations. See for instance at https://ergo.spacebel.be/collections/datasets/items/PR1_OPER_CHR_MO1_1P_20131119T035100_S25-016_E031-023_0001?httpAccept=application/geo%2Bjson;profile=%22https%3A%2F%2Fstacspec.org%22.

That seems sensible to me - communicating that it is GeoJSON (for Items), and adding a bit more. Not sure what we'd do for collections in that case though.

Eventually we'll be in OGC and can use their naming authority.

m-mohr:
+1 for media type

So GeoTiff got `image/tiff; application=geotiff; profile=cloud-optimized`

[...]

Issue #851, opened 2020-06-17
Profiled Media Type (MIME) for STAC dcouments


emmanuelmathot:
# Summary

Need for a specific media type definition that specifies modeling level for Item, Catalog and Collection.

# Context

Exposing a Web Service (OGC API Processes) with STAC documents as input. The parameter declares the media type (MIME) of expected input. The current implementation uses `application/json` and `application/geo+json`. This syntactical level of specification is necessary but insufficient to indicate that the expected JSON document must follow STAC specification.

# Proposal

A parameter can be defined as additional semantics that can be used to process a resource representation, such as constraints, conventions, extensions, or any other aspects that do not alter the basic media type semantics, which covers our use cases of structures and models. Concretely, following https://tools.ietf.org/html/rfc2046 we would define a profile extension to the media-type to specify the profile of the JSON:
- `application/geo+json; profile=stac-item`
- `application/json; profile=stac-collection`
- `application/json; profile=stac-catalogue`

# Alternatives

Another option is to defined a vendor-specific mime-type such as "application/stac-item+json", "application/stac-collection+json", "application/stac-catalog+json"

Let me know the best practice or any eventual specifications
Thanks

m-mohr:
Some remarks:
1. Implicitly, the media type for STAC Items is currently `application/geo+json` and for Catalogs/Collections is `application/json`. Introducing these new media types would have influence on existing STACs as the links would need a different media types in `type`. Alternative would be to allow new media types with and without profile.
2. This would also influence work on #848 likely other things that built on top of STAC.
3. Do implementations actually parse media types so that they may just ignore the profile if set? I doubt it.

emmanuelmathot:
Some answers:

1. The proposal would not change the media type, it remais `application/geo+json` and `application/json`. Alternatives do but we prefer not doing that. `profile` is a parameter in the content-type header or media-type and is not mandatory.
2. Since the core media-type remains the same, there is no impact. The problem resides in the software implementation. See 3.
3. Indeed, it all depends on the software implementation. Most of the software libraries parses properly a mime-type separating the core media-type from the parameters. They must often handle `charset` or `boundary` parameter so it should not be a problem if the parsing follows convention.

So in summary yes, the proposal is just about fixing a common convention for the `profile` parameter if needed.

m-mohr:
2. Have you read into the linked PR? Doesn't seem like that, because it impact on it (likely in a positive way).
[...]

Issue #930, opened 2020-12-11
Common Media Types


cholmes:
Each spec lists a number of media types. We should consolidate them into a single list that each spec can reference. This came up in #884

cholmes:
So Media Types are actually just listed in the Item spec. So I think I'm going to close this, since we're not repeating here. If the item spec was way too long I could see putting it in another file, but I think it's nicer to not have to jump around a lot. I do think #929 for rel types is one that should be consolidated. I suppose I could be ok with a file that is 'common types', and has the rels and the media in one file. @m-mohr - let me know, I don't feel strongly on this one.

cholmes:
@m-mohr and I discussed on a call. We decided to move most of the longer table to 'best practices', as the spec should just link to IANA, it doesn't need to list what is commonly used. The table should have information on the 'special' STAC ones, like the STAC-specific ones, and cloud optimized geotiff and geotiff, at least until they are fully standardized / in IANA.

cholmes:
Whoops, I thought the above had merged.

Issue #889, opened 2020-08-26
Explicit STAC Types in JSON [was] Ability to identify STAC file types with tools like file


schwehr:
Probably not something the average user is going to care about, but in exploring STAC files on my local machine, I find it hard sometimes to know which type of STAC file I'm looking at.  Or if I'm trying to find the catalog, collection, or item, it might be hard to know from the command line.  e.g. from Earth Engine:

```
file catalog.json MODIS_[MN]*
catalog.json:                            JSON data
MODIS_MCD43A4_006_BAI.json:              HTML document, ASCII text, with very long lines
MODIS_MCD43A4_006_EVI.json:              HTML document, ASCII text, with very long lines
MODIS_MCD43A4_006_NDSI.json:             HTML document, ASCII text, with very long lines
MODIS_MCD43A4_006_NDVI.json:             ASCII text
MODIS_MCD43A4_006_NDWI.json:             HTML document, ASCII text, with very long lines
MODIS_MOD09GA_006_BAI.json:              HTML document, ASCII text, with very long lines
MODIS_MOD09GA_006_EVI.json:              HTML document, ASCII text, with very long lines
MODIS_MOD09GA_006_NDSI.json:             HTML document, ASCII text, with very long lines
MODIS_MOD09GA_006_NDVI.json:             ASCII text
MODIS_MOD09GA_006_NDWI.json:             HTML document, ASCII text, with very long lines
MODIS_MOD13A1.json:                      HTML document, ASCII text, with very long lines
MODIS_MOD44W_MOD44W_005_2000_02_24.json: JSON data
MODIS_MYD09GA_006_BAI.json:              HTML document, ASCII text, with very long lines
MODIS_MYD09GA_006_EVI.json:              HTML document, ASCII text, with very long lines
MODIS_MYD09GA_006_NDSI.json:             HTML document, ASCII text, with very long lines
MODIS_MYD09GA_006_NDVI.json:             ASCII text
MODIS_MYD09GA_006_NDWI.json:             HTML document, ASCII text, with very long lines
MODIS_MYD13A1.json:                      HTML document, ASCII text, with very long lines
MODIS_NTSG_MOD16A2_105.json:             JSON data
```

This is a bummer!  It would be awesome if file was able to return `STAC`, the version, and one of {Catalog, `Collection`, `Item`}.  There might be a better way, but this could possibly work for file:

* Mandate `stac_version` be 1st
* Mandate `stac_type` (or some similar name) be 2nd

[...]

PR #944, opened 2021-01-20
Add media type

cholmes:
**Related Issue(s):** #851 #918 

**Proposed Changes:**

1. Added media types: `application/geo+json; profile=stac-item` for Items, `application/json; profile=stac-catalog` for Catalogs, and `application/json; profile=stac-collection` for Collection.

**PR Checklist:**

- [x] This PR has **no** breaking changes. (this is arguable, but an additional 'profile' on media types already used isn't supposed to break anything)
- [x] This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec), and I have opened issue/PR https://github.com/radiantearth/stac-api-spec/issues/100 to track the change.


catalog-spec/catalog-spec.md
```diff
@@ -71,6 +71,13 @@ The following types are commonly used as `rel` types in the Link Object of a STA
 
 **Note:** A link to at least one `item` or `child` catalog is **REQUIRED**.
 
+## Media Type for STAC Catalogs
+
+A STAC Catalog is a JSON file ([RFC 8259](https://tools.ietf.org/html/rfc8259)), and thus should use the 
+`[application/json](https://tools.ietf.org/html/rfc8259#section-11)` as the [Media Type](https://en.wikipedia.org/wiki/Media_type) 
+(previously known as the MIME Type). STAC Catalogs are recommended to additional specify a `profile` equal to `stac-catalog` to
```
m-mohr, line 78: In media type terminology `profile` is a parameter. Applies to all three types.

```suggestion
(previously known as the MIME Type). STAC Catalogs are recommended to additionally add the parameter `profile` with the value `stac-catalog` to
```

m-mohr:
There are tables containing media types in the specs, which should add the new media types, too.

cholmes:
Will add them. I had been thinking the lists were asset focused, but looks like they're written generically.

cholmes:
Ok, I think this is fully ready for review. @m-mohr - I reworked the media type section a bit, adding the stac types. It was also focused on 'assets', and under the asset section, which was a bit weird since catalog and collection linked to it directly in their link object descriptions, which didn't make much sense. So now it's pulled out to be a general media type section. I think it reads a bit better, but happy for any feedback / tweaks. Also updated all of them with your language on the parameter - thanks.

item-spec/item-spec.md
```diff
@@ -252,6 +218,53 @@ in the file).
 For examples of fields that this construct is recommended for, see the [section of STAC Best Practices](../best-practices.md#common-use-cases-of-additional-fields-for-assets)
 that talks about common use cases of additional fields for assets.
 
+## Media Types
+
+One of the best ways to help inform web clients about the content in a link is to use a common [Media 
+Type](https://en.wikipedia.org/wiki/Media_type) in the `type` field. In STAC the `type` field is used in both the 
+`[link](#link-object)` and `[asset](asset-object)` objects. It is quite useful for STAC browsers to better determine
```
m-mohr, line 225:
```suggestion
`[link](#link-object)` and `[asset](#asset-object)` objects. It is quite useful for STAC browsers to better determine
```

item-spec/item-spec.md
```diff
@@ -252,6 +218,53 @@ in the file).
 For examples of fields that this construct is recommended for, see the [section of STAC Best Practices](../best-practices.md#common-use-cases-of-additional-fields-for-assets)
 that talks about common use cases of additional fields for assets.
 
+## Media Types
```
m-mohr, line 221: Wasn't there an issue to move this to a separate page anyway? This would be a good chance to move it...

cholmes: There was - #930 I closed it since there was no real consolidation to do, and I don't love specs where you have to keep looking at other pages...

m-mohr: I still think it would make sense as it's a central part of all three specs and not really bound to items as you've realized yourself. Same for the rel types. At the moment you have to look at other pages anyway for Catalog and Collection, so it's only different for items.

cholmes: Ok, could you edit it in the way you want? Feel free to make commits on this branch directly. I wasn't sure where/how to make a shared page between the specs, but it seems like you have a way you want it done.

m-mohr: No clear idea yet to be honest. Could be similar to common metadata? Or should me create for everything common (rel types, media types, common metadata)?

cholmes: Ok, just made the changes as discussed on the call. Moved most of the table to 'best practices', and kept just our custom (stac and cog) ones in this table.

cholmes:
Ok, just pushed some more changes, to put the list of types in best practices. I think this one is ready for final review now.

best-practices.md
```diff
@@ -156,6 +157,38 @@ than the overall best resolution.
 - `sar:polarizations` ([sar extension](extensions/sar/)): Provide the polarization content and ordering of a specific asset, similar to `eo:bands`.
 - `sar:product_type` ([sar extension](extensions/sar/)): If mixing multiple product types within a single Item, this can be used to specify the product_type for each asset.
 
+## Working with Media Types
```
m-mohr, line 160: The headings are a bit strange. The h2 has no content and the h3 is the only header in the h2. This is discouraged in writing...

item-spec/item-spec.md
```diff
@@ -252,6 +218,48 @@ in the file).
...
+
+One of the best ways to help inform web clients about the content in a link is to use a common [Media 
+Type](https://en.wikipedia.org/wiki/Media_type) in the `type` field. In STAC the `type` field is used in both the 
+`[Link](#link-object)` and `[Asset](#asset-object)` Objects. It is quite useful for STAC browsers to better determine
+what to render and display to users searching and browsing the catalog.  Media types are often referred to by the 
+now-deprecated term "MIME types".
```
m-mohr, line 227: Not sure, is that a typo?
```suggestion
now deprecated term "MIME types". 
```

item-spec/item-spec.md
```diff
@@ -252,6 +218,48 @@ in the file).
...
+| `application/json; profile=stac-collection`             | A STAC [Collection](../collection-spec/README.md)            |
+| `image/tiff; application=geotiff`                       | GeoTIFF with standardized georeferencing metadata            |
+| `image/tiff; application=geotiff; profile=cloud-optimized` | [Cloud Optimized GeoTIFF](https://www.cogeo.org/) (unofficial). Once there is an [official media type](http://osgeo-org.1560.x6.nabble.com/Media-type-tc5411498.html) it will be added and the proprietary media type here will be deprecated. |
+
+Deprecation notice: GeoTiff previously used the media type `image/vnd.stac.geotiff` and
+Cloud Optimized GeoTiffs used `image/vnd.stac.geotiff; profile=cloud-optimized`.
+Both can still appear in old catalogues, but are deprecated and should be replaced. This may also shift in the future as
+[OGC sorts out the media types](https://github.com/opengeospatial/geotiff/issues/34).
+
+### Media Type for STAC Item
```
m-mohr, line 256: This chapter seems superfluous with the table above.

item-spec/item-spec.md
```diff
@@ -252,6 +218,48 @@ in the file).
...
+would be appropriate; if it is an XML, then `text/xml` is appropriate. For more information on media types as well as a 
+list of [common media types](../best-practices.md#common-media-types-in-stac) used in STAC see the [best practice on 
+working with media types](../best-practices.md#working-with-media-types).
+
+### STAC Media Types
+
+The following table lists the Media Types for STAC, as well as the way [GeoTIFF](https://en.wikipedia.org/wiki/GeoTIFF)'s 
+(and [COG](https://www.cogeo.org)'s) are referenced in STAC.
+
+| Media Type                                              | Description                                                  |
```
m-mohr, line 243: It's a bit strange that the Item Spec has the media types for all "specs", but Collections and Catalogs, where you'd need them first in the hierarchy, have no such table. So I don't think this is the right place to put such a table. This is better suited in catalog actually as there you link to items and catalogs (and collection inherits from there). In Items you don't link to items and catalogs at all.

item-spec/item-spec.md
kylebarron: I don't think it's a typo because both hyphenated words are describing `term`, but I wasn't an English major 😄

m-mohr: I'm not even a native speaker, just haven't seen it yet with a hypen so was confused.

kylebarron: Well I was trying to say that even native speakers have a hard time knowing when to use hyphens correctly 😄

best-practices.md
cholmes: Ah, good catch. I meant to put in more information there on actual practices around using media types, but looks like I forgot to come back and do that. Will work on it now.

item-spec/item-spec.md
cholmes: I think I didn't write that originally (copied it from another part of the spec). It may technically be correct but I think it's weird, so let's go with a space. Stated in another way - if I actually wrote that it's a typo, as I personally wouldn't mean to write it that way, but perhaps some english major did ;)

item-spec/item-spec.md
cholmes: Good points. I'm not even sure why I did it in items. I agree Catalog seems to be the most appropriate spot for it. Will move it.

cholmes: > Finally, if there's really just GeoTiff/COG left, also move it to the best practice. It seems strange to just have them listed here in its own chapter.

Yeah, I think I was compelled originally by the idea that 'we're doing our own thing', and it's worth calling out. But I agree it feels a bit weird like this. I'll put it in the best practice. Could be nice to 'set' the mime type somewhere for COG, like in the 'cog spec' or on cogeo.org.

cholmes:
Thanks for the review, I took all the suggestions and tried to make them. It ended up being a lot of moving parts, so another read through is appreciated in case I mixed something up.

emmanuelmathot:
It does not worth an official review but I took the time read all the additions to detail the media types in STAC. I find all the explanations about mime-types very descriptive and complete. The table with the list of the most common types is an excellent initiative as reference for STAC. Especially for general users that do not want to end up in endless discussion thread (cf. https://github.com/opengeospatial/geotiff/issues/34), best way to have a de facto standard ;-). Finally, having the profiled types for STAC helps a lot in the tool development. Thank you for the effort.

best-practices.md
```diff
@@ -156,6 +157,52 @@ than the overall best resolution.
...
+
+### Common Media Types in STAC
+
+The following table lists a number of commonly used media types in STAC. The first two (GeoTIFF and COG) are not fully standardized 
+yet, but reflect the community consensus direction. There are many IANA registered types that commonly show up in STAC. The 
+following table lists some of the most common ones you may encounter or use.
+
+| Media Type                                              | Description                                                  |
+| ------------------------------------------------------- | ------------------------------------------------------------ |
+| `image/tiff; application=geotiff`                       | GeoTIFF with standardized georeferencing metadata            |
```
m-mohr, line 176: This may change in the future too, but I guess we can't do anything than to stick to it for now...

item-spec/item-spec.md
```diff
@@ -169,12 +169,23 @@ or streamed. It is allowed to add additional fields.
...
+| type        | string    | [Media type](../catalog-spec/catalog-spec.md#media-types) of the asset. See the [common media types](../best-practices.md#common-media-types-in-stac) in the best practice doc for commonly used asset types. |
 | roles       | \[string] | The [semantic roles](#asset-role-types) of the asset, similar to the use of `rel` in links. |
 
 Beyond to the fields above, any of the [additional fields](#additional-fields) *may* be added to the assets. But this
 is recommended only in special cases, see [Additional Fields for Assets](#additional-fields-for-assets)) for more information.
 
+#### Asset Media Type
+
+Any media type can be used in an Item's asset `type` field, and [registered](https://www.iana.org/assignments/media-types/media-types.xhtml) 
+Media Types are preferred. In cases where custom vendor-specific media types are necessary, they should
```
m-mohr, line 181: The custom media types part seems to duplicate the best practice chapter "Formats with no registered media type". I would think it's enough to link there. People usually just use what is available and dont think about new media types or eben registering them. Seems out of scope for the item spec.

m-mohr (changes requested):
Overall looks very good. I made some additional comment, but none of them really blocking the merge.

What is really blocking this is a recent comment from the OGC: https://github.com/opengeospatial/geotiff/issues/34#issuecomment-770897766

I think with that in mind we can't use the proposed media types and need to come up with our own, which are not adaptions from application/json and application/geo+json. Having that said, I'm afraid that we break tooling when servers will return new media types that are unknown. Like currently many HTTP libraries will automatically parse JSON due to the media type, but once servers return application/stac-item or so, that will probably not happen any more. Even browsers may offer you stac files for download then instead of just serving them nicely formatted as JSON. I think this may lead to more issues than it actually solves and thus my preference at the moment would be to stick with the JSON and GeoJSON media types. Maybe things impove a bit with the `type` field we add to Catalogs and Collections in #889. If we decide to have new media types, we should run intensive tests beforehand with a number of common implementations. This would delay RC1 by a lot, but on the same hand we should probably have it in for the RC phase.

cholmes:
Yeah, I agree that we should stick with json / geojson types, and then drop our 'profile' stuff. It seems far more useful for clients to know that the linked files are json / geojson than and be able to do 'something' with it, vs 'application/vnd.stac-catalog' or whatever that most clients won't recognize. One idea would be to add some extra parameter to our 'link' object, so you'd say 'type=application/json' and then 'type-profile' or 'sub-type' or something equals 'stac-catalog'.

cholmes:
As discussed in #851 and on a call - our proposed rel types don't really work with IANA. So we must pick between a custom type that we register, which would mean applications that understand json or geojson mime types would ignore ours, or not having 'our own' mime type. The consensus was that having our own doesn't add enough value to overcome things like a browser no longer parsing and displaying STAC files (since they wouldn't be json). #944 

This PR did have some nice refactoring, but I'm going to close the issue but keep the branch open and make a new PR when it's in good shape.

Issue #952, opened 2021-01-26
Review and clean-up specs for consistency


cholmes:
Over time the specs have undergone many changes, but reviews are usually done on 'what changed', which can lead to us just missing edits that should have happened. There have also been an array of different editors who all do things at least slightly differently.

We should go over the specs in detail to make sure they are up to date with all the latest changes, and consistent with how we talk about, organize and link to things. 

- [x] Remove 'catalog evolution' in catalog readme, and find other places where we refer to instability.

PR #953, opened 2021-01-26
Relative vs absolute link edits

cholmes:
**Related Issue(s):** #952 , #957

**Proposed Changes:**

1. Clean up the relative vs absolute link section to just link to the best practice.
2. Remove talk about static and dynamic catalogs, as those are now just in the best practice, not the spec itself.

**PR Checklist:**

### DIFF
--- a/best-practices.md
+++ b/best-practices.md
@@ -253,10 +253,10 @@ STAC version. Otherwise some behaviour of functionality may be unpredictable (e.
 
 ## Use of links
 
-The main catalog specification allows both relative and absolute links, and says that `self` links are not required, but are 
+The main STAC specifications allow both relative and absolute links, and says that `self` links are not required, but are 
 strongly recommended. This is what the spec must say to enable the various use cases, but there is more subtlety for when it 
-is essential to use different link types. The best practice is to use one of the below 
-catalog types, applying the link recommendations consistently, instead of just haphazardly applying anything the spec allows.
+is essential to use different link types. The best practice is to use one of the below catalog types, applying the link 
+recommendations consistently, instead of just haphazardly applying relative links in some places and absolute ones in other places.
 
 ### Self-contained Catalogs
 
--- a/best-practices.md
+++ b/best-practices.md
@@ -254,7 +254,7 @@ STAC version. Otherwise some behaviour of functionality may be unpredictable (e.
 
 ## Use of links
 
-The main STAC specifications allow both relative and absolute links, and says that `self` links are not required, but are 
+The STAC specifications allow both relative and absolute links, and says that `self` links are not required, but are 
 strongly recommended. This is what the spec must say to enable the various use cases, but there is more subtlety for when it 
 is essential to use different link types. The best practice is to use one of the below catalog types, applying the link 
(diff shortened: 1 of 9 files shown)


catalog-spec/catalog-spec.md
```diff
@@ -53,9 +53,9 @@ with links.
 
 A more complete list of possible 'rel' types can be seen at the [IANA page of Link Relation Types](https://www.iana.org/assignments/link-relations/link-relations.xhtml).
 
-Please see the chapter 'relative vs absolute links' in the [Item spec](../item-spec/item-spec.md#relative-vs-absolute-links)
-for a discussion on that topic, as well as the [use of links](../best-practices.md#use-of-links) section of the 
-catalog best practices document.
+Currently, the JSON schema for links does not require them to be formatted as URIs, to allow implementors
```
m-mohr, line 56: Hmm, I'm actually not sure whether that is true. I need to check...

m-mohr: JSON Schema got a new format called "uri-reference" with draft 6. A URI Reference is either a URI or a relative-reference, according to RFC3986, section 4.1. So I think we should just add this to the JSON Schemas for links, urls, assets etc. Then we can adopt the language here. I'll put that on my todo.

m-mohr: I've implemented the "iri-reference" format, see commit https://github.com/radiantearth/stac-spec/pull/953/commits/2f223ae402647308bfadb258354ec55c997ec315 - This may require some additional effort in tooling to validate it correctly as iri-reference may not be directly supported. STAC Node Validator for example can't validate it by default, see https://github.com/m-mohr/stac-node-validator/issues/6

I won't approve this PR myself as I've contributed this change and would love if at least two other implementors of validation tools (looking at @lossyrob and @jbants) could check whether this would lead to major issues.

matthewhanson (approved):

best-practices.md
```diff
@@ -253,10 +253,10 @@ STAC version. Otherwise some behaviour of functionality may be unpredictable (e.
 
 ## Use of links
 
-The main catalog specification allows both relative and absolute links, and says that `self` links are not required, but are 
+The main STAC specifications allow both relative and absolute links, and says that `self` links are not required, but are
```
schwehr, line 257: Maybe `The main STAC specifications allow` -> `The STAC specification allows`

What is meant by `main`?  Are there other specifications in play here?

cholmes: I'll work to clarify this. Catalog, Item and Collection are all technically their own 'spec', designed to potentially use stand alone, so that's all I meant. I was changing it to not seem like it was implying just the 'catalog' spec. But I see how 'main' is confusing.

lossyrob (approved):
LGTM!

m-mohr:
As @jbants is also using PySTAC internally, I think we can merge.

best-practices.md
```diff
@@ -254,10 +254,10 @@ STAC version. Otherwise some behaviour of functionality may be unpredictable (e.
...
```
m-mohr, line 257:
```suggestion
The STAC specifications allow both relative and absolute links, and says that `self` links are not required, but are 
```

m-mohr (approved):

Issue #821, opened 2020-05-29
Refresh Examples 


cholmes:
Our examples have been added somewhat haphazardly, and many are just updated with a new construct without really thinking about what the show as a whole. 

We should look over all the examples, and figure out the right mix of ones that tell a coherent story. It'd be nice if some of them were more 'real', and/or if we used some consistent items / collections / catalogs. We should also be sure most all the constructs we talk about our represented in the examples in some way. 

We could also consider a full self-contained collection that is a real stac catalog but can live in github, that shows how all the pieces come together.

m-mohr:
I fully support this. When working on the EO example fixes for #829, some were really confusing.

scottyhq:
Would be nice to verify href URLs are valid. As a particular case https://github.com/radiantearth/stac-spec/blob/master/item-spec/examples/landsat8-sample.json could be updated to point to a valid level-1 collection path such as https://landsat-pds.s3.amazonaws.com/c1/L8/139/045/LC08_L1TP_139045_20170304_20170316_01_T1/index.html from https://docs.opendata.aws/landsat-pds/readme.html. I could create PR for that specific example, just filling it out by hand?

Issue #946, opened 2021-01-22
Move less mature extensions to their own repo


cholmes:
(@m-mohr & @matthewhanson - I'm forgetting the exact conclusion we reached on this, so please sound in if I got it wrong).

Once we release 1.0 our aim is to have a very stable core spec that is not releasing very often. But we also want to have a robust ecosystem of extensions that continue to evolve. Currently most all extensions live in the same repository as the core spec, which would mean that we'd need to do a 'core' release for any young extension that improves.

The current idea to balance those two tensions is to move all less mature extensions (as measured by our extension maturity) to be in their own repository, and it can release more often (like every 2-3 months). The core spec will aim to release only once or twice a year. 

We need to update the maturity of the extensions, with help from #652, so we can designate which ones are stable and will be in the main release. 

- [x] Make template repo in stac-extensions org
- [x] Make a 'central' repo in stac-extensions org that is the main 'list' of non-core extensions
- [ ] Update extensions repo in main readme to link to 'central' extensions repo, and the template repo, explain to people how to make their repo
- [ ] Port over 3-4 extensions to make sure template is working right
- [ ] remove all non-core extensions from dev branch
- [ ] port over all extensions
- [ ] put extension PR's (like cloud storage) into their own repos
- [ ] (after 1.0.0-RC1 release) - make sure CI works, and probably release each extension so they can be used?
- [ ] Review all the new extension repos to make sure they are in good shape after migration (someone other than the person who ported)
- [ ] update code owners file

jisantuc:
I have a radical opinion that every extension should live in its own repo (forced versioning independent of changes Feels Bad :tm: to me), but certainly less mature extensions would be a great start. There's always an option to separate an extension that ends up with too much activity for the main spec after the fact or migrate things back to the main repo as they achieve maturity benchmarks.

m-mohr:
It could be a good idea to make a spac-spec-extensions repo, in this case we wouldn't have the shortcut issue as we would just deploy from there.

cholmes:
@jisantuc - I don't think that's such a radical idea. But I think we're not yet ready to embrace it for the most mature extensions, as having the shortcut is really nice, and our hope is that the most stable extensions won't change too much. But yes, like you say, an extension could move out of the main spec. Or I could also see a new extension, like eo-plus or something that has more 'eo' fields. 

[...]

PR #955, opened 2021-01-27
Examples refresh

cholmes:
**Related Issue(s):** #821

**Proposed Changes:**

1. Moved examples from individual directories into a single /examples folder at the root, and evolved them to be more representative.
 
**PR Checklist:**

### TODO's

*Must have*
- [x] Readme for examples explains all that's going on, compares against best practices
- [x] update all links to refer to the new examples
- [x] validate with all the tools, check with best practices, ensure it works with stac browser

*Should have (but ok to ship without this...)*
- [x] Link to real data (use absolute url's and just put in some assets on public catalogs) for the 'main' assets (mostly done, but they all use the same assets, would be ideal if they were different, and spread out)
- [x] Update the extents of catalogs / collections throughout so they reflect the items
- [x] Update any eo:bands to follow recommendations in #974 (can be just not show the properties one)
- [x] Add proj shape and transform, once we get the answer on #976 
- [x] Add summaries

*After Merge*
 - [ ] Update the self url's to be dev. Figure out how to handle that with releases and master? Probably just released version, and master 'self' will point at the tagged on. Or #984 would be another route, make them all relative.


cholmes:
I'm still playing with the structure, but so far I'm thinking there's one collection with two items in the root examples directory, and then a catalog in the root directory, that links down to at least two different collections, which each link to items. I think I may end up doing one item per (stable) extension. There should likely be more links from the spec itself into the examples. Also thinking that everything here will be 'fictional', except for possible assets - may do absolute links to some known open locations of real satellite data, or maybe try to find sub-100mb images, but I don't really want to have image files in the github repo. I also started https://github.com/stac-utils/stac-examples where we can move 'real' examples - so that should gather production stac items all in one place for people to explore, and then we keep the actual spec to lighter, fictional ones.

After I get the rough structure down I'll need to go back and ensure everything is following as much of our best practices as possible. And make sure we are fully demonstrating all the fields and concepts we articulate in the spec (or as many as is reasonable).

m-mohr:
> Figure out how to make validation work with 'file' schema, as it's not in beta2 release (direct link to the schema?)

For STAC Node validator you can say that it should resolve against a directory, which contains the schemas. See the README.

cholmes:
> For STAC Node validator you can say that it should resolve against a directory (i.e. stac-spec dev branch cloned to your computer), which contains the schemas. See the README.

Ah, sweet - thanks! Will check it out.

cholmes:
Ok, this is pretty much ready for review. I'll update the change log and resolve the conflicts soon.

matthewhanson (approved):

examples/collection-only/collection.json
```diff
@@ -201,20 +201,20 @@
   "links": [
     {
       "rel": "self",
-      "href": "https://storage.cloud.google.com/earthengine-test/catalog/COPERNICUS_S2.json"
+      "href": "https://raw.githubusercontent.com/radiantearth/stac-spec/examples-refresh/examples/collection-example-1/collection.json"
```
m-mohr, line 204: This link will be invalid after the merge.

examples/collection-only/collection.json
```diff
@@ -1,7 +1,7 @@
 {
   "stac_version": "1.0.0-beta.2",
   "stac_extensions": [],
-  "id": "COPERNICUS/S2",
+  "id": "collection-only",
```
m-mohr, line 4: The ids are a bit strange here.

examples/collectionless-item.json
```diff
@@ -74,24 +75,34 @@
   "links": [
     {
       "rel": "self",
-      "href": "http://cool-sat.com/catalog/CS3-20160503_132130_04/CS3-20160503_132130_04.json"
+      "href": "https://raw.githubusercontent.com/radiantearth/stac-spec/examples-refresh/examples/simple-item.json",
+      "type": "application/geo+json; profile=stac-item"
```
m-mohr, line 79: The media types are outdated

examples/extensions-collection/sar-envisat-example/sar-envisat-example.json
```diff
@@ -0,0 +1,101 @@
+{
+    "type": "Feature",
```
m-mohr, line 2: SAR is not one of the stable extensions and thus I think it should be only in the extension, not here.

examples/extensions-collection/sentinel-1-example/sentinel-1-example.json
```diff
@@ -1,18 +1,33 @@
...
-  ],
-  "id": "S1A_EW_GRDM_1SSH_20181103T235855_20181103T235955_024430_02AD5D_5616",
   "type": "Feature",
-  "bbox": [
-    -70.275032,
-    -64.72924,
-    -65.087479,
-    -51.105831
-  ],
+  "stac_version": "1.0.0-beta.2",
```
m-mohr, line 3: It looks like stac_extensions is missing.

examples/item-without-collection.json
```diff
@@ -0,0 +1,163 @@
+{
+  "stac_version": "1.0.0-beta.2",
```
m-mohr, line 2: Wasn't there an item without collection already above? Just wondering whether a second example for it is needed?

extensions/file/README.md
```diff
@@ -9,7 +9,7 @@
 
 Provides a way to specify file related details such as checksum, data type and size for assets in [STAC Items](../../item-spec/item-spec.md) and [STAC Collections](../../collection-spec/collection-spec.md) implementing the [Collection Assets](../collection-assets/README.md) extension.
 
-- [Example](examples/sentinel1.json)
+- [Example](../../examples/extensions-collection/sentinel-1-example/sentinel-1-example.json)
```
m-mohr, line 12: We should leave examples in the extensions if they don't go to core, because then the examples are missing once we fork for the stac-extensions org.

examples/collectionless-item.json
cholmes: Thanks - I tried to catch all of them, but forgot to do an extensive review.

examples/collection-only/collection.json
cholmes: Yes, I have a todo on that for 'after merge'. Though I'm also contemplating making this a more self-contained one.

examples/collection-only/collection.json
cholmes: What would you have them be instead?

examples/extensions-collection/sar-envisat-example/sar-envisat-example.json
cholmes: I do think it'd be valuable to have an extension that show referencing a 'non-core' extension. But I'm fine to remove this and then add something back int.

examples/README.md
```diff
@@ -0,0 +1,95 @@
+# STAC Examples
+
+This directory contains various examples for all parts of the STAC specification. It is structured to be a valid STAC, meaning
+it should successfully load in various tools. It does not follow *all* the [best practices](../best-practices.md) for STAC, mostly
+due to the fact that it's a necessarily contrived example to show the spec, and to do it in github. But we note below where it differs
+from an ideal catalog.
```
lossyrob, line 6:
```suggestion
This directory contains various examples for all parts of the STAC specification. It is structured to be two valid STACs, meaning both [catalog.json](catalog.json) and [collection.json](collection.json) should successfully load in various tools. They do not follow *all* the [best practices](../best-practices.md) for STAC, mostly
due to the fact that they contrive examples to show the spec and we are hosting in GitHub. But we note below where they differ from an ideal catalog.
```

examples/item-without-collection.json
cholmes: Oh, shoot - I thought I got that one out of version control. Removing it.

extensions/file/README.md
cholmes: Good point. Will move them back. This was started before we had our extension decisions, and I thought sar and file might make it...

examples/collection-only/collection.json
m-mohr: Maybe `sentinel-2`?

lossyrob (approved):
Looks good, added a suggestion to reword the intro paragraph of the examples readme.

examples/collection-only/collection.json
cholmes: Ah, cool. I thought you were saying all the example ids were strange. Will update this one.

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+This STAC implementation consists of three files, all contained at the root of the examples directory
+
+**[collection.json](collection.json)** is a minimal Collection, that links to three items. 
+
+**[simple-item.json](simple-item.json)** is the most minimal possible compliant Item record. Most all data will
+include additional fields, as STAC is designed to be a minimal common subset. But it is useful for showing exactly what is
+required.
+
+**[full-item.json](full-item.json)** is a more realistic example, for a hypothetical analytic image
```
m-mohr, line 26: The name sounds a bit weird. Is full < extended? Full sounds "all in", but then below it is extended? If it's just core, maybe core-item?

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+metadata](../item-spec/common-metadata.md). It also links to a variety of assets that is typical for
+satellite imagery, as most providers include a number of complementary files.
+
+**[extended-item.json](extended-item.json)** is arguably an even more realistic example, as it includes a number of the
+[extensions](../extensions/) that are commonly used, to demonstrate how implementations tend to start with the core, and add in
+a number of the core extensions. 
+
+**[collectionless-item.json](collectionless-item.json)** demonstrates the common metadata that is only used when an Item does not have 
+a collection. It is recommended to organize items in collections, but we wanted to show how this works. This is not technically in the
+'simple collection', but it follows the same pattern, so is included here.
```
m-mohr, line 37: What is the "simple collection"?

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+of real world STAC implementations we recommend exploring the [stac-examples](http://github.com/stac-utils/stac-examples) repo, which 
+gathers in one place copies of STAC [Items](../item-spec/item-spec.md) and [Collection](../collection-spec/collection-spec.md) 
+from a number of different production catalogs that all follow good STAC practices. And you should also explore the various catalogs
+listed on [STAC Index](http://stacindex.org), to see full catalogs in production.
+
+## Organization
+
+This directory contains two STAC implementations, both valid, but simplified a bit from following all the best practices. 
+
+### Simple Collection
```
m-mohr, line 16: Why is this called "Simple Collection"?

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+'simple collection', but it follows the same pattern, so is included here.
+
+### Nested Catalog
+
+This STAC implementation shows a common pattern, starting with a catalog that links to a number of distinct collections, which may
+link down to a number of items.
+
+**[catalog.json](catalog.json)** is a minimal catalog implementation, linking to two other collections.
+
+**[collection-only/collection.json](collection-only/collection.json)** is a collection that does not link to any items. This
```
m-mohr, line 46: Maybe I don't understand the structure, but why is this part of the "Nested Catalog" chapter?

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+## In Depth
+
+As mentioned above, the files in this examples directory form valid STAC implementations. They are all based on a 
+fictional remote sensing company called 'Remote Data', with a URL at remotedata.io. This domain has not been set up, so those links
+will not work, but any valid data provider should provide valid links to their homepage. 
+
+The examples use the `rd:` prefix to show how providers can use custom fields when there are not set fields. In the examples these
+do not link to a schema which is completely valid, but it is recommended that providers do write a JSON schema that can validate 
+their custom fields (we will work to add a schema in the future).
```
m-mohr, line 63: > (we will work to add a schema in the future). 

What is meant with that? We add a "template schema" for other extensions to follow? We add a schema for the rd extension?

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+#### Relative vs Absolute Links
+
+One of the [main recommendations](../best-practices.md#use-of-links) of the best practices document is to aim for consistency in 
+the use of relative versus absolute links. STAC implementations should use relative and absolute links as recommended by one of
+the 'catalog types': '[self-contained](../best-practices.md#self-contained-catalogs)', or '[published](../best-practices.md#published-catalogs)' - absolute or relative. 
+
+The catalogs contained here don't follow any of the recommendations in full. They are closest to a 'relative published catalog',
+as they use a self link at the root and then mostly have relative links. This enables them to be mostly portable, while also having
+a stable online location that can be referenced. But they use absolute links to their assets, due to the fact that they are stored in
```
m-mohr, line 83: Talking about this is a bit confusing. This sounds like absolute assets is something not so good although it's totally valid, e.g. if a catalog links to s3 buckets from a HTTP catalog why is that not a best practice? Couldn't we make one of the best practices work in GH? We can get predictable URLs from tags...

examples/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+The catalogs contained here don't follow any of the recommendations in full. They are closest to a 'relative published catalog',
+as they use a self link at the root and then mostly have relative links. This enables them to be mostly portable, while also having
+a stable online location that can be referenced. But they use absolute links to their assets, due to the fact that they are stored in 
+github and we do not want to replicate the large asset files in every branch in github. A true relative published catalog will only
+have absolute links for the self, so that the rest is completely portable. 
+
+#### Catalog Layout
+
+Another important recommendations concerns the [layout of STAC catalogs](../best-practices.md#catalog-layout). This is important
```
m-mohr, line 89: Why not just add a folder? It's just a single folder, right?

examples/collection-only/collection.json
m-mohr: Why not do it in this PR? We can predict the future URL from the tags. 

The URL after release will be:

https://github.com/radiantearth/stac-spec/blob/v1.0.0-beta.2/extensions/README.md
```suggestion
      "href": "https://raw.githubusercontent.com/radiantearth/stac-spec/v1.0.0-rc.1/examples/collection-example-1/collection.json"
```

examples/collectionless-item.json
```diff
@@ -74,24 +75,34 @@
...
-      "rel": "collection",
-      "href": "http://cool-sat.com/catalog/CS3-20160503_132130_04/catalog.json"
+      "rel": "root",
+      "href": "./collection.json",
+      "type": "application/json"
     },
     {
       "rel": "alternate",
       "type": "text/html",
       "href": "http://cool-sat.com/catalog/CS3-20160503_132130_04/CS3-20160503_132130_04.html"
```
m-mohr, line 100: Should we add the root catalog to STAC Index and then we have real HTML Urls...

examples/README.md
cholmes: Oh yeah, we talked this - decided we should have a new category of 'self-contained catalog' that includes absolute url's. Just like we have 'relative' and 'absolute' published catalogs we can have a distinction for self-contained ones, if the assets are relative or absolute.

examples/README.md
cholmes: Yeah, that sounds good, will change.

examples/README.md
cholmes: It's the name of the collection - a simple example of a collection. I could call it 'demo collection' or something...

examples/README.md
cholmes: Because catalog.json links to it. 

There are two valid STAC's in the directory, that can be put in stac index. The collection-only one is in the one that starts with catalog.json, so it's in that section.

examples/README.md
cholmes: Add a schema for the rd extension. Will clarify.

examples/README.md
cholmes: I thought it was nice to come to a directory and see key examples right there. To be able to open them easily and compare, instead of having to dig through directories. I think with best practices we'd add a folder for each item, so it's more than a single folder.

examples/collectionless-item.json
cholmes: Maybe? But I think we should get it set up with the 'publishing' mechanism first, so that it's a final, stable location, and not one we have to update.

cholmes:
Ok, I think this is ready to go. Biggest change is the talking about the catalog type. I went back to modify the catalog type best practices, and realized that we already did allow absolute URL's in self-contained catalogs. But I gave the two types each a 'name' and h4 so we could link to them, and clarified things a bit.

cholmes:
Right before merge I'll change the self url to /dev branch. I do still hope to change it to be a self-contained catalog, with a publish step to put it on stacspec.org somewhere, as in #984 But let's get this in, and I think that's a bit more 'nice to have'.

examples/README.md
m-mohr: Indeed, it's a folder per item. So it depends on the outcome of #984? If we deploy and add the assets, then it would make sense to actually follow the best practices. If there's no assets, it doesn't make much sense to add a folder per item.

Nevertheless, I found the structure at the moment a bit confusing because there's no clear separation at first glance (GH schows you the files first and then you see the README)

examples/README.md
m-mohr: I'd really put each example in a separate directory. I find this structure very confusing and not intuitive. Also, it would allow us to follow the best practices better (i.e. collections are called collection.json, items use the id, etc.)

examples/collectionless-item.json
m-mohr: Is the version number update really a problem? Whether search & replace works on 100 or 200 matches in the files don't really care, right? If it makes it work on GH only, maybe it's enough? Otherwise, you need to implement CI and scripts again for deployment.

m-mohr:
I think I wouldn't change to dev, but to the current version number. It makes links invalid for now, but once we release rc1 they are valid. The issue with dev is that we may forget to update them as this would be a new step to be added to the process.md...

cholmes:
Ok, that makes sense, as then in rc.1 -> rc.2 or final it'll get picked up in the find and replace. Will change it now, as I think this is ready.

cholmes:
Ok, updated to v1.0.0-RC.1 for the anticipated tag. Also realized that this wasn't a true 'relative published catalog' as it had more 'self' links than it should. So removed the extra self ones.

examples/collectionless-item.json
cholmes: I don't think the version update is really a problem. I was leaning more towards just do it in github, but Matt liked the idea of publishing them to an online location, so I made #984 for him. But I see that as 'nice to have', and the default to me is just to keep it as is in github, and update the version. If we decide **not** to do the publishing thing then I'm definitely cool to just call it good in github and set the html rel=alternates to stacindex based on this. I just want to be sure that's the route we're taking.

examples/README.md
cholmes: Yeah, I agree that if we deploy and add assets it would make sense to follow the best practices, at least in the deployed version. I think that tells a nice story - in github it's a self-contained catalog, for local use, and it's maybe a bit more ok that it's not quite best practices. But publishing would use pystac or another tool to format and publish, which would use the best practices, which is where it's more important to follow all the 'rules' - when you're exposing to the world.

cholmes:
Ok, this is now really, really ready to go (I think). I'd really like to get this merged, along with #989, so that extension stuff is all in and then I can start on #946.

m-mohr:
I've lost track on open comments so let's merge and see how it looks in reality. We can do more PRs later.

cholmes:
Sounds great. I think I captured at least some of the open questions in other issues.

Issue #961, opened 2021-02-01
SAR backscatter measurement and convention


m-mohr:
> for the backscatter image assets (or as overall metadata), BackscatterMeasurement (should be gamma0 for CARD4L, but for other radar product, this can be sigma0, beta0 etc), and BackscatterConvention (amplitude or power, this is important!) should be defined. These two (and probably a few others) would make most sense to be included in the sar extension.

_Originally posted by @fangfy in https://github.com/radiantearth/stac-spec/issues/922#issuecomment-770502545_

Currently, values have the card4l prefix in the CARD4L SAR extension.

m-mohr:
| Field Name                  | Type      | Description                                                  |
| --------------------------- | --------- | ------------------------------------------------------------ |
| sar:measurement_type         | string    | The product's type of measurement. One of `beta0`, `gamma0` or `sigma0`. |
| sar:measurment_convention          | string    | The measurement convention. One of `amplitude` (linear amplitude), `power` (linear power) or `angle`.  |

m-mohr:
Additionally, it seems to be useful for have something like sar:pixel_spacing_row and sar:pixel_spacing_column...

emmanuelmathot:
Like in my comment [here](https://github.com/radiantearth/stac-spec/pull/962#issuecomment-773564964), measurements conventions applied for SAR is valid for calibrated backscatter products (ARD)

m-mohr:
Closing, no consensus found.

PR #962, opened 2021-02-02
Additional SAR fields #961

m-mohr:
**Related Issue(s):** #961, #922

**Proposed Changes:**

Adding four fields mostly for products:
* `sar:measurement_type`
* `sar:measurement_convention`
* ~`sar:product_pixel_spacing`~ use `gsd` instead
* ~`sar:product_line_spacing`~ use `gsd` instead

This has been brought up by @fangfy in https://github.com/radiantearth/stac-spec/pull/922#issuecomment-770502545 (and the follow-up comment) and seems like a good idea.

With my limited SAR knowledge, I tried to come up with a list of allowed values and descriptions, but please correct me if there are more values to add or if anything is wrong (which I assume).

**PR Checklist:**

### DIFF
--- a/extensions/sar/README.md
+++ b/extensions/sar/README.md
@@ -31,7 +31,7 @@ To describe frame start and end times, use the [Date and Time Range fields](../.
 | sar:resolution_range        | number    | The range resolution, which is the maximum ability to distinguish two adjacent targets perpendicular to the flight path, in meters (m).  |
 | sar:resolution_azimuth      | number    | The azimuth resolution, which is the maximum ability to distinguish two adjacent targets parallel to the flight path, in meters (m).  |
 | sar:pixel_spacing_range     | number    | The range pixel spacing, which is the distance between adjacent pixels perpendicular to the flight path, in meters (m). Strongly RECOMMENDED to be specified for 'ground-range detected' products. |
-| sar:pixel_spacing_azimuth   | number    | The azimuth pixel spacing, which is the distance between adjacent pixels parallel to the flight path, in meters (m). Strongly RECOMMENDED to be specified for products of type 'ground-range detected' products. |
+| sar:pixel_spacing_azimuth   | number    | The azimuth pixel spacing, which is the distance between adjacent pixels parallel to the flight path, in meters (m). Strongly RECOMMENDED to be specified for 'ground-range detected' products. |
 | sar:looks_range             | number    | Number of range looks, which is the number of groups of signal samples (looks) perpendicular to the flight path. |
 | sar:looks_azimuth           | number    | Number of azimuth looks, which is the number of groups of signal samples (looks) parallel to the flight path. |
 | sar:looks_equivalent_number | number    | The equivalent number of looks (ENL). |
--- a/extensions/sar/README.md
+++ b/extensions/sar/README.md
@@ -39,7 +39,7 @@ To describe frame start and end times, use the [Date and Time Range fields](../.
 | sar:looks_equivalent_number | number    | The equivalent number of looks (ENL). |
 | sar:observation_direction   | string    | Antenna pointing direction relative to the flight trajectory of the satellite, either `left` or `right`. |
 | sar:measurement_type        | string    | The product's type of measurement. One of `beta0`, `gamma0` or `sigma0`. |
-| sar:measurment_convention   | string    | The product's measurement convention. One of `amplitude` (linear amplitude), `power` (linear power) or `angle`. |
+| sar:measurement_convention  | string    | The product's measurement convention. For example, `amplitude` (linear amplitude), `power` (linear power) or `angle`. |
 
 **sar:polarizations** specifies a single polarization or a polarization combination. For single polarized radars one of `HH`, `VV`, `HV` or `VH` must be set. Fully polarimetric radars add all four polarizations to the array. Dual polarized radars and alternating polarization add the corresponding polarizations to the array, for instance for `HH+HV` add both `HH` and `HV`.
 
--- a/extensions/sar/README.md
+++ b/extensions/sar/README.md
@@ -30,10 +30,10 @@ To describe frame start and end times, use the [Date and Time Range fields](../.
 | sar:product_type            | string    | **REQUIRED.** The product type, for example `SSC`, `MGD`, or `SGC` |
 | sar:resolution_range        | number    | The range resolution, which is the maximum ability to distinguish two adjacent targets perpendicular to the flight path, in meters (m).  |
 | sar:resolution_azimuth      | number    | The azimuth resolution, which is the maximum ability to distinguish two adjacent targets parallel to the flight path, in meters (m).  |
-| sar:pixel_spacing_range     | number    | The range pixel spacing, which is the distance between adjacent pixels perpendicular to the flight path, in meters (m). Strongly RECOMMENDED to be specified for products of type `GRD`. |
-| sar:pixel_spacing_azimuth   | number    | The azimuth pixel spacing, which is the distance between adjacent pixels parallel to the flight path, in meters (m). Strongly RECOMMENDED to be specified for products of type `GRD`. |
-| sar:product_pixel_spacing   | number    | The actual product pixel (column) spacing, in meters (m). While `sar:pixel_spacing_range` is given in in radar geometry, this property is in ground geometry. |
-| sar:product_line_spacing    | number    | The actual product line (row) spacing, in meters (m). While `sar:pixel_spacing_azimuth` is given in in radar geometry, this property is in ground geometry. |
+| sar:pixel_spacing_range     | number    | The range pixel spacing, which is the distance between adjacent pixels perpendicular to the flight path, in meters (m). Strongly RECOMMENDED to be specified for 'ground-range detected' products. |
+| sar:pixel_spacing_azimuth   | number    | The azimuth pixel spacing, which is the distance between adjacent pixels parallel to the flight path, in meters (m). Strongly RECOMMENDED to be specified for products of type 'ground-range detected' products. |
+| sar:product_pixel_spacing   | number    | The actual product pixel (column) spacing, in meters (m). While `sar:pixel_spacing_range` is given in in radar geometry, this property is in ground geometry. The two properties are usually mutually exlusive. |
+| sar:product_line_spacing    | number    | The actual product line (row) spacing, in meters (m). While `sar:pixel_spacing_azimuth` is given in in radar geometry, this property is in ground geometry.  The two properties are usually mutually exlusive. |
 | sar:looks_range             | number    | Number of range looks, which is the number of groups of signal samples (looks) perpendicular to the flight path. |
 | sar:looks_azimuth           | number    | Number of azimuth looks, which is the number of groups of signal samples (looks) parallel to the flight path. |
 | sar:looks_equivalent_number | number    | The equivalent number of looks (ENL). |
--- a/extensions/sar/README.md
+++ b/extensions/sar/README.md
@@ -36,6 +36,8 @@ To describe frame start and end times, use the [Date and Time Range fields](../.
 | sar:looks_azimuth           | number    | Number of azimuth looks, which is the number of groups of signal samples (looks) parallel to the flight path. |
 | sar:looks_equivalent_number | number    | The equivalent number of looks (ENL). |
 | sar:observation_direction   | string    | Antenna pointing direction relative to the flight trajectory of the satellite, either `left` or `right`. |
+| sar:measurement_type        | string    | The product's type of measurement. One of `beta0`, `gamma0` or `sigma0`. |
+| sar:measurment_convention   | string    | The product's measurement convention. One of `amplitude` (linear amplitude), `power` (linear power) or `angle`. |
 
 **sar:polarizations** specifies a single polarization or a polarization combination. For single polarized radars one of `HH`, `VV`, `HV` or `VH` must be set. Fully polarimetric radars add all four polarizations to the array. Dual polarized radars and alternating polarization add the corresponding polarizations to the array, for instance for `HH+HV` add both `HH` and `HV`.
 
--- a/extensions/sar/README.md
+++ b/extensions/sar/README.md
@@ -32,6 +32,8 @@ To describe frame start and end times, use the [Date and Time Range fields](../.
 | sar:resolution_azimuth      | number    | The azimuth resolution, which is the maximum ability to distinguish two adjacent targets parallel to the flight path, in meters (m).  |
 | sar:pixel_spacing_range     | number    | The range pixel spacing, which is the distance between adjacent pixels perpendicular to the flight path, in meters (m). Strongly RECOMMENDED to be specified for products of type `GRD`. |
 | sar:pixel_spacing_azimuth   | number    | The azimuth pixel spacing, which is the distance between adjacent pixels parallel to the flight path, in meters (m). Strongly RECOMMENDED to be specified for products of type `GRD`. |
+| sar:product_pixel_spacing   | number    | The actual product pixel (column) spacing, in meters (m). While `sar:pixel_spacing_range` is given in in radar geometry, this property is in ground geometry. |
+| sar:product_line_spacing    | number    | The actual product line (row) spacing, in meters (m). While `sar:pixel_spacing_azimuth` is given in in radar geometry, this property is in ground geometry. |
 | sar:looks_range             | number    | Number of range looks, which is the number of groups of signal samples (looks) perpendicular to the flight path. |
 | sar:looks_azimuth           | number    | Number of azimuth looks, which is the number of groups of signal samples (looks) parallel to the flight path. |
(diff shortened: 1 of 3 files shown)


fangfy:
Don't think "angle" is a measurement convention? And not sure if any product is provided in log scale (dB). Maybe limit to amplitude (linear amplitude) and power (linear power) for now? The card4l:conversion_eq defines how to turn the data into db.

matthewhanson:
I'm not sure I agree that `sar:pixel_spacing_range` and `sar:pixel_spacing_azimuth` is in radar geometry....or maybe I just don't understand the difference and should catch up on the discussion in #922.
At the very least I would argue that using the term "pixel" with units of meters to refer to something other than actual distance on the ground is a poor selection of terms.

Perhaps a clear explanation of the difference between the 4 fields is warranted.

m-mohr:
I'm not sure myself, but I'm just doing what a SAR expert told me in #922 🤷‍♂️ So maybe better discuss with her there?

fangfy:
> Angle I got from the POL Card4l. It's listed there... Maybe I misunderstood...? 🤔 card4l:conversion_eq is not part of the SAR extension...

um... I didn't find reference to Angle (or mention of backscatter convention) in the POL excel file. 

Since this is is a SAR specific CARD4L, many parameters are SAR specific. I'm not sure what should be the design principle here (as how much should be incorporated into the SAR extension).

fangfy:
> I'm not sure I agree that `sar:pixel_spacing_range` and `sar:pixel_spacing_azimuth` is in radar geometry....or maybe I just don't understand the difference and should catch up on the discussion in #922.
> At the very least I would argue that using the term "pixel" with units of meters to refer to something other than actual distance on the ground is a poor selection of terms.
> 
> Perhaps a clear explanation of the difference between the 4 fields is warranted.

`_range` and `_azimuth` are defined relative to sensor, not a ground coordinate system. The 'meter' sampling unit is still valid, just in a different coordinate. A product should either have sar:pixel_spacing_range and sar:pixel_spacing_azimuth defined (usually a lower level product, such as GRD) or sar:product_pixel_spacing and sar:product_line_spacing defined (a higher level one such as a CARD4L). Not both.

m-mohr:
Thank you, @fangfy.

> um... I didn't find reference to Angle (or mention of backscatter convention) in the POL excel file.

Found it here in the spec: image

> Since this is is a SAR specific CARD4L, many parameters are SAR specific. I'm not sure what should be the design principle here (as how much should be incorporated into the SAR extension).

Me neither. I try to limit the "core" STAC extensions like SAR to fields that people would search on (STAC has a focus on search and discovery) and in addition, are not only free-texts, so fields, where you have a choice like for beta0, gamma0, etc. or can come up with a numerical representation with a single value, seem good candidates. It's difficult with SAR as it's so diverse...

> Or maybe it should always only have pixel_spacing and line_spacing and a different field for what's the coordinate system used for the pixel grid....

Honestly, I can't follow with regards to the SAR azimuth/range/pixel/.. discussion as I'm more a software developer than a remote sensing expert, so this is really difficult for me to understand. But it's not your job to teach me that ;-)

I'm happy to change to whatever makes sense to the experts.

scottyhq:
Hi ! Thanks for all the effort on this @m-mohr and @fangfy , I'll have to brush up on CARD4L SAR, but @matthewhanson pinged me to add some comments here. Regarding coordinates and units, here is a nice doc from ASF https://asf.alaska.edu/wp-content/uploads/2019/03/rtc_atbd_v1.2_final.pdf related to amplitude products. In particular, we have 3 common coordinate systems with SAR, all related by incidence angle and therefore all with units of meters: 

Screen Shot 2021-02-03 at 9 28 34 AM

SSC/SLC data is in Beta0 because it does not depend on a terrain model, it is simply what the sensor measures.  these natural radar coordinates are traditionally referred to as 'slant range' and 'azimuth'. GRD data is more amenable to STAC because it is referenced to the earth (ellipsoid, not local terrain, so Sigma0) and pixel coordinates are still commonly referred to as 'range' and 'azimuth'

Back to the spec, it seems the confusion is that `sar:pixel_spacing_range` alone is ambiguous, and the meaning depends on `sar:product_type`. 

A source of confusion I have is in which extension fields should belong. For example most users of STAC are probably interested in on ground coordinates, but for that there is the [common metadata has 'Ground Sample Distance' gsd filed](https://github.com/radiantearth/stac-spec/blob/4988a40f6071666d345e4a1c5adc9a6338973769/item-spec/common-metadata.md#instrument) 'These fields will often be combined with domain-specific extensions that describe the actual data, such as the eo or sar extensions.'

I'll leave a few more comments directly in the code, hope this helps clarify some things!

extensions/sar/README.md
```diff
@@ -30,12 +30,16 @@ To describe frame start and end times, use the [Date and Time Range fields](../.
...
+| sar:pixel_spacing_range     | number    | The range pixel spacing, which is the distance between adjacent pixels perpendicular to the flight path, in meters (m). Strongly RECOMMENDED to be specified for 'ground-range detected' products. |
+| sar:pixel_spacing_azimuth   | number    | The azimuth pixel spacing, which is the distance between adjacent pixels parallel to the flight path, in meters (m). Strongly RECOMMENDED to be specified for products of type 'ground-range detected' products. |
+| sar:product_pixel_spacing   | number    | The actual product pixel (column) spacing, in meters (m). While `sar:pixel_spacing_range` is given in in radar geometry, this property is in ground geometry. The two properties are usually mutually exlusive. |
+| sar:product_line_spacing    | number    | The actual product line (row) spacing, in meters (m). While `sar:pixel_spacing_azimuth` is given in in radar geometry, this property is in ground geometry.  The two properties are usually mutually exlusive. |
```
scottyhq, line 36: i think just the sar:pixel_spacing_range and sar:pixel_spacing_azimuth is sufficient here, since they are mutually exclusive

extensions/sar/README.md
```diff
@@ -30,12 +30,16 @@ To describe frame start and end times, use the [Date and Time Range fields](../.
...
 | sar:looks_range             | number    | Number of range looks, which is the number of groups of signal samples (looks) perpendicular to the flight path. |
 | sar:looks_azimuth           | number    | Number of azimuth looks, which is the number of groups of signal samples (looks) parallel to the flight path. |
 | sar:looks_equivalent_number | number    | The equivalent number of looks (ENL). |
 | sar:observation_direction   | string    | Antenna pointing direction relative to the flight trajectory of the satellite, either `left` or `right`. |
+| sar:measurement_type        | string    | The product's type of measurement. One of `beta0`, `gamma0` or `sigma0`. |
+| sar:measurment_convention   | string    | The product's measurement convention. One of `amplitude` (linear amplitude), `power` (linear power) or `angle`. |
```
scottyhq, line 42: These are specific to SAR *amplitude* data, so should probably be in the CARD4L-NRB? There are many other SAR products out there where these terms are irrelevant (such as phase, coherence, interferometric phase)

scottyhq: or rename as 'amplitude_type' and 'amplitude_convention'? note also the current typo in measurement_convention

m-mohr: Yeah, I'm thinking about splitting SAR a bit into Commons, Backscatter, GRD, POL, ...

extensions/sar/README.md
m-mohr: Maybe just the naming is off and we can make it more universal?

fangfy: `gsd`, that's what I was thinking. The pixel spacing for card4l product (which has to be in ground to be `analysis ready`) can be inferred from transform or explicitly defined in gsd. So they don't need to be defined elsewhere I think. pixel_spacing_range and pixel_spacing_azimuth just don't apply to these products.

fangfy: There might be special/rare case data are provided with different spacing in x and y. Not sure how STAC handles.

m-mohr: It doesn't handle it at all. There's just one gsd for x and y so far. So do I understand correctly that we can remove sar:product_pixel_spacing / sar:product_line_spacing and just replace it with a single gsd entry? That would be nice, indeed.

extensions/sar/README.md
fangfy: `amplitude` is a bit of a confusing term for me, when it can be provided as `amplitude` or `power/intensity`. I understand this is to differentiate from phase and other measurement. I do prefer the terminology we use for product type in card4l, so maybe `backscatter_measurement_type` and `backscatter_measurement_convention` (or without `_measurement`)? or split into different key parameters for different product types. I think these are critical SAR relevant parameters but don't have strong view which extension they should be in.

m-mohr: @fangfy The CARD4L SAR extension also includes POL, which has a list of these values, see the comment above: https://github.com/radiantearth/stac-spec/pull/962#issuecomment-772498604 - I'm just trying to align them somehow. That's why I did not use backscatter in the name.

extensions/sar/README.md
jbants: This is how we are defining GSD for the sensor.
https://www.ssec.wisc.edu/sose/tutor/ifov/define.html.

But, there are many definitions.

fangfy: resolution and sampling can be and often are for SAR different. My understanding of the face meaning of GSD is the same as product spacing used in card4l.

extensions/sar/README.md
fangfy: Thanks @m-mohr! The POL document defines two possible type of products and uses `Measurement Type` to distinguish between CovMat and PRD products. However, for the `measurement_type` discussed here, it should be gamma0.  So you are right, this field is common. But something else is needed to identify CovMat or PRD (maybe within product type). Measurement convention (unit) has angle because for PRD, there is an angle measurement. So in this case, the field is extended to identify what (image/asset specific) measurement it is. And this means, it can be extended further to include other complex/phase type of measurement.... I'm not sure I can provide a complete list though as I'm less familiar with other SAR products and terminologies people use. Maybe worth sharing current state with more people in CEOS CARD4L working group and see how best to proceed from here....

m-mohr: I've changed product_type in the CARD4L extension in a way that it makes clear whether it's NRB, COVMAT or PRD. In the general SAR extension, it's (for now) up to the user to make the product_type explicit enough.

Yes, the measurement type would likely need to be extended. I guess it's fine that we just start a list and add to it later. For now, I made the set of values to a list of examples and we can add definitive (enum-like) values later.

I've also fixed the typo in measurement_convention.

extensions/sar/README.md
m-mohr: So if I understood correctly, I'll remove the new product_*_spacing fields here and add gsd to the CARD4L SAR spec.

fangfy: @m-mohr missed your question above, but yes I think we gsd is sufficient for now.

extensions/sar/README.md
fangfy: I’ll bring this up in the WG but it might take some time to get feedback.

m-mohr: Thank you, @fangfy.

extensions/sar/README.md
```diff
@@ -30,12 +30,14 @@ To describe frame start and end times, use the [Date and Time Range fields](../.
...
 | sar:resolution_azimuth      | number    | The azimuth resolution, which is the maximum ability to distinguish two adjacent targets parallel to the flight path, in meters (m).  |
-| sar:pixel_spacing_range     | number    | The range pixel spacing, which is the distance between adjacent pixels perpendicular to the flight path, in meters (m). Strongly RECOMMENDED to be specified for products of type `GRD`. |
-| sar:pixel_spacing_azimuth   | number    | The azimuth pixel spacing, which is the distance between adjacent pixels parallel to the flight path, in meters (m). Strongly RECOMMENDED to be specified for products of type `GRD`. |
+| sar:pixel_spacing_range     | number    | The range pixel spacing, which is the distance between adjacent pixels perpendicular to the flight path, in meters (m). Strongly RECOMMENDED to be specified for 'ground-range detected' products. |
+| sar:pixel_spacing_azimuth   | number    | The azimuth pixel spacing, which is the distance between adjacent pixels parallel to the flight path, in meters (m). Strongly RECOMMENDED to be specified for products of type 'ground-range detected' products. |
```
m-mohr, line 34:
```suggestion
| sar:pixel_spacing_azimuth   | number    | The azimuth pixel spacing, which is the distance between adjacent pixels parallel to the flight path, in meters (m). Strongly RECOMMENDED to be specified for 'ground-range detected' products. |
```

emmanuelmathot:
Sorry to come up a bit late but I think there is a missing information here in the definition of the new `sar:measurement` fields. data can are considered as `beta0`, `gamma0` and then `sigma0` once calibrated or terrain corrected. Most of the SAR datasets are delivered as uncalibrated and are either amplitude, intensity or magnitude as measured by the instrument and defined by the `sar:product_type`. It is necessary to link this measurements fields are with a concept like `processing:level` which is not very commonly defined in SAR products.

emmanuelmathot:
I see the point of having the information of the calibration of the backscatter for ARD product and thus in CARD4L. I am less sure that this information belongs to the `sar` extension.

m-mohr:
The CARD4L extension is meant to be a mapping of CARD4L to STAC. In the best case it would define (mostly) no fields at all. Thus, either they should belong to SAR or to a new extension (e.g. sar-backscatter or so). It seemed that these two fields are at least somewhat common to higher-level products as they are not just in the NRB, but also in the POL CARD4L and also seem to have some specific values. But it seems that there's not even consensus across SAR folks so I may just move them back into card4l until we can get a range of people with a broad SAR background so that eventually we can for example define extensions for specific higher-level SAR products. At least from the comments above I'm completely confused and (without much SAR background myself) don't know how to call things, which values to allow etc.

scottyhq:
> we can get a range of people with a broad SAR background so that eventually we can for example define extensions for specific higher-level SAR products that cover more of the proprietary CARD4L fields. While I appreciate all the contributions here, I'm completely confused now and (without much SAR background myself) don't know how to call things, which values to allow etc.

Apologies for the confusion @m-mohr, again, really appreciate your efforts to accommodate. In the spirit of an implementation-driven nature of STAC development, it does seem like a way forward is to not touch the existing SAR extension, put anything new into the CARD4L extension, and then based on some concrete examples follow up with suggestions on moving/creating new fields in existing extensions? 

Are there any prototype CARD4L STAC catalogs publicly available? Obviously part of the discussion here is to define what needs to be in such a catalog, but it looks like examples for the extension are still marked as 'todo' (#922) and based on this recent webinar seems like a work in progress? https://docs.google.com/forms/d/e/1FAIpQLScCMvd6g_2FTsjw0csRymnTHl_pSSdZ_ZnPZKLnINGwm638yw/viewform 

I'm wondering if @asjohnston-asf or @talogan could point to some examples from ASF (or others involved int the CEOS efforts)? 

For what it's worth there is also this SAR Gamma0 dataset, https://registry.opendata.aws/sentinel-1-rtc-indigo, seems like another great opportunity for a CARD4L-compliant STAC example! In case it helps I started an effort to generate STAC1.0 compliant metadata for this dataset here https://github.com/stac-utils/stac-sentinel/pull/1 a while back, but have been drawn into other things for the time being. happy for anyone to take over that effort :)

m-mohr:
Thanks for sounding in, @scottyhq - I appreciate it.

> In the spirit of an implementation-driven nature of STAC development, it does seem like a way forward is to not touch the existing SAR extension, put anything new into the CARD4L extension, and then based on some concrete examples follow up with suggestions on moving/creating new fields in existing extensions?

It's a bit more complicated in this case, as we aim to submit the CARD4L extension and examples to CEOS for approval. As that's a lengthy process, I won't do that for every change we want to do in STAC (e.g. moving fields from card4l to sar). So once we have them in card4l, they'll stick there for a bit. Thus my aim was to get at least some fields that seemed to be good candidates for the SAR extensions in there. As far as I've understood you all (@fangfy @scottyhq @emmanuelmathot), there's no consensus yet and I had hoped that you'd discuss here and come to a conclusion how we could do things in the SAR extension.

> Are there any prototype CARD4L STAC catalogs publicly available?

No, not with pure STAC. All waiting for this to be solved and the CARD4L STAC spec to be finished.

> Obviously part of the discussion here is to define what needs to be in such a catalog

That's defined in the CEOS specification and I try to do the mapping to STAC now.

> but it looks like examples for the extension are still marked as 'todo' (#922)

The STAC examples are still in todo, which we'll generate once the CARD4L STAC mapping has been agreed on. There are examples in the CEOS CARD4L metadata specification, thus I doubt a STAC example would help us a lot here. What would you expect to see in such an example that's not yet clear from the CEOS documents?

> and based on this recent webinar seems like a work in progress? https://docs.google.com/forms/d/e/1FAIpQLScCMvd6g_2FTsjw0csRymnTHl_pSSdZ_ZnPZKLnINGwm638yw/viewform

CEOS specifications for NRB and POL should hopefully be stable. At least they have a 5.0 version number, which sounds stable.

m-mohr:
Closing, no consensus found.

PR #933, opened 2020-12-15
CARD4L Optical Extension (SR + ST)

m-mohr:
### The STAC CARD4L Optical Extension has been migrated to a separate repository: <https://github.com/stac-extensions/card4l/tree/main/optical>

Issue #803, opened 2020-05-12
data coverage percentage


matthewhanson:
Sentinel has a percent data coverage, representing the total amount of the scene that has data, this can be a very useful property for any data that has been cut into tiles as it can be used to filter out scenes that are just small pieces of data.  (for instance if you only want scenes that cover the entire tile).

I'd propose a new field, `data_coverage` or similar.  It is applicable to EO data, but also SAR as well as potentially any raster data.  I think it might be best suited under the Item common properties.

m-mohr:
Sounds useful, may even work for data cubes in a slightly different way. Naming should be aligned with cloud_cover, if possible in a meaningful way.

m-mohr:
Is that something that would somehow also suit into the file extension?

PowerChell:
Discussion can continue here: https://github.com/stac-extensions/raster/issues/39

PR #963, opened 2021-02-02
Additional File Fields

m-mohr:
**Related Issue(s):** #922, #933 

**Proposed Changes:**

After a discussion with Matt, we propose to add some new fields to the file extension as they seem pretty related to the file (content).

**PR Checklist:**

### DIFF
--- a/extensions/file/README.md
+++ b/extensions/file/README.md
@@ -7,14 +7,14 @@
 - **Extension [Maturity Classification](../README.md#extension-maturity): Proposal**
 - **Owner**: @m-mohr
 
-Provides a way to specify file related details such as checksum, data type and size for assets and links in [STAC Items](../../item-spec/item-spec.md), [STAC Catalogs](../../catalog-spec/catalog-spec.md) and [STAC Collections](../../collection-spec/collection-spec.md).
+Provides a way to specify file related details such as checksum, data type and size for assets in [STAC Items](../../item-spec/item-spec.md) and [STAC Collections](../../collection-spec/collection-spec.md) implementing the [Collection Assets](../collection-assets/README.md) extension.
 
 - [Example](examples/sentinel1.json)
 - [JSON Schema](json-schema/schema.json)
 
-## *Link Object* and *Asset Object* fields
+## *Asset Object* fields
 
-The following fields can be used for Links (in the [`Link Object`](../../item-spec/item-spec.md#link-object)) and assets (in the [`Asset Object`](../../item-spec/item-spec.md#asset-object)) although most of these fields make more sense for assets and are usually not very useful for links.
+The following fields can be used for assets (in the [`Asset Object`](../../item-spec/item-spec.md#asset-object)).
 
 | Field Name           | Type                                    | Description                                                  |
 | -------------------- | --------------------------------------- | ------------------------------------------------------------ |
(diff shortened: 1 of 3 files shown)


extensions/file/README.md
```diff
@@ -14,19 +14,42 @@ Provides a way to specify file related details such as checksum, data type and s
...
+| Field Name | Data Type | Description                                                  |
+| ---------- | --------- | ------------------------------------------------------------ |
+| values     | \[any]    | **REQUIRED.** The value(s) in the file. At least one array element is required. |
+| summary    | string    | **REQUIRED.** A short description of the value(s).           |
+
+ For example for a cloud cover mask, `card4l:values` property could contain the following data:
+
+```json
+[
+	{"value": [0], "summary": "clear"},
```
matthewhanson, line 47: Why not a dictionary for this, since that would enforce unique values in the mapping

m-mohr: You could have non-string values in some file formats. Also, it doesn't need you to duplicate the texts in summary.

extensions/file/README.md
```diff
@@ -14,19 +14,42 @@ Provides a way to specify file related details such as checksum, data type and s
...
+| -------------------- | --------------------------------------- | ------------------------------------------------------------ |
+| file:bits_per_sample | integer                                 | Bits per sample, e.g. 8, 16, 32, ...                         |
+| file:byte_order      | string                                  | The byte order of integer values in the file. One of `big-endian` or `little-endian`. |
+| file:checksum        | string                                  | Provides a way to specify file [checksums](#checksums) (e.g. BLAKE2, MD5, SHA1, SHA2, SHA3). The hashes are self-identifying hashes as described in the [Multihash specification](https://github.com/multiformats/multihash) and must be encoded as hexadecimal (base 16) string with lowercase letters. |
+| file:data_type       | string                                  | The data type of the file. One of the [data types](#data-types) below. |
+| file:header_size     | integer                                 | The header [size](#sizes) of the file, specified in bytes.   |
+| file:nodata          | \[any]                                  | Value(s) for no-data.                                        |
```
kylebarron, line 26:
```suggestion
| file:nodata          | \[any\]                                  | Value(s) for no-data.                                        |
```
Do you need a closing `\]`?

m-mohr: No, but it doesn't make any harm either, right?

extensions/file/README.md
matthewhanson: Well you can still have a dictionary key that is a number too.
I don't have a strong opinion about it really, I tend to lean toward dictionaries when there is an obvious unique key.

extensions/file/README.md
kylebarron: I was just confused because you included a closing `\]` a few lines below.

extensions/file/README.md
m-mohr: But then there's also null, true, false, ...

matthewhanson:
New fields look good to me, I'm going to go ahead and approve, although find with changing values to a dictionary, or not.

matthewhanson (approved):

extensions/file/README.md
m-mohr: Yeah, when writing by hand you sometimes do weird things ;-)

extensions/file/README.md
kylebarron: > Well you can still have a dictionary key that is a number too

JSON only allows string keys I believe https://stackoverflow.com/a/8758771

m-mohr:
@matthewhanson Is #803 (data coverage) something that we could also add here?

cholmes (approved):
Looks good. Discussed on the call, one change requested - to limit this just to 'assets', not including 'links'. We realized that was historical, as this started as the checksum extension, and we decided even for that it makes less sense on a link. If users are wanting to put this on a link then they should probably be defining that link as an asset in the first place.

PR #969, opened 2021-02-03
Item Properties and Item Asset clarification

philvarner:
**Related Issue(s):** #

**Proposed Changes:**

1. Clarify which objects (e.g., Item Properties or Item Asset) fields apply to
2. Change uses of "Item Assets" to "Item Asset objects" so not to confuse with the item-assets extension (in general, I think we should adopt this style -- instead of pluralizing a STAC type (Type->Types), refer to `Type objects`)

**PR Checklist:**

- [ ] n/a This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec), and I have opened issue/PR #XXX to track the change.

### DIFF
--- a/collection-spec/collection-spec.md
+++ b/collection-spec/collection-spec.md
@@ -109,7 +109,7 @@ The object provides information about a provider. A provider is any of the organ
 
 ### Link Object
 
-This object describe a relationship with another entity. Data providers are advised to be liberal with links.
+This object describes a relationship with another entity. Data providers are advised to be liberal with links.
 
 | Field Name | Type   | Description                                                  |
 | ---------- | ------ | ------------------------------------------------------------ |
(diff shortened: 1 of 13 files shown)


m-mohr:
Is that a duplicate of #900?

philvarner:
> Is that a duplicate of #900?

I branched it from #900 so it has all those changes too, but it has additional clarifications that may or may not be correct. I'll rebase when #900 is merged for a clean PR.

cholmes:
Ok, #900 us now merged, so you should be good to rebase this one.

philvarner:
h/t to @vpipkt for bringing up the need to clarify these fields

jisantuc (approved):
Good from me. This kind of change makes me curious about whether there's a latent schema that applies to extensions. Discovering that schema would be a boon to downstream extension implementers / parsers (at least it would be a boon to stac4s)

m-mohr:
> whether there's a latent schema that applies to extensions.

What is a latent schema? What exactly are you looking for?

m-mohr (approved):

m-mohr:
Approving, but CI fails, which needs to be fixed before the merge.

PR #970, opened 2021-02-04
Clean-up Media Type Info

cholmes:
**Related Issue(s):** #930 

**Proposed Changes:**

1. This organizes how we talk about media types a bit better. It makes the media types of stac structures (item/catalog/collection) more clear, moves the common media types to 'best practices', and talks about asset media types a bit differently than link media types.

**PR Checklist:**


cholmes:
Should probably squash commits on merge, since this has the history of #944 in it.

item-spec/item-spec.md
```diff
@@ -251,6 +228,13 @@ in the file).
 For examples of fields that this construct is recommended for, see the [section of STAC Best Practices](../best-practices.md#common-use-cases-of-additional-fields-for-assets)
 that talks about common use cases of additional fields for assets.
 
+## Media Type for STAC Item
+
+A STAC Item is a GeoJSON file ([RFC 7946](https://tools.ietf.org/html/rfc7946)), and thus should use the 
+`[application/geo+json](https://tools.ietf.org/html/rfc7946#section-12)` as the [Media Type](https://en.wikipedia.org/wiki/Media_type) 
+(previously known as the MIME Type). STAC Items are recommended to add the parameter `profile` with the value `stac-item` to
```
m-mohr, line 235: This is the same issue as with GeoTiff. You can't add parameters, I think it should just be the GeoJSON media type. Also, I think we should say that `application/json` can also occur in implementations. I don't think everybody implemented the GeoJSON media type.

item-spec/item-spec.md
```diff
@@ -167,12 +167,23 @@ or streamed. It is allowed to add additional fields.
...
+| type        | string    | [Media type](#asset-media-type) of the asset. See the [common media types](../best-practices.md#common-media-types-in-stac) in the best practice doc for commonly used asset types. |
 | roles       | \[string] | The [semantic roles](#asset-role-types) of the asset, similar to the use of `rel` in links. |
 
 Beyond to the fields above, any of the [additional fields](#additional-fields) *may* be added to the assets. But this
 is recommended only in special cases, see [Additional Fields for Assets](#additional-fields-for-assets)) for more information.
 
+#### Asset Media Type
+
+Any media type can be used in an Item's asset `type` field, and [registered](https://www.iana.org/assignments/media-types/media-types.xhtml) 
+Media Types are preferred. In cases where custom vendor-specific media types are necessary, they should
```
m-mohr, line 179: I think the .vnd info is a bit out of scope here. Most people will not make up their own media types and register them.

catalog-spec/catalog-spec.md
```diff
@@ -77,6 +77,30 @@ There are additional `rel` types in the [Using Relation Types](../best-practices
...
+
+The following table lists the Media Types to use for STAC structures.
+
+| Media Type                     | Description                                                  |
+| ------------------------------ | ------------------------------------------------------------ |
+| `application/geo+json`	     | A STAC [Item](../item-spec/README.md)                        |
+| `application/json`             | A STAC [Catalog](#stac-catalog-specification)                |
+| `application/json`             | A STAC [Collection](../collection-spec/README.md)            |
+
+### Media Type for STAC Catalogs
```
m-mohr, line 98: Maybe switch order of the chapters? After the table the media type definition for STAC Catalogs seems superfluous, but if you first define it and then give an overview table, it seems to have a better order.

item-spec/item-spec.md
cholmes: Ah, meant to remove the 'profile' bit. Good idea on saying application/json is ok too.

item-spec/item-spec.md
cholmes: Good catch, will move this out / link to there.

catalog-spec/catalog-spec.md
cholmes: Sounds good.

cholmes:
I think I addressed all the comments, so this should be good for final review.

m-mohr (approved):

matthewhanson (approved):

PR #971, opened 2021-02-05
Add type field to Collection & Catalog

cholmes:
**Related Issue(s):** #889 

**Proposed Changes:**

1. Added `type` field to catalog and collection spec
2. Added best practice section on how to distinguish stac files

**PR Checklist:**

- [x] This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec), and I have opened issue/PR #XXX to track the change.


cholmes:
Ok, this is up for discussion, feedback welcome. I think it's in line with what we discussed.

* re-used the `type` field from GeoJSON, so it aligns with Item and ItemCollection.
* Set value to `Catalog` and `Collection`. The capital is a bit off in our spec, but aligns with GeoJSON, where we got the type field from.
* Wasn't sure if we should prefix with STAC in some way, as `FeatureCollection` vs `Collection` might be confusing. But I don't love the options: STACCollection or StacCollection. Or we could not try to align with their capital camel case, and then it'd be stac_catalog and stac_collection.

Thoughts?

m-mohr (changes requested):
This still needs JSON Schema changes.

Also, the "How to Differentiate STAC Files" part could probably be a bit simplified.

I'll work on it.

cholmes:
Oh, yeah, apologies - I meant to mention that I'd really like your help on the schema changes. Thanks @m-mohr! And thanks for help on differentiating STAC files. Also our linter forced me to pick a 'language' for the code blocks, but I think you just wrote it in pseudo-code? If you've got better formatting ideas then go for it. Could just do it in python or javascript, and people will get the idea.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - 'summaries' are now available in the Catalog spec, so both catalogs and collections can make use of it. ([#903](https://github.com/radiantearth/stac-spec/issues/903))
 - Recommendation to enable CORS
 - A 'visual' option as an asset role.
+- Catalog and Collection now require a `type` parameter, to be set to `Catalog` or `Collection` for clients to more easily distinguish them easily.
```
jisantuc, line 19:
```suggestion
- Catalog and Collection now require a `type` parameter, to be set to `Catalog` or `Collection` for clients to more easily distinguish them.
```

best-practices.md
```diff
@@ -429,3 +430,34 @@ basic geographic filtering from listeners.
 The dynamic STAC API would then listen to the notifications and update its internal datastore whenever new data comes into
 the static catalog. Implementors have had success using AWS Lambda to do a full 'serverless' updating of the elasticsearch
 database, but it could just as easily be a server-based process.
+
+## How to Differentiate STAC Files
+
+Any tool that crawls a STAC implementation or encounters a STAC file in the wild needs a clear way to determine if it is an Item, 
+Collection, Catalog or [ItemCollection](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-beta.1/fragments/itemcollection) 
+(part of the [STAC API spec](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-beta.1). As of 1.0.0 this is done primarily with the `type` field, and secondarily in Items with `stac_version`, or optionally the `rel` of the link to it.
```
jisantuc, line 438: I think this challenge is less pronounced in API settings than static catalog settings, since the OpenAPI spec for the API tells you what to expect at each endpoint (and also catalogs don't live anywhere in the API). That said if someone saves off an `ItemCollection` and then refers to it from a static catalog, I guess all bets are off, so I think I've talked myself into this inclusion.

catalog-spec/catalog-spec.md
```diff
@@ -30,6 +30,7 @@ also a valid STAC Catalog.
 | Element         | Type          | Description                                                  |
 | --------------- | ------------- | ------------------------------------------------------------ |
 | stac_version    | string        | **REQUIRED.** The STAC version the catalog implements. STAC versions can be mixed, but please keep the [recommended best practices](../best-practices.md#mixing-stac-versions) in mind. |
+| type            | string        | **REQUIRED.** Set to `Catalog` if this catalog only implements the catalog spec, or `Collection` if it additional meets the [collection](../collection-spec/collection-spec.md) requirements and should be treated by clients as such. |
```
jisantuc, line 33: All of our models in `stac4s` have `Stac`- prefixes so it wouldn't look strange to me if we went that route (re: your [PR comment](https://github.com/radiantearth/stac-spec/pull/971#issuecomment-773821337))

best-practices.md
cholmes: Thanks for sharing your thoughts, even if you talked yourself into it. And I agree with both points - some people may save the itemcollection out in the wild.

catalog-spec/catalog-spec.md
cholmes: Cool, good to know. May be worth including.

m-mohr:
One question that came up during the JSON Schema update:

Can a collection be of type catalog? As it inherits from Catalog, it must allow "Catalog" and "Collection" as type, otherwise, the inheritance as it is doesn't work anymore. But it makes the type field less useful...

cholmes:
We discussed this on the call, and the consensus seemed to be that we should do something where the collection type includes the catalog in it, so that a client can check for 'catalog' with a regex and match both, or can check for the specific.

We should do a bit more research on options, but the one idea was to look to media types for what they do, with like geo+json. The obvious proposal based on what we have would be `Catalog` and `Catalog+Collection`. Though it's sorta half inspired from GeoJSON (with the casing) and half from media types. Open for other ideas.

matthewhanson:
I'm fine with `Catalog+Collection`, but also think that just `Collection` by itself is OK since the spec clearly defines that Collection is a Catalog.

m-mohr:
Using just "Collection" you can't use the Catalog JSON Schema to validate the Collection, which should be possible as it states that a Collection is a Catalog.

kylebarron:
It seems then that the core issue here is that JSONSchema doesn't support true inheritance? In a fuller language like Python/JS, you could presumably subclass `Catalog` to overwrite the `type` key, but that would still fail if you tried to use a `Catalog` object to validate a collection.

cholmes:
And just for posterity, the favorite idea from the STAC call was `Catalog¯\_(ツ)_/¯Collection`

m-mohr:
@kylebarron Although Python and JS would probably allow that, it's not strictly correct inheritance.
If you have a contract in a language that says getType() returns always `Catalog` and you subclass it and return something else, then you break the contract and diverge from what others would expect. You could say getType() just returns a string, but then of course you can't rely on Catalog. And if you define in Catalog that it can return Catalog and Collection, you are mixing thing and you'd need to know all subclasses upfront. If at some point we invent the SuperDuperCatalog that inherits from Catalog, you'd need the implementation of Catalog to allow that. That should be avoided at all costs in OOP. So this is not the fault of JSON Schema IMHO.

jisantuc:
I think there's a little bit of ambiguity around inheritance.

> The STAC Collection specification is **independent** of STAC Items and STAC Catalogs.

(emphasis mine)

But also

> It shares the same fields and therefore every Collection is also a valid Catalog - the JSON structure extends the core Catalog definition.

(admittedly there's much more backing up the second point, but the first is the third sentence in the collection readme)

One thing that I've never been clear about is whether it's important for every collection to be a valid catalog, or whether that's something incidental that's been carried over for a long time. Is that desire from OGC data type relationships? Or can we break it without going out on a limb?

cholmes:
> One thing that I've never been clear about is whether it's important for every collection to be a valid catalog, or whether that's something incidental that's been carried over for a long time. Is that desire from OGC data type relationships? Or can we break it without going out on a limb?

It's definitely not from OGC data type relationships, and I don't think we need to worry about that. To me the important thing is to be able to crawl catalogs and collections in a hierarchy and be able to expect them to operate the same way. But I don't know that it's important that it truly is a valid catalog. Just that it has all the same link / field requirements. But I think I'm ok with changing how we talk about it to just say that they have the same core fields and clients can crawl them equally. 

But it does make some intuitive sense to me to explain that a collection is a catalog with some extras. I do think the other option might just be to drop trying to add the type field. I was imagining we could just have a simple type and it'd work well, but now I'm less sure it's really worth the effort given the constraints.

best-practices.md
```diff
@@ -355,8 +356,10 @@ as they can only have a single representation of their catalog, since the static
 While it is up to the implementor to organize the catalog, it is recommended to arrange it in a way that would make sense 
 for a human to browse a set of STAC Items in an intuitive matter.
 
-The recommendation for static catalogs is to define them using the file name `catalog.json` or `collection.json` to distinguish 
-the Catalog from other JSON type files. In order to support multiple catalogs, the recommended practice 
+Users indicate their intent for a file to be parsed as a Collection or Catalog using the required `type` field on
+each entity. For Collections, this field must have the value `"Collection"`, while for Catalogs, it must have the
```
m-mohr, line 360:
```suggestion
each entity. For Collections, this field must have the value `Collection`, while for Catalogs, it must have the
```

best-practices.md
```diff
@@ -355,8 +356,10 @@ as they can only have a single representation of their catalog, since the static
...
+value `"Catalog"`. Additionally, we recommend for static STACs indicate contents using the filenames `catalog.json`
```
m-mohr, line 361:
```suggestion
value `Catalog`. Additionally, we recommend for static STACs indicate contents using the filenames `catalog.json`
```

best-practices.md
```diff
@@ -399,7 +402,7 @@ ended up doing. Following these recommendations makes for more legible catalogs,
 if you follow these recommendations.
 
 1. Root documents (Catalogs / Collections) should be at the root of a directory tree containing the static catalog.
-2. Catalogs that are not also Collections should be named `catalog.json` and Collections should be named `collection.json`.
+2. Catalogs should be named `catalog.json` and Collections should be named `collection.json`.
```
m-mohr, line 405: Why the removal?

jisantuc: Because no catalogs are also collections

collection-spec/collection-spec.md
```diff
@@ -172,17 +172,17 @@ This is done where there is not a clear official option, or where STAC uses an o
...
+| self    | STRONGLY RECOMMENDED. *Absolute* URL to the location that the Collection file can be found online, if available. This is particularly useful when in a download package that includes metadata, so that the downstream user can know where the data has come from. |
 | root    | URL to the root STAC Catalog or Collection. Collections should include a link to their root, even if it's the root and points to itself. |
 | parent  | URL to the parent STAC Catalog or Collection. Non-root Collections should include a link to their parent. |
 | child   | URL to a child STAC Catalog or Collection. |
-| item    | URL to a STAC Item. All Items linked from a Collection MUST refer back to its Collection with the [`collection` relation type](../item-spec/item-spec.md#relation-types). |
+| item    | URL to a STAC Item. All Items linked from a Collection MUST refer back to its Collection with the [`Collection` relation type](../item-spec/item-spec.md#relation-types). |
```
m-mohr, line 179: That must be lower-cased!

jisantuc: ah woops I thought I fixed all of them -- I had some conflicts to resolve from the capitalization PR but missed some when I went back to fix

jisantuc: e77843a

CHANGELOG.md
kylebarron: It seems like this is still `more easily distinguish them easily`?

overview.md
```diff
@@ -51,19 +51,17 @@ A Catalog is a very simple construct - it just provides links to Items or to oth
...
 also hold other containers (folders / catalogs). 
 
-The Collection specification extends the Catalog spec, so every Collection can serve as a catalog, 
-requiring the same fields. But it has a number of additional fields: license, extent (spatial and 
-temporal), providers, keywords and summaries. Every Item in a Collection links back to their Collection, 
-so clients can easily find fields like the license. Thus every Item implicitly shares the fields 
-described in their parent Collection.
+The Collection specification shares some fields with the catalog spec but has a number of additional fields:
+license, extent (spatial and temporal), providers, keywords and summaries. Every Item in a Collection links
+back to their Collection, so clients can easily find fields like the license.
```
cholmes, line 56: I'm ok with removing the bit about 'item implicitly shares the fields described in their parent collection', but I don't think it's really part of adding type field?

cholmes (commented):
Latest changes look good @jisantuc - made a couple small typo fixes. And had a question about one of the lines that was taken away, though I'm not even sure if you did that or someone else did.

best-practices.md
m-mohr: Oh yeah, I automatically still assume that. Will take some time...

best-practices.md
```diff
@@ -623,3 +626,31 @@ basic geographic filtering from listeners.
...
+
+## How to Differentiate STAC Files
+
+Any tool that crawls a STAC implementation or encounters a STAC file in the wild needs a clear way to determine if it is an Item, 
+Collection, Catalog or [ItemCollection](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-beta.1/fragments/itemcollection) 
+(part of the [STAC API spec](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-beta.1)). As of 1.0.0 this is done primarily
+with the `type` field, and secondarily in Items with `stac_version`, or optionally the `rel` of the link to it.
+
+```shell
+if type is `Collection`
```
m-mohr, line 638: While this works for new STACs, should we still give a hint how to do it for old catalogs? Most tooling will still need this...

best-practices.md
```diff
@@ -623,3 +626,31 @@ basic geographic filtering from listeners.
...
+  => ItemCollection
+else
+  => Invalid (GeoJSON)
+```
+
+When actually crawling a STAC implementation one can also make use of the [relation type](catalog-spec/catalog-spec.md#relation-types
+) (`rel` field) when following a link. If it is an `item` rel type then the file must be a STAC Item. If it is `child`, `parent` or
+`root` then it must be a Catalog or a Collection, and the `type` field can be used to distinguish which one.
+
+In versions of STAC prior to 1.0 the process was a bit more complicated, as there was no `type` field for catalogs and collections.
```
m-mohr, line 654:
```suggestion
In versions of STAC prior to 1.0.0-rc.1 the process was a bit more complicated, as there was no `type` field for catalogs and collections.
```

best-practices.md
```diff
@@ -623,3 +626,31 @@ basic geographic filtering from listeners.
...
+See [this issue comment](https://github.com/radiantearth/stac-spec/issues/889#issuecomment-684529444) for a heuristic that works
+for older STAC versions (to 0.8.0, older than that should not be needed).
```
m-mohr, line 656:
```suggestion
for older STAC versions.
```

Oldest catalogs I can find in STAC Index are 0.6.
As the post also indicates a way out for 0.6, I'd just remove the additional words.

best-practices.md
```diff
@@ -623,3 +626,31 @@ basic geographic filtering from listeners.
 The dynamic STAC API would then listen to the notifications and update its internal datastore whenever new data comes into
 the static catalog. Implementors have had success using AWS Lambda to do a full 'serverless' updating of the elasticsearch
 database, but it could just as easily be a server-based process.
+
+## How to Differentiate STAC Files
+
+Any tool that crawls a STAC implementation or encounters a STAC file in the wild needs a clear way to determine if it is an Item, 
+Collection, Catalog or [ItemCollection](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-beta.1/fragments/itemcollection) 
+(part of the [STAC API spec](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-beta.1)). As of 1.0.0 this is done primarily
```
m-mohr, line 634:
```suggestion
(part of the [STAC API spec](https://github.com/radiantearth/stac-api-spec/tree/v1.0.0-beta.1)). As of 1.0.0-rc.1 this is done primarily
```

best-practices.md
m-mohr, line 638:
```suggestion
if type is 'Collection'
```

best-practices.md
```diff
@@ -623,3 +626,31 @@ basic geographic filtering from listeners.
...
+with the `type` field, and secondarily in Items with `stac_version`, or optionally the `rel` of the link to it.
+
+```shell
+if type is `Collection`
+  => Collection
+else if type is `Catalog`
```
m-mohr, line 640:
```suggestion
else if type is 'Catalog'
```

for consistency

best-practices.md
```diff
@@ -623,3 +626,31 @@ basic geographic filtering from listeners.
...
+  => Catalog
+else if type is 'Feature' and stac_version is defined // stac_version in items is only available since 0.8, check for (stac_version or assets) to support pre-0.8 data
```
m-mohr, line 642:
```suggestion
else if type is 'Feature' and stac_version is defined
```

We only support 1.0.0-rc.1+ anyway.

best-practices.md
```diff
@@ -623,3 +626,31 @@ basic geographic filtering from listeners.
...
+  => Item
+else if type is 'FeatureCollection' and stac_version is defined
+  => ItemCollection
+else
+  => Invalid (GeoJSON)
```
m-mohr, line 647:
```suggestion
  => Invalid (JSON)
```

Not necessarily GeoJSON any longer in this code snipped

best-practices.md
```diff
@@ -623,3 +626,31 @@ basic geographic filtering from listeners.
...
+```
+
+When actually crawling a STAC implementation one can also make use of the [relation type](catalog-spec/catalog-spec.md#relation-types
+) (`rel` field) when following a link. If it is an `item` rel type then the file must be a STAC Item. If it is `child`, `parent` or
+`root` then it must be a Catalog or a Collection, and the `type` field can be used to distinguish which one.
```
m-mohr, line 652: > and the `type` field can be used to distinguish which one.

How to? We don't have distinct media types defined (for a good reason) .

catalog-spec/catalog-spec.md
```diff
@@ -40,6 +40,7 @@ also a valid STAC Catalog.
 | Element         | Type          | Description                                                  |
 | --------------- | ------------- | ------------------------------------------------------------ |
 | stac_version    | string        | **REQUIRED.** The STAC version the Catalog implements. STAC versions can be mixed, but please keep the [recommended best practices](../best-practices.md#mixing-stac-versions) in mind. |
+| type            | string        | **REQUIRED.** Set to `Catalog` if this Catalog only implements the Catalog spec, or `Collection` if it additional meets the [collection](../collection-spec/collection-spec.md) requirements and should be treated by clients as such. |
```
m-mohr, line 43: Why mention Collections here?

collection-spec/collection-spec.md
```diff
@@ -1,13 +1,13 @@
 # STAC Collection Specification
 
 The STAC Collection Specification defines a set of common fields to describe a group of Items that share properties and metadata. The 
-Collection Specification extends the STAC [Catalog Specification](../catalog-spec/catalog-spec.md) with additional fields to 
-describe the whole dataset and the included set of Items. It shares the same fields and therefore every Collection is also a valid Catalog. Collections can have both parent Catalogs and Collections and child Items, Catalogs and Collections. 
+Collection Specification shares some fields with the STAC [Catalog Specification](../catalog-spec/catalog-spec.md) with additional fields to
```
m-mohr, line 4: Should we say it shares all files with different semantics for stac_extensions and type?

collection-spec/collection-spec.md
```diff
@@ -182,7 +182,7 @@ This is done where there is not a clear official option, or where STAC uses an o
 
 A more complete list of possible `rel` types and their meaning in STAC can be found in the [Using Relation Types](../best-practices.md#using-relation-types) best practice. 
 
-**Note:** The STAC Catalog specification requires a link to at least one `item` or `child` Catalog. This is *not* a requirement for Collections, but *recommended*. In contrast to Catalogs, it is **REQUIRED** that Items linked from a Collection MUST refer back to its Collection with the [`collection` relation type](../item-spec/item-spec.md#relation-types).
+**Note:** The STAC Catalog specification requires a link to at least one `item` or `child` Catalog. This is *not* a requirement for Collections, but *recommended*. In contrast to Catalogs, it is **REQUIRED** that Items linked from a Collection MUST refer back to its Collection with the [`Collection` relation type](../item-spec/item-spec.md#relation-types).
```
m-mohr, line 185:
```suggestion
**Note:** The STAC Catalog specification requires a link to at least one `item` or `child` Catalog. This is *not* a requirement for Collections, but *recommended*. In contrast to Catalogs, it is **REQUIRED** that Items linked from a Collection MUST refer back to its Collection with the [`collection` relation type](../item-spec/item-spec.md#relation-types).
```

best-practices.md
cholmes: I think this means the new type field in the JSON, not the type on the link. But obviously that needs to be made more clear.

best-practices.md
```diff
@@ -623,3 +626,31 @@ basic geographic filtering from listeners.
...
+  => Catalog
+else if type is 'Feature' and stac_version is defined
+  => Item
+else if type is 'FeatureCollection' and stac_version is defined
+  => ItemCollection
+else
+  => Invalid (JSON)
+```
+
+When actually crawling a STAC implementation one can also make use of the [relation type](catalog-spec/catalog-spec.md#relation-types
```
philvarner, line 650:
```suggestion
When actually crawling a STAC implementation, one can also make use of the [relation type](catalog-spec/catalog-spec.md#relation-types
```

philvarner, line 650:
```suggestion
When crawling a STAC implementation, one can also make use of the [relation type](catalog-spec/catalog-spec.md#relation-types
```

best-practices.md
```diff
@@ -623,3 +626,31 @@ basic geographic filtering from listeners.
...
+) (`rel` field) when following a link. If it is an `item` rel type then the file must be a STAC Item. If it is `child`, `parent` or
+`root` then it must be a Catalog or a Collection, and the `type` field can be used to distinguish which one.
+
+In versions of STAC prior to 1.0 the process was a bit more complicated, as there was no `type` field for catalogs and collections.
+See [this issue comment](https://github.com/radiantearth/stac-spec/issues/889#issuecomment-684529444) for a heuristic that works
+for older STAC versions.
```
philvarner, line 656: maybe also add mention that Asset#type is the media type, rather than the GeoJSON/STAC type.

cholmes: Ok, added in some more language and links here, to help clarify.

m-mohr (approved):

overview.md
```diff
@@ -165,20 +162,20 @@ each Item and Catalog, as well as ways to achieve that.
 
 ## Collection Overview
 
-A STAC Collection extends the Catalog construct to provide additional metadata to describe the set of Items it
-contains. So every Collection *is* a Catalog, and can be used in the same ways. The required fields are fairly 
+A STAC Collection extends the core fields of the Catalog construct to provide additional metadata to describe the set of Items it
+contains. The required fields are fairly 
 minimal - it inherits the 4 required Catalog fields (id, description, stac_version and links), and adds license
```
matthewhanson, line 167: Not inherit

matthewhanson (approved):

Issue #976, opened 2021-02-08
projection extension best practices?


cholmes:
I'd like to include proj shape and transform in one of the examples, and I realized I have no idea how to generate that.

Which makes me think we should have a best practice recommending the use of 'proj', and why it's good to include shape and transform - so that virtual catalogs don't have to open the geotiff (in my understanding). And how to generate it. 

Is there a rasterio and/or gdal command that can get it for me easily?

kylebarron:
They're both included as part of the `rio info` command. To use a 6-item transform you just discard the last three numbers.

```
> export AWS_REQUEST_PAYER="requester"
> rio info s3://usgs-landsat/collection02/level-2/standard/oli-tirs/2020/072/076/LC08_L2SR_072076_20201203_20201218_02_T2/LC08_L2SR_072076_20201203_20201218_02_T2_SR_B1.TIF | jq
{
  "blockxsize": 256,
  "blockysize": 256,
  "bounds": [
    142185,
    -2673615,
    369915,
    -2442885
  ],
  "colorinterp": [
    "gray"
  ],
  "compress": "deflate",
  "count": 1,
  "crs": "EPSG:32601",
  "descriptions": [
    null
  ],
  "driver": "GTiff",
  "dtype": "uint16",
  "height": 7691,
  "indexes": [
    1
  ],
  "interleave": "band",
  "lnglat": [
    -179.3819753343787,
    -23.115073747911566
  ],
  "mask_flags": [
    [
      "nodata"
    ]
  ],
  "nodata": 0,
  "res": [
    30,
    30
  ],
  "shape": [
    7691,
    7591
  ],
  "tiled": true,
  "transform": [
    30,
    0,
    142185,
    0,
    -30,
    -2442885,
    0,
    0,
    1
  ],
  "units": [
    null
  ],
  "width": 7591
}
```

cholmes:
Sweet, thanks @kylebarron! That's easy. Just updated my example https://github.com/radiantearth/stac-spec/pull/955/commits/1f1023e746f7727abf4e0c0ee6adb0b7d56c733f

Will write this up and add it. If there are other tools (gdal, etc) that others know let me know and I can add them too.

kylebarron:
`gdalinfo` also prints related information, though it doesn't give the transform directly. From the `gdalinfo` output you could construct the transform using the origin and pixel size.

```
[...]

PR #980, opened 2021-02-09
Proj recommendations and formatting

cholmes:
**Related Issue(s):** #976 #822 

**Proposed Changes:**

1. Added best practices section
2. More tools
3. Added h4's instead of bold, for linking.

**PR Checklist:**

- [x]  a CHANGELOG entry is not required.

### DIFF
--- a/extensions/projection/README.md
+++ b/extensions/projection/README.md
@@ -36,7 +36,7 @@ The `proj` prefix is short for "projection", and is not a reference to the PROJ/
 | proj:projjson    | [PROJJSON Object](https://proj.org/specifications/projjson.html) \|null   | PROJJSON object representing the Coordinate Reference System (CRS) that the `proj:geometry` and `proj:bbox` fields represent |
 | proj:geometry    | [Polygon Object](https://geojson.org/schema/Polygon.json)  | Defines the footprint of this Item. |
 | proj:bbox        | \[number]       | Bounding box of the Item in the asset CRS in 2 or 3 dimensions. |
-| proj:centroid    | Centroid Object | Coordinates representing the centroid of the Item (in lat/long)|
+| proj:centroid    | Centroid Object | Coordinates representing the centroid of the Item (in lat/long) |
 | proj:shape       | \[integer]      | Number of pixels in Y and X directions for the default grid |
 | proj:transform   | \[number]       | The affine transformation coefficients for the default grid  |
 
--- a/extensions/projection/README.md
+++ b/extensions/projection/README.md
@@ -40,43 +40,67 @@ The `proj` prefix is short for "projection", and is not a reference to the PROJ/
 | proj:shape       | \[integer]      | Number of pixels in Y and X directions for the default grid |
 | proj:transform   | \[number]       | The affine transformation coefficients for the default grid  |
 
-**proj:epsg** - A Coordinate Reference System (CRS) is the data reference system (sometimes called a
-'projection') used by the asset data, and can usually be referenced using an [EPSG code](http://epsg.io).
+### proj:epsg
+
+A Coordinate Reference System (CRS) is the data reference system (sometimes called a
+'projection') used by the asset data, and can usually be referenced using an [EPSG code](https://en.wikipedia.org/wiki/EPSG_Geodetic_Parameter_Dataset).
 If the asset data does not have a CRS, such as in the case of non-rectified imagery with Ground Control
 Points, `proj:epsg` should be set to null. It should also be set to null if a CRS exists, but for which
-there is no valid EPSG code.
+there is no valid EPSG code. A great tool to help find EPSG codes is [epsg.io](http://epsg.io/).
 
-**proj:wkt2** - A Coordinate Reference System (CRS) is the data reference system (sometimes called a
+### proj:wkt2
+
+A Coordinate Reference System (CRS) is the data reference system (sometimes called a
 'projection') used by the asset data. This value is a [WKT2](http://docs.opengeospatial.org/is/12-063r5/12-063r5.html) string.
 If the data does not have a CRS, such as in the case of non-rectified imagery with Ground Control
 Points, proj:wkt2 should be set to null. It should also be set to null if a CRS exists, but for which
 a WKT2 string does not exist.
 
-**proj:projjson** - A Coordinate Reference System (CRS) is the data reference system (sometimes called a
+### proj:projjson
+
+A Coordinate Reference System (CRS) is the data reference system (sometimes called a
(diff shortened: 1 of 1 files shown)


extensions/projection/README.md
```diff
@@ -40,43 +40,67 @@ The `proj` prefix is short for "projection", and is not a reference to the PROJ/
 | proj:shape       | \[integer]      | Number of pixels in Y and X directions for the default grid |
 | proj:transform   | \[number]       | The affine transformation coefficients for the default grid  |
 
-**proj:epsg** - A Coordinate Reference System (CRS) is the data reference system (sometimes called a
-'projection') used by the asset data, and can usually be referenced using an [EPSG code](http://epsg.io).
+### proj:epsg
```
m-mohr, line 43: Sometimes these are h4 (with another h3 "Additional Properties") now and sometimes h3. We should be consistent here.

cholmes: Good catch, will update.

extensions/projection/README.md
```diff
@@ -40,43 +40,70 @@ The `proj` prefix is short for "projection", and is not a reference to the PROJ/
...
+\[xmin, ymin, xmax, ymax], \[left, down, right, up], or \[west, south, lowest, east, north, highest]. The length of the array must be 2\*n where n is the number of dimensions. The array contains all axes of the southwesterly most extent followed by all axes of the northeasterly most extent specified in Longitude/Latitude or Longitude/Latitude/Elevation based on [WGS 84](http://www.opengis.net/def/crs/OGC/1.3/CRS84). When using 3D geometries, the elevation of the southwesterly most extent is the minimum elevation in meters and the elevation of the northeasterly most extent is the maximum in meters.
+
+#### proj:centroid
+
+Coordinates representing the centroid of the item in the asset data CRS.  Coordinates are defined in latitude and longitude, even if 
+the data coordinate system does not use lat/long. This is useful as the projection from the native CRS into GeoJSON (lat/long) can
```
cholmes, line 87: @philvarner - is this additional information correct? I was quite confused by the idea of the centroid of the asset data in CRS that needed to be represented in lat/long, but @matthewhanson explained to me what I tried to explain. But he wasn't totally sure. Just want to help people like me to understand why you would use this.

philvarner: I'm not sure the warping explanation is correct -- I'll look into that. In our implementation, we added this for a very practical reason -- in Elasticsearch, a field with an object containing `lat` and `lon` can be put on a Kibana Map. There may be other similar uses.

philvarner: on further reflection, the warping explanation is correct -- the centroid will be in a different place after reprojection to 4326. (though, I'm not entirely sure in my impl that I'm not reprojecting and taking the centroid instead of taking the centroid and reprojecting it. 😬 )

extensions/projection/README.md
```diff
@@ -94,3 +121,38 @@ This object represents the centroid of the Item Geometry.
 | ------------------- | ------ | ------------------------------------------------------------ |
 | lat                 | number | The latitude of the centroid.  |
 | lon                 | number | The longitude of the centroid. |
+
+## Best Practices
+
+There are clearly a number of projection extension fields, with potentially overlapping functionality. This section attempts to
```
philvarner, line 127: ==> "There are a number of projection extension fields with potentially overlapping functionality."

extensions/projection/README.md
```diff
@@ -94,3 +121,38 @@ This object represents the centroid of the Item Geometry.
...
 | lat                 | number | The latitude of the centroid.  |
 | lon                 | number | The longitude of the centroid. |
+
+## Best Practices
+
+There are clearly a number of projection extension fields, with potentially overlapping functionality. This section attempts to 
+give an overview of which ones you should consider using. They fit into three general categories:
+
+- **Description of the coordinate reference system:** [EPSG code](#projepsg) is the default, but it is just a reference to known
+projection information. [WKT2](#projwkt2) and [PROJJSON](#projprojjson) are two options to fully describe the projection information.
```
philvarner, line 131: + "For example, the MODIS Sinusoidal projection does not have an EPSG code, but can be described using WKT2 or PROJJSON"

extensions/projection/README.md
```diff
@@ -94,3 +121,38 @@ This object represents the centroid of the Item Geometry.
...
+This is most typically done for projections that aren't available or fully described in the [EPSG Registry](https://epsg.org/).
```
philvarner, line 132: - "most"

extensions/projection/README.md
```diff
@@ -94,3 +121,38 @@ This object represents the centroid of the Item Geometry.
...
+- **Description of the native geometry information:** STAC requires the geometry and bounding box, but they are only available
+in lat/long (EPSG:4326). But most remote sensing data does not come in that projection, so it is often useful for clients to have 
+the geometry information ([geometry](#projgeometry), [bbox](#projbbox), [centroid](#projcentroid)) in the coordinate reference system
+of the asset's data, so it doesn't have to reproject (which can be lossy and takes time). 
+- **Information to enable cataloging of data without opening assets:** Often it is useful to be able to construct a 'virtual layer',
+like GDAL's [VRT](https://gdal.org/drivers/raster/vrt.html) without having to open the actual imagery file. [shape](#projshape) and
+[transform](#projtransform) together with the core description of the CRS provide enough information about the size and shape of
+the data in the file so that tools don't have to open it.
+
+None of these is truly necessary for 'search' of data, the main use case of STAC. But all enable more 'cloud native' use of data,
```
philvarner, line 142: "None of these are necessary for..."

philvarner (commented):
Few wordsmithing suggestions, but otherwise looks good.

extensions/projection/README.md
cholmes: Ah, so you just did it so you'd have a centroid in lat/long. I think the thing that confused me was the description said: 'Coordinates representing the centroid of the Item in the asset CRS'. And it directly followed proj:bbox, where the point of that *is* to have the bbox in the asset. If the main case is just to have a lat/long centroid I think maybe we leave out the emphasis on the 'asset crs'. Like have the description be 'Coordinates representing the centroid of the Item, in lat/long', and the detail be about how having the centroid lets you easily plot things on less sophisticated mapping systems. And I'd mention last that it should be in the CRS / before projection.

The way it was written definitely had me overthinking the use case, and it sounds like the use case is much simpler - get a lat/lon for kibana.

cholmes: Went ahead and just made the changes, as this makes a lot more sense. I was just trying to figure out the purpose of the field and trying to explain it, and missed the obvious one.

matthewhanson (approved):

jbants (approved):

Issue #965, opened 2021-02-03
Make summaries 'strongly recommended' in Collections


m-mohr:
We've recently discussed how important summaries for discovery of data, but on the other hand most people actually dont implement summaries. Therefore, I propose to make summaries required in Collections with a minimum amount of proerties set to 1 in JSON Schema. Summaries in Catalogs would still be optional! This issue is to spur discussions.

cholmes:
In general I like the thought a lot, and was all for it when I first thought about it. They're important to have, and making them required is a good nudge towards getting them used more.

On the other hand I think we're already introducing more change than I was hoping we would for RC1. If we were to do another beta I'd probably be more into it. But I think it's really important for STAC momentum to get something stable soon. 

If all the tooling supported making summaries very easily then that would make it easier to 'require' as well. Admittedly that's a bit 'chicken and egg' - obviously the tools will figure out how to support it if the spec makes it required. But I worry that many may just do the 'bare minimum'. 

So my current thinking is really prioritize the tooling to be there to make it very easy to create summaries, and to put stronger recommendations in the spec, and add a best practices section that also really emphasizes the importance. I think this can be done with #820 - I wasn't planning quite so strong recommendations, but easily can. And then in any funding of [the ecosystem](https://github.com/stac-utils/stac-ecosystem) I can prioritize funding their support. 

It might be interesting to also think about ways to make summaries look 'cool' in STAC Browser, like do some graphical representations of some of them. Like gsd could have a little line graph absolute (or log?) scale where you'd see the resolution relative to others. And the various view angles could actually show the range of possibiliites, etc.

m-mohr:
Hmm, as we are going to do the RC1 soon, it's the last chance to get it in any time soon. Collections and items anyway need to be regenerated with the "type" change. Thus, it seems the right time to solve the chicken and egg issue by just requiring it in the spec. If you want to get funding for those functions anyway, isn't that even more an argument to require it? I don't understand the stability issue to be honest. How does requiring a field make it less stable after we are breaking all collections anway?

jjrom:
[...]

Issue #820, opened 2020-05-29
Explain summaries more


cholmes:
Whenever I read the summaries section I'm always a bit unsure how to actually use them. It doesn't feel like quite enough information in the spec. 

I think having them more prominent in the examples would help, and a good best practices section as well.

m-mohr:
What exactly is missing for you? What should be covered in a best practice? There's already a bit on what to include and what not.

m-mohr:
I need a response so that I can work on it for beta.2. Otherwise I have to move it to the next release. @cholmes

m-mohr:
@cholmes This has been addressed with #985 or is there anything missing?

cholmes:
Nope, nothing missing - was just waiting for that one to close. Thanks!

PR #985, opened 2021-02-10
Summaries updates - to 'strongly recommended' and with a best practice

cholmes:
**Related Issue(s):** #820 #965 

**Proposed Changes:**

1. Made summaries to be 'strongly recommended' and explained them a bit more
2. added a best practices section on summaries.

**PR Checklist:**

- [x] This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec), and I have opened issue https://github.com/radiantearth/stac-api-spec/issues/101 to track the change.


best-practices.md
```diff
@@ -297,6 +298,42 @@ is strongly recommended that Catalogs don't contain differently versioned Items
...
+Any field that is a range of data (like numbers or dates) is a great candidate to summarize, to give people a sense what values
+the data might be. For example in overhead imagery, a 
+[`view:off_nadir`](extensions/view/README.md#item-properties-and-item-asset-fields) with a range of 0 to 3 would tell people this 
+imagery is all pretty much straight down, while a value of 15 to 40 would tell them that it's oblique imagery, or 0 to 60 that it's 
+a collection with lots of different look angles. 
+
+Fields that have only one or a handful of values are also great to summarize. Collections with a single satellite may
+use a single [`gsd`](item-spec/common-metadata.md#instrument) field in the summary, and it's quite useful for users to know
+that all data is going to be the same resolution. Similarly it's useful to know the names of all the 
+[`platform`s](item-spec/common-metadata.md#instrument) that are used in the collection.
```
philvarner, line 317: this combination of backtick formatting, plural, and link looks odd ==> maybe "`platform` objects"

best-practices.md
```diff
@@ -297,6 +298,42 @@ is strongly recommended that Catalogs don't contain differently versioned Items
...
+[`view:off_nadir`](extensions/view/README.md#item-properties-and-item-asset-fields) with a range of 0 to 3 would tell people this 
+imagery is all pretty much straight down, while a value of 15 to 40 would tell them that it's oblique imagery, or 0 to 60 that it's 
+a collection with lots of different look angles. 
+
+Fields that have only one or a handful of values are also great to summarize. Collections with a single satellite may
+use a single [`gsd`](item-spec/common-metadata.md#instrument) field in the summary, and it's quite useful for users to know
+that all data is going to be the same resolution. Similarly it's useful to know the names of all the 
+[`platform`s](item-spec/common-metadata.md#instrument) that are used in the collection. 
+
+It makes much less sense to summarize fields that have numerous different discrete values that can't easily be represented
```
philvarner, line 319: "It is less useful to"

best-practices.md
```diff
@@ -297,6 +298,42 @@ is strongly recommended that Catalogs don't contain differently versioned Items
...
+in a range. These will mostly be string values, when there aren't just a handful of options. For example if you had a 
+'location' field that gave 3 levels of administrative region (like 'San Francisco, California, United States') to help people
+understand more intuitively where a shot was taken. If your collection has millions of items, or even hundreds, you don't want
+to include all those values as a summary. 
+
+We do highly recommend including an [`eo:bands`](extensions/eo/README.md#eobands) summary if your Items implement `eo:bands`, 
+especially if it represents just one satellite or constellation. This should be a union of all the potential bands that you 
+have in assets. It is ok to just put the summary at the Collection level without putting an explicit `eo:bands` summary at the 
+`properties` level, since that is optional. This gives users of the Collection a sense of the sensor capabilities without having
+to dig into particular items.
```
philvarner, line 329: "without having to examine specific items or aggregate across every item"

philvarner (commented):
few wordsmithing comments, looks good!

cholmes:
Wordsmith edits incorporated, thanks!

best-practices.md
```diff
@@ -297,6 +298,42 @@ is strongly recommended that Catalogs don't contain differently versioned Items
...
+It is less useful to summarize fields that have numerous different discrete values that can't easily be represented
+in a range. These will mostly be string values, when there aren't just a handful of options. For example if you had a 
+'location' field that gave 3 levels of administrative region (like 'San Francisco, California, United States') to help people
+understand more intuitively where a shot was taken. If your collection has millions of items, or even hundreds, you don't want
+to include all those values as a summary. 
+
+We do highly recommend including an [`eo:bands`](extensions/eo/README.md#eobands) summary if your Items implement `eo:bands`, 
+especially if it represents just one satellite or constellation. This should be a union of all the potential bands that you 
+have in assets. It is ok to only add the summary at the Collection level without putting an explicit `eo:bands` summary at the 
+`properties` level, since that is optional. This gives users of the Collection a sense of the sensor capabilities without having
```
matthewhanson, line 328: Explicitly state "at the `properties` level of the Item", since someone new might get confused about what `properties` are being referred to here and if collections have properties.

matthewhanson (approved):

jbants (approved):

best-practices.md
```diff
@@ -297,6 +298,42 @@ is strongly recommended that Catalogs don't contain differently versioned Items
...
+that all data is going to be the same resolution. Similarly it's useful to know the names of all the 
+[`platform` values](item-spec/common-metadata.md#instrument) that are used in the collection. 
+
+It is less useful to summarize fields that have numerous different discrete values that can't easily be represented
+in a range. These will mostly be string values, when there aren't just a handful of options. For example if you had a 
+'location' field that gave 3 levels of administrative region (like 'San Francisco, California, United States') to help people
+understand more intuitively where a shot was taken. If your collection has millions of items, or even hundreds, you don't want
+to include all those values as a summary. 
+
+We do highly recommend including an [`eo:bands`](extensions/eo/README.md#eobands) summary if your Items implement `eo:bands`,
```
m-mohr, line 325: Did we somewhere explain that some arrays can be merged and some shouldn't be merged? I thought we had, but I can't find it right now.
Like eo:bands makes sense to be merged, but on the other hand proj:shape shouldn't be merged at all in the summary.

m-mohr (changes requested):
One minor thing to add.

best-practices.md
cholmes: Yeah, that does sound familiar. I'll try to find it. Though do we have any guidance as to what can be merged and what can't?

cholmes: Didn't find it, but made an attempt to cover it. Let me know if I got anything wrong in it.

best-practices.md
```diff
@@ -297,6 +298,51 @@ is strongly recommended that Catalogs don't contain differently versioned Items
...
+
+* It is less useful to summarize fields that have numerous different discrete values that can't easily be represented
+in a range. These will mostly be string values, when there aren't just a handful of options. For example if you had a 
+'location' field that gave 3 levels of administrative region (like 'San Francisco, California, United States') to help people
+understand more intuitively where a shot was taken. If your collection has millions of items, or even hundreds, you don't want
+to include all those values as a summary. 
+
+* Fields that consist of arrays are more of a judgement call. For example [`instruments`](item-spec/common-metadata.md#instrument)
+is straightforward and recommended, as the elements of the array are a discrete set of options. On the other hand 
+[proj:transform](extensions/projection/README.md#projtransform) makes no sense to summarize, as the union of all the values
```
m-mohr, line 329:
```suggestion
[`proj:transform`](extensions/projection/README.md#projtransform) makes no sense to summarize, as the union of all the values
```

best-practices.md
```diff
@@ -297,6 +298,51 @@ is strongly recommended that Catalogs don't contain differently versioned Items
...
+that all data is going to be the same resolution. Similarly it's useful to know the names of all the 
+[`platform` values](item-spec/common-metadata.md#instrument) that are used in the collection. 
+
+* It is less useful to summarize fields that have numerous different discrete values that can't easily be represented
+in a range. These will mostly be string values, when there aren't just a handful of options. For example if you had a 
+'location' field that gave 3 levels of administrative region (like 'San Francisco, California, United States') to help people
+understand more intuitively where a shot was taken. If your collection has millions of items, or even hundreds, you don't want
+to include all those values as a summary. 
+
+* Fields that consist of arrays are more of a judgement call. For example [`instruments`](item-spec/common-metadata.md#instrument)
```
m-mohr, line 327: Maybe add eo:bands, too? That's the second most prominent example and probably one that people would think about.

best-practices.md
m-mohr: Looks good/better. We could maybe link to the stac-fields json file and say that there are indicatiors for the array merge included?

m-mohr: Or do you think it would make sense to actually give the array merge details for summaries directly in the spec for each of the array-fields? Could be useful indeed.

cholmes: I'm more inclined to link to stac-fields json. I'm not sure that we want to lock-in exactly how people should do array merges. To me the core spec should be flexible, with tooling/best practices explaining practically how most people do things. This falls into that camp for me.

best-practices.md
cholmes: I had thought that having the following paragraph talk explicitly about eo:bands was sufficient, as I think people will read the whole section if they are actually thinking about it. But can add here too if you think it's important.

m-mohr: I just realized we have a "false" statement here: The collection spec says:

> If the field summarizes an array (e.g. instruments), the field's array elements of each Item must be merged to a single array with unique elements.

So it's not "a judgement call" at the moment. The current spec basically don't work for data types that use arrays with some hidden sematics, like "proj:shape", where it feels like the structure should be object as it's not really a list.

m-mohr: I've looked into the arrays we define in extensions as new field and except proj, this requirement works for all arrays, I think. I'm not 100% sure about label extension, but it seems okay. So it's just proj, but for those fields I'd say (1) they should not have been defined as arrays - they are a better fit for an object - and (2) they shouldn't get summarized anyway.

cholmes: I think I understand, and agree with the logic but what's the next step? What changes do I need to make on this PR?

best-practices.md
m-mohr: I've read it again and indeed there's nothing to do here, I had understood the "judgement call" differently before.

best-practices.md
m-mohr: I've read it again and indeed there's nothing to do here, I had understood the "judgement call" differently before.

m-mohr (approved):
Thanks, looks good now. I think we can do the stac-fields link later, once it's stable (see #1004).

Issue #966, opened 2021-02-03
Best practice chapter on 'roles' of assets


cholmes:
In adding #945 we realized it'd be good to have a best practice section discussing how to use those, when to put in your own roles, what the recommended sizes are (like X% of image size for thumbnail, max of y by z pixels), clarify what the existing roles mean / are used for, etc.

PR #989, opened 2021-02-12
Roles best practice

cholmes:
**Related Issue(s):** #966 

**Proposed Changes:**

1. Added best practice section on asset roles
2. Made a list of common roles in best practice, borrowing many from card4l
2. Moved visual and overview roles out to be in the best practice, as they aren't actually used that much today.

**PR Checklist:**


cholmes:
Still some more to do here, like adding a bit more on the common roles and when / how to use them. But I think it's mostly done, any review is appreciated. As well as other 'role' ideas, I think a long list makes sense.

best-practices.md
```diff
@@ -202,12 +288,72 @@ register the media type with IANA, so that other STAC clients can find it easily
 [acceptable to not register](https://stackoverflow.com/questions/29121241/custom-content-type-is-registering-with-iana-mandatory) 
 it. It is relatively easy to [register](https://www.iana.org/form/media-types) a `vnd` media type.
 
-## Static and Dynamic Catalogs
+### Asset Roles
+
+[Asset roles](item-spec/item-spec.md#asset-roles) are used to describe what each asset is used for. They are particular useful 
+when several assets have the same media type, such as when an Item has a multispectral analytic asset, a 3-band full resolution 
+visual asset, a downsampled overview, and an cloud mask, all stored as Cloud Optimized GeoTIFF's. It is recommended to use a role
```
philvarner, line 295:
```suggestion
visual asset, a down-sampled preview asset, and a cloud mask asset, all stored as Cloud Optimized GeoTIFF (COG) images. It is recommended to use a role
```

best-practices.md
```diff
@@ -202,12 +288,72 @@ register the media type with IANA, so that other STAC clients can find it easily
...
+
+In addition to the thumbnail, data and overview [roles listed](item-spec/item-spec.md#asset-role-types) in the Item spec, there
+are a number of roles that are emerging in practice, but don't have enough widespread use to justify standardizing them. So if
+you want to re-use other roles then try to find them on the list below, and also feel free to suggest more to include here.
+
+#### List of Asset Roles
+
+| Role Name | Description                                                                           |
+| --------- | ------------------------------------------------------------------------------------- |
+| overview  | An asset that represents a possibly larger view than the thumbnail of the Item, for example, a true color composite of multi-band data. |
```
philvarner, line 307: looks like `thumbnail` is missing from this list, but mentioned in the text above

best-practices.md
```diff
@@ -202,12 +288,72 @@ register the media type with IANA, so that other STAC clients can find it easily
...
+| visual    | An asset that is a full resolution version of the data, processed for visual use (RGB only, often sharpened ([pan-sharpened](https://en.wikipedia.org/wiki/Pansharpened_image) and/or using an [unsharp mask](https://en.wikipedia.org/wiki/Unsharp_masking))). |
+| reflectance | An asset the provides [reflectance](https://www.l3harrisgeospatial.com/Support/Self-Help-Tools/Help-Articles/Help-Articles-Detail/ArtMID/10220/ArticleID/19247/3377) values, instead of just radiance. |
+| temperature | An asset that provides actual temperature measurements. |
+| date | An asset that provides per-pixel acquisition timestamps, typically serving as metadata to another asset |
+| saturation | Points to a file that indicates where pixels in the input spectral bands are saturated. |
+| cloud | Points to a file that indicates whether a pixel is assessed as being cloud |
+| cloud-shadow | Points to a file that indicates whether a pixel is assessed as being cloud shadow. |
```
philvarner, line 314: do we have the convention of using kebab-case for enums described anywhere in the spec?

best-practices.md
```diff
@@ -202,12 +288,72 @@ register the media type with IANA, so that other STAC clients can find it easily
...
+[Asset roles](item-spec/item-spec.md#asset-roles) are used to describe what each asset is used for. They are particular useful 
+when several assets have the same media type, such as when an Item has a multispectral analytic asset, a 3-band full resolution 
+visual asset, a downsampled overview, and an cloud mask, all stored as Cloud Optimized GeoTIFF's. It is recommended to use a role
+for every asset available, and just pick a sensible name for the role if there is not a clear one in the [Asset Role 
+Types](item-spec/item-spec.md#asset-role-types). You can also propose new roles to the specification by raising an issue/PR. 
+
+In addition to the thumbnail, data and overview [roles listed](item-spec/item-spec.md#asset-role-types) in the Item spec, there
+are a number of roles that are emerging in practice, but don't have enough widespread use to justify standardizing them. So if
+you want to re-use other roles then try to find them on the list below, and also feel free to suggest more to include here.
+
```
philvarner, line 302: Maybe also mention their use in Extensions -- for example, we use (proprietarily) a role of `labels` on assets in Items using the Label Extension, and there's discussion of changing that spec to use this role instead of requiring the asset be named 'labels'

best-practices.md
```diff
@@ -202,12 +288,72 @@ register the media type with IANA, so that other STAC clients can find it easily
...
+#### List of Asset Roles
```
philvarner, line 303: maybe this should be in item-spec.md?

philvarner: oh, i see -- maybe make it clear that these are ideas for some, but aren't part of the blessed set of them in the spec.

best-practices.md
m-mohr: No, I made this decision arbitrarily in the CARD4L extensions (where this list originates from).

Having that said, I'm not sure we should list these all here, but limit to things that seem generally used like cloud-cover, cloud-shadow, snow/ice, land/water, ... 
The issue with this list is that is only captures a small subset of use cases and could become **very* long or always incomplete. Some things are also very specific like date, date-offset, noise-power, gamma-sigma... backscatter, prd, covmat are specific products, but by no mean complete.

best-practices.md
```diff
@@ -202,12 +288,72 @@ register the media type with IANA, so that other STAC clients can find it easily
...
+| visual    | An asset that is a full resolution version of the data, processed for visual use (RGB only, often sharpened ([pan-sharpened](https://en.wikipedia.org/wiki/Pansharpened_image) and/or using an [unsharp mask](https://en.wikipedia.org/wiki/Unsharp_masking))). |
+| reflectance | An asset the provides [reflectance](https://www.l3harrisgeospatial.com/Support/Self-Help-Tools/Help-Articles/Help-Articles-Detail/ArtMID/10220/ArticleID/19247/3377) values, instead of just radiance. |
+| temperature | An asset that provides actual temperature measurements. |
+| date | An asset that provides per-pixel acquisition timestamps, typically serving as metadata to another asset |
+| saturation | Points to a file that indicates where pixels in the input spectral bands are saturated. |
+| cloud | Points to a file that indicates whether a pixel is assessed as being cloud |
+| cloud-shadow | Points to a file that indicates whether a pixel is assessed as being cloud shadow. |
+| snow-ice | Points to a file that indicates whether a pixel is assessed as being snow/ice or not. |
+| land-water | Points to a file that indicates whether a pixel is assessed as being snow/ice or not land or water. |
```
m-mohr, line 316:
```suggestion
| land-water | Points to a file that indicates whether a pixel is assessed as being land or water. |
```

best-practices.md
```diff
@@ -202,12 +288,72 @@ register the media type with IANA, so that other STAC clients can find it easily
...
+| terrain-shadow | Points to a file that indicates whether a pixel is not directly illuminated due to terrain shadowing. |
+| terrain-occlusion | Points to a file that indicates whether a pixel is not visible to the sensor due to terrain occlusion during off-nadir viewing. |
+| terrain-illumination | Points to a file with coefficients used for terrain illumination correction are provided for each pixel. |
+| mask | Points to a data mask file | 
+| contributing-area | Points to the normalized scattering area file (typically used in SAR). |
+| local-incidence-angle | Points to the local incidence angle file (typically used in SAR). |
+| ellipsoid-incidence-angle | Points to the ellipsoid incidence angle file (typically used in SAR). |
+| noise-power | Points to the noise power file (typically used in SAR). |
+| gamma-sigma | Points to the gamma-sigma file (typically used in SAR). |
+| date-offset | Points to the date-offset file (typically used in SAR). |
```
m-mohr, line 330: typically used in SAR is not really true. Although it originates in CARD4L SAR, this is rarely used in SAR and more related to composites.

m-mohr: Also, it feels like SAR is a second class citizen here as there is a "(typically used in SAR)", but not a "(typically used in Optical/EO)" or so for the ones that refer to spectral assets.

best-practices.md
```diff
@@ -2,20 +2,24 @@
...
+  * [STAC on the Web](#stac-on-the-web)
+* **[Item Best Practices](#item-practices)**
+  * [Field and ID formatting](#field-and-id-formatting)
+  * [Field selection and Metadata Linking](#field-selection-and-metadata-linking)
+  * [Datetime selection](#datetime-selection)
+  * [Unlocated Items](#unlocated-items)
+  * [Representing Vector Layers in STAC](#representing-vector-layers-in-stac)
+* **[Asset Best Practices](#asset-practices)**
+  * [Common Use Cases of Additional Fields for Assets](#common-use-cases-of-additional-fields-for-assets)
+  * [Working with Media Types](#working-with-media-types)
```
m-mohr, line 16: Also applies to links?!

best-practices.md
```diff
@@ -2,20 +2,24 @@
...
+* **[Catalog Best Practices](#catalog-practices)**
+  * [Static and Dynamic Catalogs](#static-and-dynamic-catalogs)
+  * [Catalog Layout](#catalog-layout)
+  * [Use of Links](#use-of-links)
```
m-mohr, line 20: Use of Links and Using Relation Types seems not so Catalog specific.

best-practices.md
emmanuelmathot: cant we have gamma, sigma and beta in separate roles?

best-practices.md
```diff
@@ -202,12 +288,72 @@ register the media type with IANA, so that other STAC clients can find it easily
...
+| sun-elevation | Points to a file with per-pixel sun elevation angles. |
+| terrain-shadow | Points to a file that indicates whether a pixel is not directly illuminated due to terrain shadowing. |
+| terrain-occlusion | Points to a file that indicates whether a pixel is not visible to the sensor due to terrain occlusion during off-nadir viewing. |
+| terrain-illumination | Points to a file with coefficients used for terrain illumination correction are provided for each pixel. |
+| mask | Points to a data mask file | 
+| contributing-area | Points to the normalized scattering area file (typically used in SAR). |
+| local-incidence-angle | Points to the local incidence angle file (typically used in SAR). |
+| ellipsoid-incidence-angle | Points to the ellipsoid incidence angle file (typically used in SAR). |
+| noise-power | Points to the noise power file (typically used in SAR). |
+| gamma-sigma | Points to the gamma-sigma file (typically used in SAR). |
```
emmanuelmathot, line 329: related to the SAR, I would the following specific roles:
- amplitude (GRD)
- magnitude (SLC)
- coherence
- phase
- sigma0
- gamma0
- beta0
- displacement

emmanuelmathot (commented):
related to the SAR, I would the following specific roles:

    amplitude (GRD)
    magnitude (SLC)
    coherence
    phase
    sigma0
    gamma0
    beta0
    displacement

matthewhanson:
I like most of the roles added in the new list, but rather than include some really specific and little used SAR roles, what about just including those with the SAR extension as recommended roles?  

There might be other roles tied to specific extensions, but SAR always seems to be the main source of lots of extra complicating fields. 
I'd like to see the SAR specific ones removed and added to the SAR extension along with @emmanuelmathot's suggestions.

matthewhanson:
Mentioned in mtg the other day, but a common asset among scientific geospatial data are plots or other graphics. These might indicate a fit of a curve to generate the main derived products, QC summary info, or ancillary info about applied corrections.

Would `graphic` be a good name for this?

m-mohr:
> related to the SAR, I would the following specific roles:

And that's why it's a bad idea to copy/past just from CARD4L: Some entries only make sense in combination with the CARD4L spec. What is meant with sigma-gamma was the sigma-gamma ratio and not a specific sigma or gamma asset. 
Do we actually need roles for those files? Isn't it better to have a SAR field for the values above and then use data? How often are these really combined so that roles make sense?

m-mohr:
> I like most of the roles added in the new list, but rather than include some really specific and little used SAR roles, what about just including those with the SAR extension as recommended roles?

But then I'm wondering why not add all the cloud related stuff also to the EO extension? cloud_cover is also in eo and thus listing the roles there makes more sense to me.

m-mohr:
> Would `graphic` be a good name for this?

Is this an asset or a link? I think in CARD4L we somehow imply to use a link with rel type `related`...

emmanuelmathot:
> Do we actually need roles for those files? Isn't it better to have a SAR field for the values above and then use data? How often are these really combined so that roles make sense?

Ok but then the rationale should also apply for optical data. For instance, `reflectance` should be part of the `eo` extension. if the `roles` field "are used to describe what each asset is used for.", then being exhaustive to be sure we share common terms to describe the same measurements is not useless.

best-practices.md
cholmes: Cool. I'm up for a more curated list. But I do want us to list a number of potentially relevant ones - I just didn't have enough knowledge of all the options to cut it down. I do think it's ok if it's a pretty long list, and if it's incomplete, just so people don't end up making their own things that are all slightly different.

best-practices.md
cholmes: Sounds good. Will add `labels` to the list, and mention extensions.

best-practices.md
cholmes: Sounds good, will work to clarify.

best-practices.md
cholmes: Cool, will update. And can add typically used in Optical/EO, I realized halfway through it'd be valuable to clarify that.

best-practices.md
cholmes: It does, but that whole section is really focused on asset media types. I can rename the title to 'working with asset media types'. Or could move it, but not sure where - I'm mostly looking for the best way to 'group' things to help people navigate a bit. Links cuts across all of them...

best-practices.md
cholmes: I think it's the place they make the most sense, as it's where you do the majority of the linking. These are meant to just be rough groupings, not 'this only applies here'. I suppose we could make a 'general STAC best practices' section, but it's always going to be a bit fuzzy where to group things.

cholmes:
>> I like most of the roles added in the new list, but rather than include some really specific and little used SAR roles, what about just including those with the SAR extension as recommended roles?

> But then I'm wondering why not add all the cloud related stuff also to the EO extension etc?

Sure, that makes sense to me. I think my only small hesitation is that these are at the level of 'best practice', not 'standard', and in the extension feels a bit more standard. But I think we have talked about a 'best practices' section in extensions, so I can just use this to start those. Would want to link from the overall best practice to the eo and sar roles sections, but that seems reasonable.

matthewhanson:
This makes sense to me too (to put most of these in EO/SAR extension).

So with that I would say keep the roles here as the most generic ones:
**thumbnail**
**overview**
**date**
**graphic** - supporting plot, illustration, or graph associated with the Item

Plus **data** and **metadata** which are in the item-spec but didn't make it here (I think anything in item-spec should be included here)

Maybe **saturation** ?
I can also include including general masks here: **snow-ice**, **land-water**, **mask**

For the others:

**EO Extension**
reflectance
temperature
cloud
cloud-shadow

**View Extension**
azimuth
sun-azimuth
sun-elevation
terrain-shadow
terrain-occlusion
terrain-illumination


**SAR Extension**
contributing-area
local-incidence-angle
noise-power
gamma-sigma
date-offset
backscatter
covmat
prd

m-mohr:
> gamma-sigma

Please note that this has been renamed in CARD4L for clarity reasons to gamma-sigma-ratio

matthewhanson:
I agree `mask` is too generic, what does it represent in CARD4L, nodata?

`nodata-mask` would be good for the general Best Practices

m-mohr:
regarding mask:

It's from the SAR spec and includes

- Valid data
- Invalid data
- No-data

Optionally:
- Layover (included as invalid data in Threshold)
- Radar shadow (included as invalid data in Threshold)
- Ocean water, etc.

matthewhanson:
Differentiating between Invalid and Nodata doesn't seem useful, IMHO, does CARD4L require a difference?  Without it this would just be a nodata-mask

m-mohr:
I feel like no-data-mask doesn't fully capture it as it may contain other details and another term would be good. In CARD4L it's called "data mask", maybe that's the better term? On the other hand, we can define whatever we think is useful here, I don't need to adopt it in CARD4L (and I think I'll change it to `data-mask` there).

philvarner:
Would having values related to derived products make sense? E.g., `ndvi`, `evi`, etc? Though that feels like `roles` is more just becoming `tags`.  This is also related to something I think we need in the recommendations for extensions -- that extensions define recommended/required roles to have rather than asset names (looking at you, Label).

m-mohr:
Yeah, maybe we are going over the top here? We should think about use cases for every role and how that would actually help.

matthewhanson:
Adding in those types of products does seem like we are expanding roles, although one could argue that the various masks are just like products.

Not opposed to adding in roles to represent products like ndvi, but I'd say let's defer for now.

`data-mask` is a fine (and probably better) alternative to `nodata-mask`

cholmes:
> Would having values related to derived products make sense? E.g., ndvi, evi

This makes less sense to me. If you are producing an ndvi asset then imho you should model that as a new STAC Item, with 'derived_from' to the one that has the asset. To me the asset roles should be things that make sense in a 'bundle'.

cholmes:
Also, I think the grouping makes sense, but I think it's really valuable to have one location in best practices where you can see a list of all the roles in use, instead of having to click into many extensions. I'm not opposed to also listing these in extensions, but I think the priority for me is to get a list. Happy to 'classify' that list by the rubric above, so that 'eo' has the same 'qualifier' as 'sar'.

cholmes:
Ok, made a go at the 'master' table in best practices. If people want I can add best practice sections to each of the extensions with proposed roles. Don't love the wording on the 'specified in' and the ones listed there, but just wanted to get something in. Suggestions appreciated.

I also reverted the grouping / reorg of best practice section. DIdn't think it'd be worthy of comments, but looks like there's some, so I can make it its own PR and we can discuss.

emmanuelmathot:
> Ok, made a go at the 'master' table in best practices. If people want I can add best practice sections to each of the extensions with proposed roles. Don't love the wording on the 'specified in' and the ones listed there, but just wanted to get something in. Suggestions appreciated.

We are currently working on a multi-mission project with 50+ sensors of various type (optical, sar, atmo...) with long processing chain of value added product (L1 -> L3 processing). Those roles have become essential for the generic processing services to identify what they are dealing with.
For instance, we have a SAR chain that takes `amplitude` or `magnitude` (complex) data to derive ARD datasets with normalized `sigma0` and `gamma0`, then pass in a flood mapping synthetic service to produce `water-mask` datasets in different point in time. Finally, they are passing in a final service to produce a flood grading maps with `water-frequency` data.
The field must stay open for any data semantic but having a kind of baseline for those roles may help for sure the data (ex/inter)change. I don't mind so much about where are specified but having them classified by their thematic domain helps in finding the right one for your data.

BTW, could I contribute directly to the role list? If then, how can I do?

cholmes:
> BTW, could I contribute directly to the role list? If then, how can I do?

@emmanuelmathot that'd be great. It looks like you have an invite to be a commiter, but haven't accepted. Go to https://github.com/radiantearth/stac-spec/invitations and then you should be able to checkout the [roles best practice branch](https://github.com/radiantearth/stac-spec/tree/roles-best-practice) and you can contribute directly there. Just try to push often, and I'll do the same, as I've got a little more to do there.

best-practices.md
cholmes: Actually I don't think I'll add this yet, since right now it's not actually specified in the label extension. When #1001 is complete we can add it (or it can add it to this list if we merge this one first).

best-practices.md
cholmes: #1006 to track this - reverted any grouping from this PR.

cholmes:
Ok, I think this one is good to merge. It's not perfect by any means, but I think we should merge and then can continue to evolve it with dedicated PR's.

emmanuelmathot:
Sorry for the late contribution. I added roles corresponding to the conventions used in SAR and covering previous roles. May be reviewed in the future.

cholmes:
@emmanuelmathot - thanks! I didn't realize you'd be working so late, and checked to see if you had accepted the invite, so thought you were done for the day - would have waited for your to come in. 

I think your commit missed my table of contents update, and unfortunately right now the SAR table needs to be updated in two places (included it in the SAR extension directly, but wanted a good list in one place in best practices). If you are able to clean those up that'd be great, but I can also probably do it later today.

cholmes:
Oh great, I see you got in the SAR extension one as I was writing. thanks

emmanuelmathot:
I merged your table of content change. didn't I?

cholmes:
@emmanuelmathot - you did, but I took it out due to https://github.com/radiantearth/stac-spec/pull/989#discussion_r576351670  Hrmm... I thought I reverted it, due to #1006 But now I can't find my reverting commit, so maybe I just thought I did. I'll figure it out.

item-spec/item-spec.md
```diff
@@ -182,22 +182,33 @@ would be appropriate; if it is an XML, then `text/xml` is appropriate. For more
...
-| overview  | An asset that represents a possibly larger view than the thumbnail of the Item, for example, a true color composite of multi-band data. |
 | data      | The data itself. This is a suggestion for a common role for data files to be used in case data providers don't come up with their own names and semantics. |
 | metadata  | A metadata sidecar file describing the data in this item, for example the Landsat-8 MTL file. |
-| visual    | An asset that is a full resolution version of the data, processed for visual use (RGB only, often sharpened ([pan-sharpened](https://en.wikipedia.org/wiki/Pansharpened_image) and/or using an [unsharp mask](https://en.wikipedia.org/wiki/Unsharp_masking))). |
 
 It is STRONGLY RECOMMENDED to add to each STAC Item
 * a thumbnail with the role `thumbnail` for preview purposes
 * one or more data file although it doesn't need to use the suggested role `data`
 
+Note that it multiple roles per asset are encouraged, pick all the ones that apply. So many should have the 'data' role, and then
```
m-mohr, line 206: There seems to be a typo in here...

best-practices.md
```diff
@@ -203,6 +222,81 @@ register the media type with IANA, so that other STAC clients can find it easily
...
+
+The 'source' field lists where the role comes from. The ones the say Item Spec are the only 'official' roles that are fully
+standardized. In time others on this list may migrate to a more 'official' list. Those that say 'best practice' are just from this 
+doc, the listing is the table below. The ones from extensions are mostly just 'best practices' in the extensions, as there are few
+actual role requirements.
+
+| Role Name | Source | Description                                                            |
+| --------- | -------------|----------------------------------------------------------------------- |
+| thumbnail | [Item Spec](item-spec/item-spec.md#asset-role-types) | An asset that represents a thumbnail of the item, typically a true color image (for items with assets in the visible wavelengths), lower-resolution (typically smaller 600x600 pixels), and typically a JPEG or PNG (suitable for display in a web browser). Multiple assets may have this purpose, but it recommended that the `type` and `roles` be unique tuples. For example, Sentinel-2 L2A provides thumbnail images in both JPEG and JPEG2000 formats, and would be distinguished by their media types. |
```
m-mohr, line 251: Let's not repeat all roles from the extensions here. Better Add a list that links to additional roles specified in extensions.

best-practices.md
```diff
@@ -203,6 +222,81 @@ register the media type with IANA, so that other STAC clients can find it easily
...
+| beta0 | [SAR Extension](extensions/sar/README.md#best-practices) | Points to the radar backscatter file where data is referenced in the slant range plane and is radiometrically calibrated.  It is often derived from an `amplitude` or a `magnitude` role asset. |
+| gamma0 | [SAR Extension](extensions/sar/README.md#best-practices)  | Points to the radar backscatter file where data is referenced in the plane perpendicular to the line of sight. It is often derived from an `amplitude` or a `magnitude` role asset. |
+| date-offset | [SAR Extension](extensions/sar/README.md#best-practices) | Points to the date-offset file. |
+| covmat | [SAR Extension](extensions/sar/README.md#best-practices) | Points to the Points to the Normalized Polarimetric Radar Covariance Matrix (CovMat) file. |
+| prd | [SAR Extension](extensions/sar/README.md#best-practices) | Points to the Polarimetric Radar Decomposition (PRD) file. |
+
+Some of the particular asset roles also have some best practices.
+
+### Thumbnails
```
m-mohr, line 288: I thought we agreed on having a chapter that defines recommended sizes and formats for thumbnail, overview and visual? maybe don't say a lot in the table and add a link there to this chapter while extending this chapter with recommendation about thumbnails, overviews etc. in general?

matthewhanson:
I think @m-mohr and I were thinking the same thing, that the generic Roles table wouldn't include extension specific Roles, such as the SAR related ones.

But I don't have a really strong opinion here so I'm going to get this approved.

matthewhanson (approved):

best-practices.md
cholmes: I think it is useful to at least get a sense of what all the potential roles are in one list. But I can play with listing them quickly here but not having all the detail.

best-practices.md
cholmes: Yeah, I started this with the thumbnails, but then I didn't actually know what to recommend for the rest. I think this is the right spot for it.

cholmes: I'll add an issue, and then if someone fills in the actual recommendations on that issue I can do the PR to add them.

best-practices.md
m-mohr: Maybe just a list?

* [SAR](extensions/sar/) (noise-power, amplitude, magnitude, ... [list all])
* EO...

best-practices.md
m-mohr: I thought we had some criteria defined in the last call. Can't exactly recall. Can you, @matthewhanson?

best-practices.md
cholmes: I added them to the table, but just one line for each with a list.

cholmes:
Ok, I think this is good to go. Final approval appreciated.

m-mohr (approved):
Looks good, but still fearing that the lists of roles get out of sync.

cholmes:
I hear you, but I do think the value of being able to see all the potential roles in one place outweighs the downside if they get a little out of date.

m-mohr:
@cholmes It will likely not be all roles in the future as extension development will be in separate repo for many, releases will be separate, new extensions will occur and stac spec will release much less often so basically this will always be an incomplete list. People should actually not understand it as a complete list. Give it a month or two after the release and I think it will be outdated and incomplete.

m-mohr:
I just saw the last commit. That's a good change!

Issue #896, opened 2020-09-10
Best Practice: requestor pays


cholmes:
As suggested on gitter by @matthewhanson - it'd be good to have a best practice on URL's that are 'requestor pays'. We should capture these thoughts and put them in best practices.

'For requester pays URLs I’ve been using the s3 URL, e.g., s3://syncarto-data-rp/stac/naip/catalog.json
Since the http URL is useless on it’s own unless you sign it, so just working with the s3 URLs directly (with AWS CLI or boto3) I think is easier. Plus you can use PySTAC to support s3 reads/writes.
If public, then I use the actual http URL.
This might be a good thing to add to best practice'

'Even better might be to keep the STAC metadata in a different, and completely public, bucket that isn’t requester pays. Normally I like the data alongside the STAC Items, but I think it’s better if it’s public. That way you can use tools like STAC browser, PySTAC, without authentication for just the metadata.'

m-mohr:
Ist this just for S3? Would a person buying data at Planet also be "requester pays"? Or how exactly is that defined outside of S3?

davidraleigh:
This is a field I have on the stac version of assets:
https://geo-grpc.github.io/api/#epl.protobuf.v1.Asset

It's also used in google cloud:
https://cloud.google.com/storage/docs/requester-pays

And I image it also exists in Azure

jflasher:
I think it'd definitely be good to have requester pays called out in the metadata as it presents a technical and financial difference in how you access the data. I have tried to create the signatures myself for use with straight HTTP requests, but always fall back on the available SDKs. Also, at least for AWS, there are two costs incurred with requester pays, the egress and a per-request fee. The per-request fee is generally very small compared to egress cost, but this is not always the case (specifically in cases of listing the bucket contents) and likely should be mentioned for completeness.

philvarner:
S3 and Google have RP, Azure apparently does not.

I like some of the fields in @davidraleigh 's  link -- a few comments on them:

* cloud_platform (Google cloud, Azure, etc) -- I think this is good to have.  I use the "s3://" scheme in most of my asset hrefs, but I could just as well have used the virtual-host-style https uris. I assume this can be done with Google storage instead of "gs://", and Azure blob storage only uses https uris (e.g., https://myaccount.blob.core.windows.net/mycontainer/myblob)
[...]

Issue #956, opened 2021-01-27
Archive (ZIP, 7z, rar, ...) extension


m-mohr:
Taken from https://github.com/radiantearth/stac-spec/issues/921#issuecomment-746398428:

What I thought would be useful to add something like an "archive" extension, which allows to specify individual files from archives in assets, e.g. linking to a xml and tiff in a zip so that you can use for example eo:bands to describe the tiff in the archive. Something like archive:path could describe the path of the file in the archive (e.g. ZIP) and allows direct mapping and extraction. May also be useful for Zarr or so?

@DanielJDufour agreed in a private message:

> I saw your comments here about a possible "archive" extension [..] and just wanted to say that I think it would be a great idea. I often see data producers who don't have any experience with web technologies releasing files as zip files and it could be a good opportunity for software devs to create STAC catalogs of the metadata without having to host the actual files.

brentfraser:
I'm very much in favour of an archive extension.  We're currently experimenting with using STAC to automate data management (downloading) between server and client applications.  We need the client to check if files have already been downloaded and extracted, so it needs to know the contents of the archive.  We added the `contents` and `target_folder` objects to aid in this:

```
"assets": {
        "analytic": {
            "href": "https://my.server.com/data/arctic/2020_37_258.zip",
            "roles": ["data"],
            "type": "application/zip",
            "contents": [
                {
                    "name": "2020_37_258.cnt"
                }
            ],
            "target_folder": "./data/artic/"
        },
```

m-mohr:
@brentfraser What exactly is target_folder referring to? I don't understand the purpose...

Just for future reference, here's an example of how I imagined it to be, e.g. for a ZIP file containing an image (data/image.tiff) and a metadata file (meta/iso.xml):

```
{
  ...
  "assets": {
    "analytic": {
      "href": "https://my.server.com/data/category/example.zip",
      "roles": [
        "data",
        "archive"
      ],
      "type": "application/zip",
      "archive:href": "data/image.tif",
      "archive:type": "image/tiff"
    },
    "metadata": {
      "href": "https://my.server.com/data/category/example.zip",
      "roles": [
        "metadata",
        "archive"
      ],
      "type": "application/zip",
      "archive:href": "meta/iso.xml",
      "archive:type": "application/xml"
    }
  }
}
```

brentfraser:
[...]

Issue #891, opened 2020-09-02
Order offline assets


gilltk:
This issue is about ordering offline assets. Such assets could be rasters that reside a tape archive and must be 'activated' (to borrow Planet terminology) before they're available for download.
 
I am after some feedback on the following scheme for users to be able to order offline data. The vision our team has is shown at the video on this page: https://medium.com/@mojodna/a-stac-browser-348a60674061. If we get this right, then:
- a user will use a STAC browser to navigate a catalogue until they get to a StacItem
- the browser renders the Assets for the StacItem as hyperlinks
- the user can select an 'order' link to request the asset be activated
- in response, the browser renders the StacItem again, but with some updated properties on the status of the order and a link that the user can select to check the progress
- the user-browser interaction loops until the browser display's the StacItem's  'download' link, which the user selects to access the image.
 
There is more detail about our scheme below for the interested reader. If you read it, then I'd be very grateful:
- for people's thoughts on this scheme; example: is there a simpler approach? are we using the notion of an asset correctly?
- to know if we need to propose a STAC extension (like an 'order' extension) to implement this scheme? Or does the existing STAC 1.0.0-beta support it
 
Now to the detail. In our scheme, a StacItem has an 'order' asset, the URL to which is a request to our server to 'activate' the asset. The server responds with a 202 (processing) and a StacItem in the body. The StacItem's properties has an order:status field with the status of the order ('offline', 'processing', or 'online'). The returned StacItem's assets has a URL so the status can be queried again.

The following example workflow may make the scheme clearer. Below is a snippet of a StacItem returned from a search result. Note:
- The 'order' asset; it contains a link to our order endpoint so the file can be ordered
- The 'order:status' attribute in the properties; it is currently offline
 
```
{
  "id": "cemsre_t54jvn_20161118_abam4.img",
  "assets": {
    "metadata": {
      "href": "https://data-access.jrsrp.org.au/productmetadata"
    }
    "order": {
      "href": "https://stac.jrsrp.org.au/order?filenames=cemsre_t54jvn_20161108_abam4.img&state=NSW"
    }
  },
  ...

   "properties": {
    "collection": "nsw-Sentinel2-reflectance_10m",
     ...
     "order:status": "offline"
  },
}
```

A request is placed to order the file using the order-asset link. The response is a 202 (processing) and the StacItem is returned. Note:
- the order:status attribute in the returned StacItem has changed to 'processing'

```
[...]

PR #991, opened 2021-02-16
requester pays and other storage details in Asset and Item

davidraleigh:
**Related Issue(s):** #
https://github.com/radiantearth/stac-spec/issues/896

**Proposed Changes:**

1. adding in storage details for asset and item 

Overview of suggestions:
## Asset fields

| Field Name  | Type   | Description |
| ----------- | ------ | ----------- |
| storage:platform              | string    | (REQUIRED) The [cloud provider](#providers) where data is stored |
| storage:manager               | string    | The entity in charge of managing the data. |
| storage:region              | string | (REQUIRED) The region where the data is stored. Relevant to speed of access and inter region egress costs (as defined by PaaS provider) |
| storage:bucket                | string    | The bucket for the asset, used along with object path |
| storage:object_path           | string    | The object_path for the asset, used along with bucket |
| storage:requester_pays        | bool      | (REQUIRED) Is the data requester pays or is it data manager/cloud provider pays |
| storage:tier                  | string    | The title for the tier type (as defined by PaaS provider) |
| storage:tier_duration         | integer   | Minimum storage duration (in days) required before additional fees |
| storage:date_stored           | string    | Date and time the corresponding asset placed into the current storage tier (relevant for tier_duration > 0). Format is RFC 3339. |
| storage:first_byte_latency    | string    | approximate time unit (milliseconds, minutes or hours) for accessing first byte of data |

## Item fields

| Field Name  | Type   | Description |
| ----------- | ------ | ----------- |
| storage:min_tier_duration   | integer  | number of days for the shortest time tier restriction on access of an asset. |
| storage:max_tier_duration   | integer  | number of days for the longest time tier restrictions on access of an asset. |
| storage:archived            | bool    | descriptor for whether the data is "properly" archived according to implementers discretion |

**PR Checklist:**


davidraleigh:
Points of discussion:

- Separation of `bucket` and `object_path`
  - [IBM](https://cloud.ibm.com/docs/cloud-object-storage/api-reference?topic=cloud-object-storage-compatibility-api-bucket-operations#compatibility-api-storage-class), [GCP](https://cloud.google.com/storage/docs/creating-buckets), [Alibaba](https://www.alibabacloud.com/help/doc-detail/32135.htm), [AWS](https://docs.aws.amazon.com/AmazonS3/latest/dev-retired/Introduction.html), [Azure](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blobs-introduction) and [Oracle](https://docs.oracle.com/en-us/iaas/Content/Object/Tasks/managingbuckets.htm) all have the concept of a bucket (Azure changes the name to a container).
  - All have a concept of an object or blob that is placed within the bucket
  - As all PaaS maintain a separation of these concepts and many API calls treat them as separate concepts it make sense to keep them separate in our definition
- `regions`
  - useful for understanding egress costs within a PaaS (lower if your worker is in the same region)
  - also can help for understanding latency restrictions your workers will experience
- `requester_pays`
  - while this concept may only exists in GCP and AWS it has cost some users a lot of money by mistakenly using "Open" data that costs the user when they download it
- `tier`
  - names from the [Cloud Provider Storage Tiers](https://github.com/davidraleigh/stac-spec/tree/issue-896/extensions/storage#cloud-provider-storage-tiers) table
- `tier_duration`
  - allows the user to combine with the `date_stored` field to know if requesting data will incur an additional PaaS cost (this is related to the Order Offline data issue: https://github.com/radiantearth/stac-spec/issues/891)
- `first_byte_latency`
   - archive tier data often has some extended latency when data is moved back into Standard tier (or hot tier in Azure)
- Item's `min_tier_duration` and `max_tier_duration`
   - I imagine the scenario of occasionally wanting to push data from Standard storage tier into an archived storage tier state. I'd want to query for STAC items to see what is currently not archived. Instead of querying every STAC item and searching their Assets, I'd want to have the status on the STAC Item itself. Using the `min_tier_duration` and `max_tier_duration` fields we can inspect items that may require archiving.
   -  `archive_small` if we want to archive the large data, but leave the thumbnail's available to users for search, we could set `archive_small` to false and then the thumbs would be expected to be in Standard storage while the `min_tier_duration` 
 - `archived`
   - There may be a case where `min_tier_duration` and `max_tier_duration` do not effectively differentiate archived and non-archived data. so it might be nice to have a flag where the implementer can assign the status of a STAC Item's storage as being properly archived.

davidraleigh:
@cholmes, @matthewhanson  and @m-mohr before I do more work on this I'd love to get your insight on the above.

m-mohr:
I have not a lot of experience with cloud storage so I let others decide on the content, but overall I like how you have approached this.

The markdown needs some improvements and also a JSON Schema and example should be added, but we can do that after the content/fields have been agreed on.

cholmes:
@davidraleigh - overall same perspective as @m-mohr: it's a great start and structured well, and I personally don't have enough experience with cloud buckets to have real opinions.

One small note, you can run `npm run check-markdown` locally to see the checks that the test_docs check runs. It's just minor markdown stuff to allow us to be consistent.

@jflasher & @philvarner - would be great if you could review.

extensions/storage/README.md
```diff
@@ -0,0 +1,78 @@
...
+This document explains the fields of the Object Storage Extension to a STAC Item and an Asset. This does not cover NFS solutions provided by PaaS cloud companies. Allows user to specify details related to cloud storage access and costs for assets and their STAC item.
+
+- [Example (Landsat 8)](examples/example-landsat8.json)
+- [JSON Schema](json-schema/schema.json)
+
+## Asset fields
+
+| Field Name  | Type   | Description |
+| ----------- | ------ | ----------- |
+| storage:cloud_platform        | string    | The [cloud provider](#providers) where data is stored |
```
kylebarron, line 19: I agree the acceptable values here should be a well-defined enum. E.g. `aws` vs `amazon_web_services`

extensions/storage/README.md
```diff
@@ -0,0 +1,78 @@
...
+| storage:date_stored           | string    | Date and time the corresponding asset placed into the current storage tier (relevant for tier_duration > 0) |
+| storage:first_byte_latency    | string    | time unit (milliseconds, minutes or hours) for accessing first byte of data |
+
+## Item fields
+
+| Field Name  | Type   | Description |
+| ----------- | ------ | ----------- |
+| storage:min_tier_duration   | string  | Date and timestamp type describing the shortest time tier restrictions on access of assets. If `storage:archived_small` is false then thumbnails and other small overviews (implementer's discretion) are not included in `min_tier_duration` calculation |
+| storage:max_tier_duration   | string  | Date and timestamp type describing the longest time tier restrictions on access of assets. |
+| storage:archived_small         | bool      | Should the thumbnails or other small data be assumed to be included in the `min_tier_duration` and `max_tier_duration` |
```
kylebarron, line 36: "small data" seems quite vague. If the `storage` extension is applied to each asset, this shouldn't be necessary, because you can just see if there are `storage` keys within the thumbnail asset?

kylebarron (commented):
It'll be great to see an example Item. `href` is required on an asset, right? So does `storage:bucket` and `storage:object_path` need to come in addition to the href? Consider Landsat Collection 2 data on AWS (ignore the `landsatlook` service), then what would be the right way to define assets? Since the data is in a requester pays bucket, there's no simple HTTP way to access the data. My first thought is to use `s3://usgs-landsat/...` on the href, then also include the other `storage` fields?

Also, it might be worth considering putting `storage:bucket` at the Item level instead of the Asset level, since that's unlikely to change (though it could be overridden at the Asset level)

extensions/storage/README.md
```diff
@@ -0,0 +1,78 @@
+# Object Storage Extension Specification
+
+- **Title: Object Storage**
+- **Identifier: storage**
+- **Field Name Prefix: -**
+- **Scope: Item, Assets**
+- **Extension [Maturity Classification](../README.md#extension-maturity): Proposal**
+- **Owner**: @davidraleigh
+
+This document explains the fields of the Object Storage Extension to a STAC Item and an Asset. This does not cover NFS solutions provided by PaaS cloud companies. Allows user to specify details related to cloud storage access and costs for assets and their STAC item.
```
philvarner, line 10:
```suggestion
This document describes the Object Storage Extension.  This extension adds fields to STAC Item and Asset objects. This allows for details related to cloud storage access and costs for assets to be associated with a STAC Item.

This extension does not cover NFS solutions provided by PaaS cloud companies. 
```

extensions/storage/README.md
```diff
@@ -0,0 +1,78 @@
...
+| storage:first_byte_latency    | string    | time unit (milliseconds, minutes or hours) for accessing first byte of data |
+
+## Item fields
+
+| Field Name  | Type   | Description |
+| ----------- | ------ | ----------- |
+| storage:min_tier_duration   | string  | Date and timestamp type describing the shortest time tier restrictions on access of assets. If `storage:archived_small` is false then thumbnails and other small overviews (implementer's discretion) are not included in `min_tier_duration` calculation |
+| storage:max_tier_duration   | string  | Date and timestamp type describing the longest time tier restrictions on access of assets. |
+| storage:archived_small         | bool      | Should the thumbnails or other small data be assumed to be included in the `min_tier_duration` and `max_tier_duration` |
+| storage:archived              | bool      | descriptor for whether the data is "properly" archived according to implementers discretion |
```
philvarner, line 37: I don't understand what this is supposed to be from the description

extensions/storage/README.md
```diff
@@ -0,0 +1,78 @@
...
+| storage:manager               | string    | The entity in charge of managing the data. |
+| storage:regions               | [string]  | The region(s) where the data is stored. Relevant to speed of access and inter region egress costs (as defined by PaaS provider) |
+| storage:bucket                | string    | The bucket for the asset, used along with object path |
+| storage:object_path           | string    | The object_path for the asset, used along with bucket |
+| storage:requester_pays        | bool      | Is the data requester pays or is it data manager/cloud provider pays |
+| storage:tier                  | string    | The title for the tier type (as defined by PaaS provider) |
+| storage:tier_duration         | integer   | Minimum storage duration required before additional fees |
+| storage:date_stored           | string    | Date and time the corresponding asset placed into the current storage tier (relevant for tier_duration > 0) |
+| storage:first_byte_latency    | string    | time unit (milliseconds, minutes or hours) for accessing first byte of data |
+
```
philvarner, line 29:
```suggestion
**storage:cloud_platform** value representing the name of the object store. Valid values for common cloud storage providers are `aws_s3`, `alibaba_oss`, `microsoft_azure_storage`, `google_cs`, `ibm_os`, and `oracle_os`
```

extensions/storage/README.md
philvarner, line 19:
```suggestion
| storage:platform        | string    | The [cloud provider](#providers) where data is stored |
```

maybe just platform?

extensions/storage/README.md
m-mohr: How can we make that in the future we don't need to change the spec if a new cloud service appears? The ones here sound a bit US centric.

m-mohr: Maybe better have a table below with common services and their identifier, but make that list non-exclusive?

extensions/storage/README.md
```diff
@@ -0,0 +1,78 @@
...
+| Field Name  | Type   | Description |
+| ----------- | ------ | ----------- |
+| storage:cloud_platform        | string    | The [cloud provider](#providers) where data is stored |
+| storage:manager               | string    | The entity in charge of managing the data. |
+| storage:regions               | [string]  | The region(s) where the data is stored. Relevant to speed of access and inter region egress costs (as defined by PaaS provider) |
+| storage:bucket                | string    | The bucket for the asset, used along with object path |
+| storage:object_path           | string    | The object_path for the asset, used along with bucket |
+| storage:requester_pays        | bool      | Is the data requester pays or is it data manager/cloud provider pays |
+| storage:tier                  | string    | The title for the tier type (as defined by PaaS provider) |
+| storage:tier_duration         | integer   | Minimum storage duration required before additional fees |
```
philvarner, line 26: tier_duration seems like we're putting the cost structure schedule of the cloud provider into the metadata of the stac item. I'm not sure what the usefulness of this is.

extensions/storage/README.md
```diff
@@ -0,0 +1,78 @@
...
+## Asset fields
+
+| Field Name  | Type   | Description |
+| ----------- | ------ | ----------- |
+| storage:cloud_platform        | string    | The [cloud provider](#providers) where data is stored |
+| storage:manager               | string    | The entity in charge of managing the data. |
+| storage:regions               | [string]  | The region(s) where the data is stored. Relevant to speed of access and inter region egress costs (as defined by PaaS provider) |
+| storage:bucket                | string    | The bucket for the asset, used along with object path |
+| storage:object_path           | string    | The object_path for the asset, used along with bucket |
+| storage:requester_pays        | bool      | Is the data requester pays or is it data manager/cloud provider pays |
```
philvarner, line 24: requester_pays as a term is AWS specific, but I think it's also become the standard way of describing this

extensions/storage/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Asset fields
+
+| Field Name  | Type   | Description |
+| ----------- | ------ | ----------- |
+| storage:cloud_platform        | string    | The [cloud provider](#providers) where data is stored |
+| storage:manager               | string    | The entity in charge of managing the data. |
+| storage:regions               | [string]  | The region(s) where the data is stored. Relevant to speed of access and inter region egress costs (as defined by PaaS provider) |
+| storage:bucket                | string    | The bucket for the asset, used along with object path |
+| storage:object_path           | string    | The object_path for the asset, used along with bucket |
```
m-mohr, line 23: Maybe just `path`?

extensions/storage/README.md
```diff
@@ -0,0 +1,78 @@
...
+| storage:requester_pays        | bool      | Is the data requester pays or is it data manager/cloud provider pays |
+| storage:tier                  | string    | The title for the tier type (as defined by PaaS provider) |
+| storage:tier_duration         | integer   | Minimum storage duration required before additional fees |
+| storage:date_stored           | string    | Date and time the corresponding asset placed into the current storage tier (relevant for tier_duration > 0) |
+| storage:first_byte_latency    | string    | time unit (milliseconds, minutes or hours) for accessing first byte of data |
```
philvarner, line 28:
```suggestion
| storage:first_byte_latency    | string    | magnitude for time for accessing first byte of data (e.g., `milliseconds`, `seconds`,  `minutes`, or `hours`) |
```

extensions/storage/README.md
```diff
@@ -0,0 +1,78 @@
...
+| ----------- | ------ | ----------- |
+| storage:cloud_platform        | string    | The [cloud provider](#providers) where data is stored |
+| storage:manager               | string    | The entity in charge of managing the data. |
+| storage:regions               | [string]  | The region(s) where the data is stored. Relevant to speed of access and inter region egress costs (as defined by PaaS provider) |
+| storage:bucket                | string    | The bucket for the asset, used along with object path |
+| storage:object_path           | string    | The object_path for the asset, used along with bucket |
+| storage:requester_pays        | bool      | Is the data requester pays or is it data manager/cloud provider pays |
+| storage:tier                  | string    | The title for the tier type (as defined by PaaS provider) |
+| storage:tier_duration         | integer   | Minimum storage duration required before additional fees |
+| storage:date_stored           | string    | Date and time the corresponding asset placed into the current storage tier (relevant for tier_duration > 0) |
```
philvarner, line 27:
```suggestion
| storage:date_stored           | string    | Date and time the corresponding asset placed into the current storage tier (relevant for tier_duration > 0). Format is RFC 3339. |
```

extensions/storage/README.md
m-mohr: Could we make that a definitive list of options? (i.e. not "e.g." but "one of")

- milliseconds
- seconds
- minutes
- hours
- days
- weeks
- months
?

extensions/storage/README.md
```diff
@@ -0,0 +1,78 @@
+# Object Storage Extension Specification
+
+- **Title: Object Storage**
+- **Identifier: storage**
+- **Field Name Prefix: -**
```
philvarner, line 5:
```suggestion
- **Field Name Prefix: storage**
```

extensions/storage/README.md
philvarner: i think hours is probably the highest we need to go.

extensions/storage/README.md
philvarner: Yes, agree -- I phrased that wrong in text. I meant that for these, providers, use these values -- but there can and will be more

philvarner:
Here's an example I created from an item in my catalog 

```
{
  "assets": {
    "B5": {
      "eo:bands": [{ "common_name": "nir", "name": "B5" }],
      "gsd": 30.0,
      "href": "https://landsat-pds.s3.us-west-2.amazonaws.com/c1/L8/060/247/LC08_L1TP_060247_20180905_20180912_01_T1/LC08_L1TP_060247_20180905_20180912_01_T1_B5.TIF",
      "roles": ["data"],
      "size": 60084563,
      "title": "Band 5 - Near Infrared (NIR) - 851-879nm (30m)",
      "type": "image/tiff; application=geotiff; profile=cloud-optimized",
      "updated": "2018-09-14T05:05:04.000Z",
      "storage:cloud_platform": "aws_s3",
      "storage:manager": "???",
      "storage:bucket": "landsat-pds",
      "storage:object_path": "c1/L8/060/247/LC08_L1TP_060247_20180905_20180912_01_T1/LC08_L1TP_060247_20180905_20180912_01_T1_B5.TIF",
      "storage:requester_pays": false,
      "storage:tier": "standard",
      "storage:tier_duration": "???",
      "storage:date_stored": "2021-02-17T21:11:21Z",
      "storage:first_byte_latency": "milliseconds"
    }
  },
  "bbox": [
    -102.33259279506724,
    80.62959444469965,
    -86.8612845861914,
    82.67608173251769
  ],
  "collection": "landsat8_l1tp",
  "geometry": {
    "coordinates": [
      [
        [
          [-100, 80],
          [-102, 80],
          [-102, 82],
          [-100, 82],
          [-100, 80]
        ]
      ]
    ],
    "type": "MultiPolygon"
  },
  "id": "LC08_L1TP_060247_20180905_20180912_01_T1_L1TP",
  "links": [],
  "properties": {
    "datetime": "2018-09-05T21:49:29.9922900Z",
    "storage:min_tier_duration": "???",
    "storage:max_tier_duration": "???",
    "storage:archived_small": "???",
    "storage:archived": "???"
  },
  "stac_extensions": ["eo", "projection"],
  "type": "Feature"
}
```

extensions/storage/README.md
m-mohr: We have some companies that store on tape, which can take days to get the data from. Not sure this is covered by this extension though. And also not sure "first_byte_latency"  is the best term then

extensions/storage/README.md
```diff
@@ -0,0 +1,78 @@
...
+- [JSON Schema](json-schema/schema.json)
+
+## Asset fields
+
+| Field Name  | Type   | Description |
+| ----------- | ------ | ----------- |
+| storage:cloud_platform        | string    | The [cloud provider](#providers) where data is stored |
+| storage:manager               | string    | The entity in charge of managing the data. |
+| storage:regions               | [string]  | The region(s) where the data is stored. Relevant to speed of access and inter region egress costs (as defined by PaaS provider) |
+| storage:bucket                | string    | The bucket for the asset, used along with object path |
```
lossyrob, line 22: Azure doesn't have a single concept of a 'bucket', but has a 'storage account' and 'container'.

It seems like the `storage:bucket` and `storage:object_path` are redundant information for the asset `href`. Is the idea the `href` would be an http(s) url? I think if we were to drop the bucket and object_path properties, clients would be able to parse information from either the http endpoint or something like `s3://...`.

philvarner:
> Also, it might be worth considering putting `storage:bucket` at the Item level instead of the Asset level, since that's unlikely to change (though it could be overridden at the Asset level)

I feel these need to be on the asset level -- I think we made a mistake early on in stac that a lot of new fields that should have been on assets (such as gsd) where defined in Properties and therefore applied to the entire Item. Something like this is definitely specific to each asset, and the Item Properties fields are simply summarizing what's in the Assets (similar to the meaning of gsd in Properties now)

philvarner:
> It'll be great to see an example Item. `href` is required on an asset, right? So does `storage:bucket` and `storage:object_path` need to come in addition to the href? Consider Landsat Collection 2 data on AWS (ignore the `landsatlook` service), then what would be the right way to define assets? Since the data is in a requester pays bucket, there's no simple HTTP way to access the data. My first thought is to use `s3://usgs-landsat/...` on the href, then also include the other `storage` fields?

`href` is required for asset, correct. I year ago, I think I would have said that (in AWS terms) bucket and object_path would need to be separate fields.  Now I feel like it's redundant.  The href value is going to be one of the forms:

1. s3://{bucket}/{object_path}
2. https://{bucket}.s3.{region}.amazonaws.com/{object_path} or https://{bucket}.s3.amazonaws.com/{object_path}
3. https://s3-{region}.amazonaws.com/{bucket}/{object_path} or https://s3.amazonaws.com/{bucket}/{object_path}

But, now that I've written out those 5 options, maybe i'd rather just have region + bucket + object_path and construct my own URL to do whatever I need. So... I guess I'm in favor of splitting them out, but with a more vendor-agnostic name than bucket.

Also, someone brought up in the work session today that `region` would be a good field to add -- most services have some notion like this, and from a practical perspective, if I have an `s3://` url, I want to be able to connect directly to that region.

lossyrob:
> But, now that I've written out those 5 options, maybe i'd rather just have region + bucket + object_path and construct my own URL to do whatever I need. So... I guess I'm in favor of splitting them out, but with a more vendor-agnostic name than bucket.

Perhaps there can be a vendor-specific object to specify the location information? e.g. instead of 

```
      "storage:manager": "AWS",
      "storage:bucket": "landsat-pds",
      "storage:object_path": "c1/L8/060/247/LC08_L1TP_060247_20180905_20180912_01_T1/LC08_L1TP_060247_20180905_20180912_01_T1_B5.TIF",
```

it would be

```
      "storage:manager": "AWS",
      "storage:location": {
          "region": "us-east-1",
          "bucket": "landsat-pds",
          "object": "c1/L8/060/247/LC08_L1TP_060247_20180905_20180912_01_T1/LC08_L1TP_060247_20180905_20180912_01_T1_B5.TIF",
      }
```

or for Azure

```
      "storage:manager": "Azure",
      "storage:location": {
          "region": "west europe",
          "storageAccount": "naipeuwest",
          "container": "naip",
          "blob": "v002/md/2017/md_100cm_2017/39077/m_3907755_ne_18_1_20170730.tif",
      }
```

This maybe gets to be too much of an extension-on-an-extension, but we could provide a JSON schema that accounts for common storage providers' storage:location, as well as one with a more generic path to use in other cases.

davidraleigh:
updating with samples and explanations. please hold.

extensions/storage/README.md
philvarner: > Maybe just path?

I think `storage:container` and `storage:path` are good names. With an explicit example, e.g., 

```
{ 
 "href": "https://foobar-data.s3.us-west-2.amazonaws.com/some/path/to/a/cog.tif",
  "storage:platform": "aws_s3",
  "storage:container": "foobar-data",
  "storage:path": "some/path/to/a/cog.tif",
  "storage:region":"us-west-2"
}
```

extensions/storage/README.md
matthewhanson: agree here, how do I know what is to be considered small?

extensions/storage/README.md
matthewhanson: Yes, shorter is better and cloud seems superfluous

extensions/storage/README.md
matthewhanson: Does it need to be `aws_s3` or is `s3` sufficient, where we can rely on established protocol names already used in URLs? 
e.g., s3://, gs://   

`s3` for AWS and `gs` for Google seem to be widely used, there just doesn't seem to be an accepted upon standard for Azure.

extensions/storage/README.md
matthewhanson: Yeah, since Google uses this as well, and it's also a very descriptive term I think it's good to use generally.  

If Azure implements a similar mechanism my guess is they will adopt the same term

extensions/storage/README.md
matthewhanson: I'm not convinced of this field is needed at all.  Doesn't `tier` imply a certain latency for a given provider?  
As a provider how would you populate this field? Would you basically look up the latency for the tier you are storing your data and use that?  If so then I don't think it is needed.

If is is needed, than `latency` would seem to be sufficient.

extensions/storage/README.md
davidraleigh: we can remove `archived_small` and just put the responsibility on the maintainer of the STAC service to not include thumbnails that are in Standard storage in the calculation of the min_tier_duration

davidraleigh: or to include them if they want all items archived

extensions/storage/README.md
matthewhanson: I thought we might have had a note on using the `created` and `updated` fields under Assets, but we don't:
https://github.com/radiantearth/stac-spec/blob/dev/best-practices.md#common-use-cases-of-additional-fields-for-assets

But you could use `created` under Assets which would imply when the asset was first created, and `updated` would be when it was last updated.  I think that covers this particular use, @m-mohr thoughts?

extensions/storage/README.md
davidraleigh: @philvarner I'll put more details in the readme. This is about being able to query all your STAC items to see which items need to be transitioned to an Archived state.

extensions/storage/README.md
davidraleigh: I was avoiding platform because C++ and enums and platform clashing with stac item platform. But I can make it `platform` for this pull request and make it `storage_platform` in protobuf.

extensions/storage/README.md
matthewhanson: I'm not a big fan of the redundant information by having `bucket` and `path` since it is already in href which encapsulates not only bucket, path, but also protocol.

extensions/storage/README.md
davidraleigh: all cloud providers have a concept of a duration that a storage object is required to remain at that storage level before it can be retrieved without cost. It's not as precise as cost structure information, but it's similar to `requester_pays` in importance.

matthewhanson:
> > Also, it might be worth considering putting `storage:bucket` at the Item level instead of the Asset level, since that's unlikely to change (though it could be overridden at the Asset level)
> 
> I feel these need to be on the asset level -- I think we made a mistake early on in stac that a lot of new fields that should have been on assets (such as gsd) where defined in Properties and therefore applied to the entire Item. Something like this is definitely specific to each asset, and the Item Properties fields are simply summarizing what's in the Assets (similar to the meaning of gsd in Properties now)

I get why you would want to put this at the Item level to remove redundancy, but if `href` is per Item and we actually keep `bucket` and `path` then seems like they should be at the same level.

Also, I can't think of a good use case where a user would want to search on bucket, so there's no need to have it at the Item level.

extensions/storage/README.md
davidraleigh: > If Azure implements a similar mechanism my guess is they will adopt the same term
they were the only one of the six cloud providers I investigated to buck the trend and use `container` instead of `bucket`...

matthewhanson:
> > It'll be great to see an example Item. `href` is required on an asset, right? So does `storage:bucket` and `storage:object_path` need to come in addition to the href? Consider Landsat Collection 2 data on AWS (ignore the `landsatlook` service), then what would be the right way to define assets? Since the data is in a requester pays bucket, there's no simple HTTP way to access the data. My first thought is to use `s3://usgs-landsat/...` on the href, then also include the other `storage` fields?
> 
> `href` is required for asset, correct. I year ago, I think I would have said that (in AWS terms) bucket and object_path would need to be separate fields. Now I feel like it's redundant. The href value is going to be one of the forms:
> 
> 1. s3://{bucket}/{object_path}
> 2. https://{bucket}.s3.{region}.amazonaws.com/{object_path} or https://{bucket}.s3.amazonaws.com/{object_path}
> 3. [https://s3-{region}.amazonaws.com/{bucket}/{object_path}](https://s3-%7Bregion%7D.amazonaws.com/%7Bbucket%7D/%7Bobject_path%7D) or [https://s3.amazonaws.com/{bucket}/{object_path}](https://s3.amazonaws.com/%7Bbucket%7D/%7Bobject_path%7D)
> 
> But, now that I've written out those 5 options, maybe i'd rather just have region + bucket + object_path and construct my own URL to do whatever I need. So... I guess I'm in favor of splitting them out, but with a more vendor-agnostic name than bucket.
> 
> Also, someone brought up in the work session today that `region` would be a good field to add -- most services have some notion like this, and from a practical perspective, if I have an `s3://` url, I want to be able to connect directly to that region.


+1 to add region.  All cloud providers use data centers, and those data centers are somewhere, so the `region` concept is universal.

The problem with constructing your own URL....how does a user know how to do that?  It's specific to each provider.   

I'm in the camp of only `href` is necessary and that can be parsed into components for a given provider by client software if that is needed.

extensions/storage/README.md
davidraleigh: > I think `storage:container` and `storage:path` are good names. With an explicit example, e.g.,
It would be strange to adopt Azure's `container` when all the other providers use the term `bucket`

`key` is a more common term used by cloud providers (in addition to `object` and `blob`). `path` isn't really anywhere in documentation. I just added it to `object` to make that more clear.

matthewhanson:
STAC Items are generally going to be parsed and used by a client library of some sort to access the data, so I'm not sure we can't just leave the logic of parsing the `href` into whatever is needed can't be left to these client libraries.

This is a common function that we've all done, I have a [simple library for AWS ](https://github.com/matthewhanson/boto3-utils/blob/master/boto3utils/s3.py#L32)that I use in a lot of my other code, and abstraction libraries such as `rasterio` and `fsspec` utilize URLs and don't require those to be broken down.  

I don't see how adding several new fields that duplicate existing information is better in a specification than in an abstraction library.

extensions/storage/README.md
m-mohr: We have, but it's in this chapter: https://github.com/radiantearth/stac-spec/blob/dev/item-spec/common-metadata.md#date-and-time

m-mohr: And there's also the timestamps extension, which covers this probably better with "published"

m-mohr:
> Also, I can't think of a good use case where a user would want to search on bucket, so there's no need to have it at the Item level.

I guess it could be useful to have region in the item properties as I could imagine people searching for that. Also, requester pays seems to be useful for search. Maybe also others...

extensions/storage/README.md
davidraleigh: @matthewhanson tier isn't directly related. For Alibaba's "Cold Archive", there are different options for your speed to first byte depending on the fees you're willing to pay, 1 hour, 2 hours, etc. 

it's not exactly latency. it really is a delay of some period of time and then your latency to download is the normal access speed. Oracle and Microsoft describe it as "Time to first byte" and AWS uses "First Byte Latency".

extensions/storage/README.md
davidraleigh: We use STAC a lot internally, so object_path and bucket are useful to those internal users who have access permissions to use them, but for customers there is an href that isn't constructed from bucket + object_path.

https://github.com/davidraleigh/stac-spec/blob/issue-896/extensions/storage/examples/example-nsl.json#L66
https://github.com/davidraleigh/stac-spec/blob/issue-896/extensions/storage/examples/example-nsl.json#L69

davidraleigh:
Please ignore the schema for now. I still haven't figure out how to make a schema.json that is for both an Item *and* and Asset. If there's an example of one I can copy that would be great.

I can imagine a lot of my suggestions can be optional (or maybe even just internal for Near Space Labs). But a lot of the motivations is to cover a few different situations.

- One is that our `href` is not directly related to our bucket and object key.
- requester pays is valuable information to prevent running up a terrible bill
- archiving
  - we want to be able to archive older data. so we want to search by the `min_tier_duration` to see  what data hasn't been archived
  - we want to know what data need to be "reconstituted" and will require a delay for a customer.
  - if it's a lot of archived data that we're pulling out of archive prematurely, we want to be able to get billing insights


I've included some samples. 
https://github.com/davidraleigh/stac-spec/blob/issue-896/extensions/storage/examples/example-nsl.json
https://github.com/davidraleigh/stac-spec/blob/issue-896/extensions/storage/examples/example-nsl.json

extensions/storage/README.md
davidraleigh: https://github.com/radiantearth/stac-spec/pull/991/commits/09510ce3e91b73010a93991defdd51856c47c3b1

extensions/storage/README.md
davidraleigh: https://github.com/radiantearth/stac-spec/pull/991/commits/09510ce3e91b73010a93991defdd51856c47c3b1

extensions/storage/README.md
matthewhanson: key is more descriptive as well, I think it's a better term than path

extensions/storage/README.md
matthewhanson: I see, that makes sense. I don't have a strong opinion here, but agree with @m-mohr that this should be an enumerated list of options.

matthewhanson:
Thanks @davidraleigh , those are some helpful use cases.  It reminded me that I wanted to open an issue on alternate locations, which thinking about now is directly related to this issue, see #997

davidraleigh:
@matthewhanson, I think an `alternate_hrefs` array would be an additionally useful field, and honestly I think it should be standard on the Asset Object.

So these are the outstanding points:
1.  replace `object_path` with `key` or `path` or leave as is
2. replace `first_byte_latency` with `latency` or leave as is
3. agree on a set of enums for `platform`. in [shema.json](https://github.com/davidraleigh/stac-spec/blob/341242dfb43b80bc40fe1d500fea925ca8f3d87b/extensions/storage/json-schema/schema.json#L43-L51) currently: 
```json
              "enum": [
                "UNKNOWN_CLOUD_PLATFORM",
                "AWS",
                "GCP",
                "AZURE",
                "IBM",
                "ALIBABA",
                "ORACLE"
              ],
```
4. agree on set of enums (or some other format of description) for `first_byte_latency`
5. define the required fields for extension users (currently `platform`, `region` and `requester_pays`
6. define fields that should be dropped altogether

extensions/storage/README.md
m-mohr: I think I'd also like to use the protocol names, although several providers implement s3. Do they all use the same protocol then?

extensions/storage/README.md
m-mohr: I'm fine either way as long as we avoid the `_` 😅. For people not so much into cloud storage path seems more intuitive, but if key is an accepted term across providers, then I guess it's fine.

m-mohr:
> @matthewhanson, I think an `alternate_hrefs` array would be an additionally useful field, and honestly I think it should be standard on the Asset Object.

I'm wondering whether this extension is meant to be just "cloud storage" platforms related or anything storage related? Like would alternative hrefs and potentially also things like the archive extension be put in here? Could we also use it to describe other types of storage? E.g. "latency" for some kind of "we need a day for you to get the data from tape". Then storage is fine, but I think if this is just for cloud storage, we should rename the extension and use another prefix maybe.

> 2. replace `first_byte_latency` with `latency` or leave as is

I feel like latency (if just for cloud storage, see paragraph above) is good with an enum that gives milliseconds, seconds, minutes, hours, days, weeks or so.

> 3. agree on a set of enums for `platform`.

I think above we agreed more or less that it shouldn't be an enum, but just a table that maps the common cloud providers to a specific key. I'm wondering whether a good "best-practice" for the field could either be the domain of the service in general (e.g. `cloud.google.com`, `aws.amazon.com`, `azure.microsoft.com`) or the protocol (e.g. s3:// gs://) - Then you also have a recommended way for new providers to be added without us changing the spec too often.

> 2. define the required fields for extension users (currently `platform`, `region` and `requester_pays`

I think region could be an issue if a platform doesn't implement this concept. Then implementors would need to come up with arbitrary names...

> 3. define fields that should be dropped altogether

I'm thinking maybe manager? Could we put this into the provider array in collections, potentially with just a new role?

extensions/storage/README.md
philvarner: I think implementing s3 just means you conform to the AWS S3 REST API that clients (such as the cli) use anyway. 

I'd prefer separating the scheme from the actual service being used, and to follow the general pattern of `{provider}_{service_name}` -- otherwise, we'd have to assume `aws` meant `s3` (what if they later introduce `s4`?), whereas the other more generic names are just like `cloud_storage` have to have the provider. So, consistency.

Also, the Asset#href might be an ARN (`arn:aws:s3:::bucket_name/key_name`) and then you're back to parsing the URI to find the service.

extensions/storage/README.md
philvarner: I get requester_pays, since that has to be set in the request or you can't retrieve the item. I don't know if that's ever used to make a decision about whether it should be retrieved or not (e.g., don't retrieve anything you have to pay for).  How would tier_duration be used? I just want to know what the use case is.

extensions/storage/README.md
philvarner: oh, interesting -- I only use AWS, so I just assumed that other services didn't use bucket out of spite -- I'm fine with it if that's the common term.

extensions/storage/README.md
philvarner: bucket and key sound fine then, if all the other providers use it. I was trying to counter my AWS-bias and overcorrected.

philvarner:
> @matthewhanson, I think an `alternate_hrefs` array would be an additionally useful field, and honestly I think it should be standard on the Asset Object.

I might even go further and just say it's `alternate_assets`, and if you only want the href, then putting an Asset that has only the href in it is valid. Assets all the way down...

davidraleigh:
@m-mohr 

> I'm wondering whether this extension is meant to be just "cloud storage" platforms related or anything storage related? 

Maybe this should just be separated into `cloud_storage` and `archive`. I can see how some of the archiving features are nice for on prem data storage. Maybe something like this:

## Asset fields

| Field Name  | Type   | Description |
| ----------- | ------ | ----------- |
| cloud_storage:platform              | string    | (REQUIRED) The [cloud provider](#providers) where data is stored |
| cloud_storage:manager               | string    | The entity in charge of managing the data. |
| cloud_storage:region                | string    | (REQUIRED) The region where the data is stored. Relevant to speed of access and inter region egress costs (as defined by PaaS provider) |
| cloud_storage:bucket                | string    | The bucket for the asset, used along with object path |
| cloud_storage:key           | string    | The key describes the virtual object path for the asset, used along with bucket |
| cloud_storage:requester_pays        | bool      | (REQUIRED) Is the data requester pays or is it data manager/cloud provider pays |

## Asset fields

| Field Name  | Type   | Description |
| ----------- | ------ | ----------- |
| archive:title                 | string    | The title for the tier type (as defined by PaaS provider) |
| archive:min_duration          | integer   | Minimum storage duration (in days) required before additional fees. If you're using a cloud provider this is something like 0, 30, 60, 90, 180, 365 days |
| archive:date_stored           | string    | Date and time the corresponding asset placed into the current storage tier (relevant for tier_duration > 0). Format is RFC 3339. |
| archive:latency               | string    | approximate time unit (milliseconds, minutes or hours) for accessing the first byte of data |

## Item fields

| Field Name  | Type   | Description |
| ----------- | ------ | ----------- |
| archive:min_duration   | integer  | number of days for the shortest time tier restriction on access of an asset. |
| archive:max_duration   | integer  | number of days for the longest time tier restrictions on access of an asset. |
| archive:archived       | bool    | descriptor for whether the data is "properly" archived according to implementers discretion |

> Like would alternative hrefs and potentially also things like the archive extension be put in here?

I actually don't think alternative hrefs should be a part of the `cloud_storage` definition. All the 6 cloud providers I've researched for this pull request have separated concepts of `bucket` and `key`. There are many  API calls interacting specifically with `bucket`, getting back a bucket object which you then interact on with the `key`. This leads me to believe that putting them as one string in an alternate hrefs for users to parse themselves is a bit messy.

> What about a flag that says you need to authenticate to access the assets?

It would be useful on `cloud_storage`, where you need authentication for GCP landsat, but not AWS. But what if the href has one authentication method and the `cloud_storage` requires a different? It's almost like there should be one for the Asset Object and an additional one for the `cloud_storage`.

> I'm wondering whether a good "best-practice" for the field could either be the domain of the service in general (e.g. cloud.google.com, aws.amazon.com, azure.microsoft.com) or the protocol (e.g. s3:// gs://) - Then you also have a recommended way for new providers to be added without us changing the spec too often.

I prefer enums whenever possible. But I can see your point. I can update that.

>I think region could be an issue if a platform doesn't implement this concept. Then implementers would need to come up with arbitrary names...

with the above sample of separating into `cloud_storage` and `archive` prevents this situation.

> I'm thinking maybe manager? Could we put this into the provider array in collections, potentially with just a new role?
Likely also date_stored as this can probably be expressed with one of the existing timestamps (published).

If you look at my example for [NAIP Item](https://github.com/davidraleigh/stac-spec/blob/341242dfb43b80bc40fe1d500fea925ca8f3d87b/extensions/storage/examples/example-naip.json#L61-L100), we have 3 different managers for the Assets in Azure, AWS and GCP:
```json
    "CO_GEOTIFF_GCP_RGB": {
      "href": "https://storage.googleapis.com/naip-data/tx/2016/100cm/rgb/30097/m_3009743_sw_14_1_20160928.tif",
      "type": "image/vnd.stac.geotiff; cloud-optimized=true",
      "storage:bucket": "naip-data",
      "storage:object_path": "tx/2016/100cm/rgb/30097/m_3009743_sw_14_1_20160928.tif",
      "storage:platform": "GCP",
      "storage:manager": "Near Space Labs",
      "storage:region": "us-central1",
      "storage:requester_pays": true,
      "storage:tier": "COLDLINE",
      "storage:tier_duration": 90,
      "storage:date_stored": "2021-02-02T00:00:00+00:00",
      "storage:first_byte_latency": "milliseconds"
    },
    "CO_GEOTIFF_AWS_RGB": {
      "href": "https://naip-visualization.s3.amazonaws.com/tx/2016/100cm/rgb/30097/m_3009743_sw_14_1_20160928.tif",
      "type": "image/vnd.stac.geotiff; cloud-optimized=true",
      "storage:bucket": "naip-visualization",
      "storage:object_path": "tx/2016/100cm/rgb/30097/m_3009743_sw_14_1_20160928.tif",
      "storage:platform": "AWS",
      "storage:manager": "ESRI",
      "storage:region": "us-west-2",
      "storage:requester_pays": true,
      "storage:tier": "standard",
      "storage:tier_duration": 0,
      "storage:first_byte_latency": "milliseconds"
    },
    "CO_GEOTIFF_AZURE_RGB": {
      "href": "https://naipeuwest.blob.core.windows.net/naip/v002/tx/2016/tx_100cm_2016/30097/m_3009743_sw_14_1_20160928.tif",
      "type": "image/vnd.stac.geotiff; cloud-optimized=true",
      "storage:bucket": "naipeuwest",
      "storage:object_path": "naip/v002/tx/2016/tx_100cm_2016/30097/m_3009743_sw_14_1_20160928.tif",
      "storage:platform": "AZURE",
      "storage:manager": "AZURE",
      "storage:region": "westeurope",
      "storage:requester_pays": false,
      "storage:tier": "hot tier",
      "storage:tier_duration": 0,
      "storage:first_byte_latency": "milliseconds"
    },
```

jflasher:
Hi all, jumping in a bit late so apologies if this was already mentioned, but while I think the storage tier pieces are very important, tracking things like `archive:date_stored` might be hard to track in practice? I believe there will be systems where the data is put into cold storage and brought up for short period of times. In that case, would the STAC entry need to be updated when it was brought to warm storage and then updated again when it was put back down into cold storage? Having STAC know what is currently in what type of storage does seem useful, but potentially adding a lot of overhead?

davidraleigh:
> Having STAC know what is currently in what type of storage does seem useful, but potentially adding a lot of overhead?

yeah, I think users that implement `archive` are the type that are concerned about the costs of their data, and the overhead of tracking this `date_stored`--updating it when it is moved between tiers--will be worth it for the savings in cloud expenses. I guess that user who are always using standard storage tier assets will not ever bother with all these complexities.

m-mohr:
FYI: `archive:` would conflict with #956

cholmes:
Note we just established a new github org at https://github.com/stac-extensions which will be the home for all but the most mature extensions after #946 So this PR will ultimately land there. If this group is willing to live on the edge we could move this PR into its own repo. Our goal is to make templates for publishing schemas, using CI, etc, so this would need to update to that when it comes. But it should enable more infrastructure for discussion, as could make PR's and open different issues. Let me know, and we can create a repo and give the people on this thread permissions on it.

matthewhanson:
PR Closing as this is being added as an extension in stac-extensions:

https://github.com/stac-extensions/stac-extensions.github.io/issues/4

PR #994, opened 2021-02-17
Projection doc fix

davidraleigh:
**Related Issue(s):** #
None

**Proposed Changes:**

1. Update broken link for `proj:projjson`
2.  Update to CONTRIBUTING.md to allow for using Docker instead of installing npm locally

**PR Checklist:**


davidraleigh:
@m-mohr small docs fix for projection

m-mohr:
@davidraleigh The link is not broken for me and I think it's better to link to the HTML description page rather than linking to the JSON Schema.

m-mohr (changes requested):
@davidraleigh The link is not broken for me and I think it's better to link to the HTML description page rather than linking to the JSON Schema.

I can't run Docker, so I can't test, but it looks like it would only run on *nix systemns with that command. We should add a note or provide it platform independant.

davidraleigh:
@m-mohr it looks like dev used to have the old https://proj.org/usage/projjson.html link when I made this change. Now that it's https://proj.org/specifications/projjson.html, it's totally fine

Issue #979, opened 2021-02-08
Clarify how to get 'full width half max' from a range


cholmes:
As discussed on the call today, @cholmes and @lossyrob both want to fill out the full_width_half_max value, but don't know what a 'top hat' distribution is or how you go from a range to the fwhm. Would be good to discuss what this is, how it works, and what to do if your spec sheet only has a 'range', or a range and a center value.

@matthewhanson - if you could write this up in the ticket I can make a PR. Or you can just make the PR.

PR #996, opened 2021-02-18
add additional text for calculating center_wavelengt and full_width_h…

matthewhanson:
**Related Issue(s):** #
#979 

**Proposed Changes:**

1. Add text to go from min/max wavelengths to center_wavelength and full_width_half_max

**PR Checklist:**


cholmes:
This is a good start, but still not enough for me as a non-remote-sensing expert to know what to do. Am I allowed to just assume it's a 'top-hat' filter? Or am I supposed to go ask some expert to figure out if it is actually a top-hat? I'm pretty sure Planet's is not a top hat... (I did get the fwhm for at least skysat, so this isn't pressing for me, just want it to be easier for whoever comes next) And it would probably be good to have an example conversion too.

matthewhanson:
You're right it's not really clear.  In reality, there's no such thing as a top-filter, it's only a theoretical concept and an approximation in the absence of a full spectral transmission profile.  But if all you have is min and max, you can convert that to center and fwhm, but that is also an approximation. 

These numbers are helpful to determine the approximate location (color) of a band, but if you actually needed it for analysis/pre-processing (especially required in harmonization between sensors), then you'd want the complete profile.  As an aside, the spectral profile can shift over time as well. They are usually characterized before launch.

I'll update to make it simpler, provide an example, but also want to make it clear these are approximate characterizations.

cholmes:
Sounds great - that's exactly what I'd want to see. 

That reminds me, the other note that we should add is to encourage people to use the 'common names' even if they don't have center wavelength and full width half max. That came up with NAIP, and I hadn't actually known to do that. I see the value, but the eo extension doesn't come right out and say it (though they aren't required).

kylebarron:
> encourage people to use the 'common names' even if they don't have center wavelength and full width half max

Big +1. It makes it a lot easier to interpret the data if you can look at `common_name` instead of parsing the wavelength, and `common_name` without wavelength/width vastly improves ease of use of the image.

lossyrob (approved):

cholmes (approved):

Issue #1013, opened 2021-02-24
Use collection assets in new examples


cholmes:
We moved collection assets to core with #1008 but the new examples refresh in #955 hadn't landed when that PR was made, so we just removed the old examples. But we should make sure that the new examples show at least one instance of a collection asset.

cholmes:
Easiest way to do this is just add a thumbnail in one or more of the collections

```
"assets": {
    "thumbnail": {
      "href": "thumbnail.png",
      "title": "Preview",
      "type": "image/png",
      "roles": [
        "thumbnail"
      ]
    }
  },
```

PR #1023, opened 2021-03-02
Collection assets example

cholmes:
**Related Issue(s):** #1013 

**Proposed Changes:**

1. Added a metadata collection asset to the collection-only example
2. added iso 19115 and 19139 to the 'roles' list in best practices.

**PR Checklist:**

### DIFF
--- a/examples/collection-only/collection.json
+++ b/examples/collection-only/collection.json
@@ -44,13 +44,14 @@
     }
   },
   "assets": {
-    "thumbnail": {
-      "href": "thumbnail.png",
-      "title": "Preview",
-      "type": "image/png",
+    "metadata_iso_19139": {
       "roles": [
-        "thumbnail"
-      ]
+        "metadata",
+        "iso-19139"
+      ],
+      "href": "https://storage.googleapis.com/open-cogs/stac-examples/sentinel-2-iso-19139.xml",
+      "title": "ISO 19139 metadata",
+      "type": "vnd.iso.19139+xml"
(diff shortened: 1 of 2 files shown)


lossyrob (approved):

examples/collection-only/collection.json
```diff
@@ -44,13 +44,14 @@
...
+    "metadata_iso_19139": {
       "roles": [
-        "thumbnail"
-      ]
+        "metadata",
+        "iso-19139"
+      ],
+      "href": "https://storage.googleapis.com/open-cogs/stac-examples/sentinel-2-iso-19139.xml",
+      "title": "ISO 19139 metadata",
+      "type": "vnd.iso.19139+xml"
```
m-mohr, line 54: That doesn't seem like a valid media type.

cholmes: Yeah... I got it from spacebel, and found reference in [geoserver](https://geoserver.geo-solutions.it/downloads/releases/2.8.x-ld/doc/extensions/inspire/using.html) and [mapproxy](https://mapproxy.org/docs/latest/inspire.html) so seemed like it may be some official INSPIRE thing? 

I'm also fine to just use xml type...

m-mohr: Some googling brought up that it should probably be:

```suggestion
      "type": "application/vnd.iso.19139+xml"
```

m-mohr (approved):

PR #1047, opened 2021-03-11
Don't require Collection Assets

m-mohr:
**Related Issue(s):** None

**Proposed Changes:**

- Collection Assets were specified as required (only in written text, not in JSON Schema), but that was incorrectly copied over from the former `collection-assets` extension. Collection Assets are not required.

(If this was meant to be required, then the JSON Schema needs a fix.)

**PR Checklist:**


cholmes (approved):
D'oh. My bad on this one, thanks for catching. rc.2 here we come! :)

Issue #1030, opened 2021-03-03
Update UML diagram


cholmes:
@hgs-msmith - could you update the UML diagram? I'm going to remove it for RC1, but I think we could put it back into 1.0.0. Or if you could do it asap for RC1 that'd be awesome.

cholmes:
dupe of #831

PR #1051, opened 2021-03-15
Restored UML diagram

hgs-msmith:
Updated to spec version 1.0.0

**Related Issue(s):** #1030 

**Proposed Changes:**

1. Updated to spec version 1.0.0
2. 

**PR Checklist:**


matthewhanson (approved):

cholmes:
@hgs-msmith - one of the changes we made in 1.0.0-rc.1 is that collection no longer inherits from catalog. Could you update the UML so it doesn't have that relationship?

philvarner:
After the mtg this morning, it dawned on me what the problem is and what the relation is:

* Catalog and Collection have a has-a relationship -- a Catalog has-a zero-or-more Collections.  
* Collection is **NOT** a Catalog (is-a)
* Collection and Catalog both is-a (are-a?) "STAC Aggregate" -- they both relate together other object (a Collection relates together other Collections and Items; a Catalog relates together other Catalogs, Collections, and Items). Aggregate is probably a bad term to use here b/c it's too easy to get confused with SQL-like aggregation functions **instead** of having the domain-driven design meaning. 
* I think it's actually simpler just to say they're two types not related by subtying, but happen to have several fields that are named the same.

hgs-msmith:
Updated UML diagram to remove inheritance relationship between Collection and Catalog.

m-mohr (changes requested):
Thanks Michael, here are some suggestions for improvements:
- Items can also link to collections with rel types root and parent.
- The note about the extensions can be removed or shortened ("Note: This diagram does not contain extensions.") as we are going to remove all extensions from core.
- link:rel=child is 0..* in catalogs and 0..1 in collections. I think it should be 0..* in both.

m-mohr (approved):

PR #1054, opened 2021-03-19
New common band name mwir38

emmanuelmathot:
**Proposed Changes:**

1. new Common band `mwir38`. used for instance for gaofen-4 mission

**PR Checklist:**


extensions/eo/README.md
```diff
@@ -119,6 +119,7 @@ numbers of several popular instruments.
 | cirrus      | 1.35 - 1.40     |             | 9         | 10         | 26    |      |
 | swir16      | 1.55 - 1.75     | 5           | 6         | 11         | 6     |      |
 | swir22      | 2.10 - 2.30     | 7           | 7         | 12         | 7     |      |
+| mwir38      | 3,50 - 4.10     |             |           |            |       |      |
```
philvarner, line 122:
```suggestion
| mwir38      | 3.50 - 4.10     |             |           |            |       |      |
```

cholmes (approved):

matthewhanson:
I thought I had written a response to this, must have never hit Comment.

The criteria for adding new common names is that the approximate location and size appears in multiple instruments. This is clearly the case for all the standard colors, Near IR and the 2 main SWIR atmospheric windows.    Narrow bands tend to be too specific and unique.

MWIR is from 3-5um, and gaofen-4 MWIR is 3.5 - 4.1 um. 

I've got a couple questions:
- Are there other sensors that have an equivalent MWIR band?
- What are some other bandpasses in use for MWIR sensors?

emmanuelmathot:
To my knowledge, in spaceborn remote sensing, theere is MODIS instrument on aqua and terra sat: https://webapps.itc.utwente.nl/sensor/getsen.aspx?name=MODIS with 5 mwir band:

matthewhanson:
It's not a question if MWIR is commonly used, but rather how common is a bandpass that is centered on 3.8um and do they tend to have a similar bandwidth.  None of the MODIS bands are similar to the gaofen band.

What we want to avoid is giving these sorts of mission-specific bands a common name using the center wavelength if they aren't actually common.

emmanuelmathot:
I understand your point. As long as this field remains open to any value, this is fine with me.
Be aware (but you probably are ;-) that this nomenclature is very useful and adopted very quickly (that's our case). I don't know where to set the cursor on the completeness of the existing sensor wavelengths but, for now, I agree with your criteria for adding new common names.
You can close this PR for now if you prefer.

philvarner:
@emmanuelmathot It won't be open to any name going forward, pending resolution of this issue https://github.com/radiantearth/stac-spec/issues/1062

The reason is that these should be well-defined and have a very high bar (either used by one of the big 3 of Landsat, MODIS, or Sentinel-2) or be used by 3 or more other satellites. 

I understand your use case exactly because I've had the same need. You're always able to set another proprietary field in the band description to whatever you want.

emmanuelmathot:
ho that's bad news :-(. Indeed, we will have to use another field but that is the kind of ugly workaround I'd like to avoid because they become permanent and creates incompatibilities in the future.

cholmes:
Feel free to give feedback on #1062 - it made sense on our call, but I could see an argument of allowing more, and just giving a warning if it's not one of the defined ones. Certainly I want the spec to support your use case.

matthewhanson:
@emmanuelmathot Can you not just use the "name" field the way you'd like?

emmanuelmathot:
yes no worry. We will do that.

Issue #1045, opened 2021-03-09
Extend summary to support list of objects / Better describe summary values


jjrom:
In the Collection "summaries" property, a summary for a field can be specified in two ways - either 1) an array of "all distinct values in an array", or 2) "a Stats Object".

With this definition, the summary for the platform field with two platforms (sentinel-2a and sentinel-2b) would be:
"platform": [
     "sentinel-2a",
     "sentinel-2b"
]

In resto, we have additional statistical information for each fields (including platform). In particular, the number of items for each platforms can be provided to the user. It sounds natural to provide this information through the "summaries" property. As a consequence in resto, the previous example is written this way:
 "platform": [
            {
                "id": "sentinel-2a",
                "name": "Sentinel - 2A",
                "parentId": "root",
                "count": 211199
            },
            {
                "id": "sentinel-2b",
                "name": "Sentinel - 2B",
                "parentId": "root",
                "count": 208632
            }
]

Does the "all distinct values in an array" mean that the "value" is a string or a number or does it also mean that the "value" can be any object ? In the latter case, the resto approach is valid against the specification. Otherwise, could we update the specification to allow this approach ?

m-mohr:
The values in the summary array can be objects, but the summary values need to be compliant with the original types as defined by STAC. So your platform example is invalid, it can only be an array of strings. But a field that defines an object could be summarized with an array of objects, of course.

That Stats Object for summaries allows objects, but in this case it's a single object per field and it must contain minimum and maximum values. So that's not so helpful for your use case.

Now that 1.0.0 RC1 has been released, we also can't really change the summaries anymore. So you need to put your additional details somewhere else to be STAC-compliant. That could be a new extension or maybe for APIs it would be covered by the queryables endpoint from OGC APIs?

jjrom:
Ok but in this example (https://github.com/radiantearth/stac-spec/blob/master/examples/collection-only/collection.json), the summaries property contains a "eo:bands" entry that does not fulfill the Stats Object

m-mohr:
eo:bands is contained in an array there. I don't understand the issue?

jjrom:
Sorry I misunderstood the first explanation, i.e. that the properties in the summaries "must be compliant with the original types as defined by STAC". This is not clear from the specification (https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md#summaries). Perhaps this should be explicitly added in the summary chapter ?
[...]

PR #1093, opened 2021-04-20
JSON Schema in summaries #1045

m-mohr:
**Related Issue(s):** #1045

**Proposed Changes:**

1. Support JSON Schema in summaries - I hadn't had a lot of time to flesh out the language, improvements appreciated.
2. Rename Stats Object to Range Object
3. Fix collection examples (add missing stac_extensions entries), but that fails right now due to https://github.com/stac-extensions/scientific/pull/2

**PR Checklist:**

- [x] This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec), and I have NOT opened an issue/PR yet!

### DIFF
--- a/collection-spec/collection-spec.md
+++ b/collection-spec/collection-spec.md
@@ -16,7 +16,8 @@
   - [Link Object](#link-object)
     - [Relation types](#relation-types)
   - [Asset Object](#asset-object)
-  - [Stats Object](#stats-object)
+  - [Range Object](#range-object)
+  - [JSON Schema Object](#json-schema-object)
 - [Media Type for STAC Collections](#media-type-for-stac-collections)
 - [Standalone Collections](#standalone-collections)
 
@@ -54,7 +55,7 @@ specified in [*OGC API - Features*](https://ogcapi.ogc.org/features/), but they 
 | license         | string                                           | **REQUIRED.** Collection's license(s), either a SPDX [License identifier](https://spdx.org/licenses/), `various` if multiple licenses apply or `proprietary` for all other cases. |
 | providers       | \[[Provider Object](#provider-object)]           | A list of providers, which may include all organizations capturing or processing the data or the hosting provider. Providers should be listed in chronological order with the most recent provider being the last element of the list. |
 | extent          | [Extent Object](#extent-object)                  | **REQUIRED.** Spatial and temporal extents.                  |
-| summaries       | Map<string, \[*]\|[Stats Object](#stats-object)> | STRONGLY RECOMMENDED. A map of property summaries, either a set of values or statistics such as a range. |
+| summaries       | Map<string, \[\*]\|[Range Object](#range-object)\|[JSON Schema Object](#json-schema-object)> | STRONGLY RECOMMENDED. A map of property summaries, either a set of values, a range of values or a [JSON Schema](https://json-schema.org). |
 | links           | \[[Link Object](#link-object)]                   | **REQUIRED.** A list of references to other documents.       |
 | assets          | Map<string, [Asset Object](#asset-object)>       | Dictionary of asset objects that can be downloaded, each with a unique key. |
 
@@ -95,12 +96,12 @@ Summaries help to fully define Collections, especially if they don't link to any
 build tailored user interfaces for querying the data, by presenting the potential values that are available.
 Summaries should summarize all values in every Item underneath the collection, including in any nested sub-Catalogs.
 
-A summary for a field can be specified in two ways:
+A summary for a field can be specified in three ways:
 
 1. A set of all distinct values in an array: The set of values must contain at least one element and it is strongly recommended to list all values.
    If the field summarizes an array (e.g. [`instruments`](../item-spec/common-metadata.md#instrument)),
    the field's array elements of each Item must be merged to a single array with unique elements.
-2. Statistics in a [Stats Object](#stats-object): Statistics by default only specify the range (minimum and maximum values),
+2. A Range in a [Range Object](#range-object): Statistics by default only specify the range (minimum and maximum values),
    but can optionally be accompanied by additional statistical values.
    The range specified by the `minimum` and `maximum` properties can specify the potential range of values,
    but it is recommended to be as precise as possible.
@@ -281,9 +282,9 @@ or streamed. The definition provided here, at the Collection level, is the same 
 | type        | string    | [Media type](../item-spec/item-spec.md#asset-media-type) of the asset. See the [common media types](../best-practices.md#common-media-types-in-stac) in the best practice doc for commonly used asset types. |
 | roles       | \[string] | The [semantic roles](../item-spec/item-spec.md#asset-role-types) of the asset, similar to the use of `rel` in links. |
 
-### Stats Object
+### Range Object
 
-For a good understanding of the summarized field, statistics can be added.
+For summaries with a lot of continuous values, statistics can be added.
 By default, only ranges with a minimum and a maximum value can be specified.
 Ranges can be specified for [ordinal](https://en.wikipedia.org/wiki/Level_of_measurement#Ordinal_scale) values only,
 which means they need to have a rank order.
@@ -294,6 +295,14 @@ Implementors are free to add other derived statistical values to the object, for
 | ---------- | -------------- | ----------- |
 | minimum    | number\|string | **REQUIRED.** Minimum value. |
 | maximum    | number\|string | **REQUIRED.** Maximum value. |
+
+### JSON Schema Object
+
+For a full understanding of the summarized field, a JSON Schema can be added for each summarized field.
+This allows very fine-grained information for each field and each value as JSON Schema is also extensible.
+
+It is recommended to use [JSON Schema draft-07](https://json-schema.org/specification-links.html#draft-7)
+to align with the JSON Schemas provided by STAC.
 
 ## Media Type for STAC Collections
 
--- a/collection-spec/collection-spec.md
+++ b/collection-spec/collection-spec.md
@@ -284,7 +284,7 @@ or streamed. The definition provided here, at the Collection level, is the same 
 
 ### Range Object
 
-For summaries with a lot of continuous values, statistics can be added.
+For summaries that would normally consist of a lot of continuous values, statistics can be added instead.
 By default, only ranges with a minimum and a maximum value can be specified.
 Ranges can be specified for [ordinal](https://en.wikipedia.org/wiki/Level_of_measurement#Ordinal_scale) values only,
 which means they need to have a rank order.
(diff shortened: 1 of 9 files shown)


m-mohr:
@jjrom Could you please have a look at this PR? I had to derive a bit from our discussion (see the issue for some context). I'm happy to discuss this again in a short call, too.

jjrom (changes requested):
Thanks @m-mohr for this PR ! Looks good to use json-schema in summaries.

I have 2 remarks :
- I would suggest to add an example of a summaries property including an array of object with the const keyword (as you provided here https://github.com/radiantearth/stac-spec/issues/1045#issuecomment-823332018).
- In https://github.com/radiantearth/stac-spec/blob/6755d41c68f45cdcbf908b20c91c101fce685f2a/collection-spec/collection-spec.md#json-schema-object the sentence "For a full understanding of the summarized field, a JSON Schema can be added for each summarized field" is not very clear. We should be more explicit with a focus on a real use case (the count use case for instance ?)

m-mohr:
@jjrom Do you have an example at hand that fits into the examples folder? Could be an old one that we convert to the new schema.
We don't want too many/long examples in markdown to limit the length of the documents, so I'd put them somewhere in the examples folder and link to it, but the existing examples I also don't really want to touch, because I don't have a clue what values for counts etc to put in.

collection-spec/collection-spec.md
```diff
@@ -281,9 +282,9 @@ or streamed. The definition provided here, at the Collection level, is the same
 | type        | string    | [Media type](../item-spec/item-spec.md#asset-media-type) of the asset. See the [common media types](../best-practices.md#common-media-types-in-stac) in the best practice doc for commonly used asset types. |
 | roles       | \[string] | The [semantic roles](../item-spec/item-spec.md#asset-role-types) of the asset, similar to the use of `rel` in links. |
 
-### Stats Object
+### Range Object
 
-For a good understanding of the summarized field, statistics can be added.
+For summaries with a lot of continuous values, statistics can be added.
```
m-mohr, line 287:
```suggestion
For summaries that would normally consist of a lot of continuous values, statistics can be added instead.
```

jjrom:
Perhaps we could duplicate the "collection.json" example (https://github.com/radiantearth/stac-spec/blob/master/examples/collection.json) to a "collection-extended-summaries.json" example in which we replace 
```
"platform": [
      "cool_sat2",
      "cool_sat1"
]
```
with
```
"platform": {
      "type":"string",
      "oneOf":[
          {
              "const":"cool_sat1",
              "count": 103489
           },
          {
              "const":"cool_sat2",
              "count": 50700
          }
      ]
}
```
?

m-mohr:
@jjrom I think we can only add it to the collection-only folder right now. The others are a full catalog, so the count would likely be 1 or 2 as there are not so many items and we also can't add two collections, because then the items would need to link to two of them. 

Maybe we could duplicate, shorten and then convert https://github.com/radiantearth/stac-spec/blob/master/examples/collection-only/collection.json a bit and also add titles to epsg codes as you mentioned in the call?

jjrom:
@m-mohr Ok that makes sense. In this case we can duplicate  https://github.com/radiantearth/stac-spec/blob/master/examples/collection-only/collection.json and update the platform property. Concerning the EPSG codes, I'm fine to add titles but only if we shorten massively the property content - for instance limits it to two EPSG codes (e.g. EPSG:4326 and EPSG:3857). Otherwise it would be hard to read

m-mohr:
Yes, agreed! Or can you export your "best" example from your implementation, which we then try to convert to the new format? Would be a good verification that the new schema works well with your data.

jjrom:
I plan to update my implementation to fit the new schema next week. If it's not too late I can provide a real example by the end of next week. Otherwise we stick to the existing collection.json duplication/shortening as we discussed (?)

m-mohr:
@jjrom I thought about an example from your current implementation. That should be easy enough to migrate "manually".

cholmes:
+1 on 'manual migration'. I have some time today and can try to help.

m-mohr:
@schwehr This looks very much like it could be useful for you as you have this gee:schema thing in your catalogs, right? Any thought on the PR? Could this be useful for you, too?

schwehr:
Some initial thoughts while reading to catch up...  I'm not familiar enough with jsonschema.  And I haven't had a chance to talk it over with simonf.

1. Looking at the example, it would be nice to have simple (even if contrived) examples of all of them (as much as jsonschema can support). 
2. At least one example without a constrained set of values for string and int.
3. An example that used more of the capabilities of [json-schema](http://json-schema.org/draft-07/schema) would be good.
4. Is there a way to specify the units?
We expected to have types of INT, DOUBLE, STRING, INT_LIST, DOUBLE_LIST, or STRING_LIST.  I looks like we are not using the DOUBLE_LIST.  The proto enum: 
```
enum PropertyType {
  PROPERTY_TYPE_UNSPECIFIED = 0;  // Something is wrong.
  STRING = 1;
  INT = 2;
  DOUBLE = 3;
  STRING_LIST = 4;
  INT_LIST = 5;
  DOUBLE_LIST = 6;
}
```

m-mohr:
The CI fails due to https://github.com/stac-extensions/scientific/pull/2

m-mohr:
I'm wondering whether we should mark this new feature 'experimental' (with the chance of breaking in a minor release)?

schwehr:
Put in a suggestion for units: https://github.com/json-schema-org/json-schema-spec/issues/1094

cholmes:
> I'm wondering whether we should mark this new feature 'experimental' (with the chance of breaking in a minor release)?

Is there a way to make the feature an extension? Happy for it to not be namespaced and even mentioned in the main spec. But that could let us evolve it without having to cut stac-spec releases.

m-mohr:
No, it's not really extensible, I'd say. At least the extension would have the potential to break implementations and thus it's not a good extension.

m-mohr:
Thanks, @schwehr.

> 1. Is there a way to specify the units?

Not standardized, but custom keywords are a thing.

> 2. Looking at the example, it would be nice to have simple (even if contrived) examples of all of them (as much as jsonschema can support).

Not sure what exactly you are asking for?

> 3. At least one example without a constrained set of values for string and int.

We encourage to specify what values to expect to make it more useful for users. Just specifying that the field exists (with a specific data type?) would not be a best practice IMHO. Ideally there would be more information.

> 4. An example that used more of the capabilities of [json-schema](http://json-schema.org/draft-07/schema) would be good

I'm not sure STAC is the right place to go into JSON Schema details. Wouldn't it be better to just link to https://json-schema.org/learn/ ?

> 5. We expected to have types of INT, DOUBLE, STRING, INT_LIST, DOUBLE_LIST, or STRING_LIST.  I looks like we are not using the DOUBLE_LIST.

There's no direct type value for types arrays in JSON Schema, instead you do something like:
`{ type: 'array', items: { type: 'integer' } }` (INT_LIST) or `{ type: 'array', items: { type: 'string' } }` (STRING_LIST)
The other types would be `integer` (INT), `float` (DOUBLE) and `string` (STRING).

> Put in a suggestion for units: [json-schema-org/json-schema-spec#1094](https://github.com/json-schema-org/json-schema-spec/issues/1094)

JSON Schema allows for custom keywords, so I'm not sure there's a need for another annotation in a validation language as you can simply define it on your own. But let's see what they think.

examples/collection-only/collection.json
```diff
@@ -1,7 +1,12 @@
 {
   "type": "Collection",
   "stac_version": "1.0.0-rc.2",
-  "stac_extensions": [],
+  "stac_extensions": [
```
m-mohr, line 4: We'd need to make the following lines a separate PR if we want to merge this now. CI fails due to this change and the following issue: https://github.com/stac-extensions/scientific/pull/2

Still, these changes should be in RC3...

collection-spec/collection-spec.md
```diff
@@ -295,6 +296,15 @@ Implementors are free to add other derived statistical values to the object, for
...
 
+### JSON Schema Object
+
+For a full understanding of the summarized field, a JSON Schema can be added for each summarized field.
+This allows very fine-grained information for each field and each value as JSON Schema is also extensible.
+Each schema must be valid against all corresponding values available for the property in the sub-Items.
+
+It is recommended to use [JSON Schema draft-07](https://json-schema.org/specification-links.html#draft-7)
+to align with the JSON Schemas provided by STAC. Empty schemas are not allowed.
+
```
m-mohr, line 307:
```suggestion

For an introduction to JSON Schema, see "[Learn JSON Schema](https://json-schema.org/learn/)".

```

examples/collection-only/collection.json
cholmes: Ok, I think I didn't quite understand what you were saying on the call @m-mohr, but I think I've got what you wanted, though feel free to correct. I made a PR https://github.com/radiantearth/stac-spec/pull/1103 to do the scientific change, that we'll keep as an open PR until we have scientific 1.0.1. 

And then just need to remove the scientific stuff from here. I would just edit directly, but guess I'll do as a PR since I'm not 100% this is the path you wanted.

cholmes: And PR to do the removal is #1104, just against this branch.

cholmes (approved):

schwehr:
Is a custom type available within the actual schema being defined inside the STAC summary?  In the case I'm talking about,  the data isn't what's going to have the units.  The schema says what the units are going to be.

> > 1. Looking at the example, it would be nice to have simple (even if contrived) examples of all of them (as much as jsonschema can support).
> 
> Not sure what exactly you are asking for?

There is little sense of the flexibility of the possibilities here.  I agree that the reader should go look at the json schema spec, but the examples here are extremely narrow.

> 
> > 1. At least one example without a constrained set of values for string and int.
> 
> We encourage to specify what values to expect to make it more useful for users. Just specifying that the field exists (with a specific data type?) would not be a best practice IMHO. Ideally, there would be more information.

Examples of non-categorical data:

- string: A field with the name of the person who processed the data, or a field
- int: in the range of 0..1e6.  say population in a polygon
- float: A float between 0..1.  The fraction of cloudy pixels in an image

collection-spec/collection-spec.md
```diff
@@ -95,12 +96,12 @@ Summaries help to fully define Collections, especially if they don't link to any
...
+A summary for a field can be specified in three ways:
 
 1. A set of all distinct values in an array: The set of values must contain at least one element and it is strongly recommended to list all values.
    If the field summarizes an array (e.g. [`instruments`](../item-spec/common-metadata.md#instrument)),
    the field's array elements of each Item must be merged to a single array with unique elements.
-2. Statistics in a [Stats Object](#stats-object): Statistics by default only specify the range (minimum and maximum values),
+2. A Range in a [Range Object](#range-object): Statistics by default only specify the range (minimum and maximum values),
    but can optionally be accompanied by additional statistical values.
    The range specified by the `minimum` and `maximum` properties can specify the potential range of values,
    but it is recommended to be as precise as possible.
```
lossyrob, line 107: If I'm reading correctly, should have the 3rd way, describing the JSON Schema (and perhaps linking to the below paragraph on the JSON Schema summary)

lossyrob (approved):

m-mohr:
This seems good for now. We can fine-tune docs and examples later.

Issue #1098, opened 2021-04-23
Best practice when root == parent?


m-mohr:
When doing some examples, I was wondering what the best practice is for STAC entities for which the parent and the root catalog are the same (i.e. childs of the root). Should we have links with:
1. parent and root defined?
2. only root defined?

matthewhanson:
+1 on both

cholmes:
I'll do a PR to update the examples to always have both.

Not sure if we need to actually say this anywhere? I lean towards no, but happy to add something in.

m-mohr:
I would mention it somewhere, at least it was something I was not sure about. Could be BP though.

PR #1101, opened 2021-04-23
Best practice when root == parent clarification

cholmes:
**Related Issue(s):** #1098 

**Proposed Changes:**

1. Added line in best practices about having root and parent point at the same place
2. updated examples to follow this best practice.

**PR Checklist:**


best-practices.md
```diff
@@ -641,6 +641,10 @@ a number of the common official relations that are used in production STAC imple
 | next      | Indicates that the link's context is a part of a series, and that the next in the series is the link target. Typically used in STAC by API's, to return smaller groups of Items or Catalogs/Collections. |
 | preview   | Refers to a resource that serves as a preview (see [RFC 6903, sec. 3](https://tools.ietf.org/html/rfc6903#section-3)), usually a lower resolution thumbnail. In STAC this would usually be the same URL as the [thumbnail](#thumbnail) asset, but adding it as a link in addition enables OGC API clients that can't read assets to make use of it. It also adds support for thumbnails to STAC Catalogs as they can't list assets. |
 
+Being liberal with the `links` also means that it's ok to have repeated links with the same `rel`. For example the
```
m-mohr, line 644: Was that meant to say href instead of rel?

matthewhanson: In the example here it should be href, but good to point out that same `rel` types are also ok

matthewhanson (approved):

best-practices.md
cholmes: Yes, will fix.

m-mohr (approved):

PR #1103, opened 2021-04-26
Validate scientific extension, update for #1045 (merge after scientific extension 1.0.0 release)

cholmes:
**Related Issue(s):** #1093 https://github.com/stac-extensions/scientific/pull/2

**Proposed Changes:**

1. Bring in the updated scientific example, so we can release rc.3 without waiting on  https://github.com/stac-extensions/scientific/pull/2, as discussed in https://github.com/radiantearth/stac-spec/pull/1093/files#r620248117

**PR Checklist:**

- [x]  a CHANGELOG entry is not required.

### DIFF
--- a/examples/collection-only/collection.json
+++ b/examples/collection-only/collection.json
@@ -2,7 +2,7 @@
   "type": "Collection",
   "stac_version": "1.0.0-rc.2",
   "stac_extensions": [
-    "https://stac-extensions.github.io/scientific/v1.0.0/schema.json"
+    "https://stac-extensions.github.io/scientific/v1.0.1/schema.json"
   ],
   "id": "sentinel-2",
   "title": "Sentinel-2 MSI: MultiSpectral Instrument, Level-1C",
--- a/examples/collection-only/collection.json
+++ b/examples/collection-only/collection.json
@@ -1,7 +1,9 @@
 {
   "type": "Collection",
   "stac_version": "1.0.0-rc.2",
-  "stac_extensions": [],
+  "stac_extensions": [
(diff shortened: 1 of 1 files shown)


m-mohr:
I don't understand this PR? Do we remove the example for RC3 through #1104 and then add it back again after RC3? In this case the Schema URL needs to update from 1.0.0 to 1.0.1.

cholmes:
I thought that's what you were intending, but I wasn't sure, so happy to take another route.

My take was we should try to release rc.3 asap, and if it's an example of an extension that is blocking us then we should just take that out. But I think it's a good example, so whenever it's in good shape then we should merge it back in.

But if it can happen today/tomorrow I'm for just merging in the scientific fix and cutting 1.0.1 from there as the route to unblock this.

m-mohr:
So this PR is not meant to be merged for RC3, understood. Then we need to update the schema URI, of course.

examples/collection-only/collection.json
```diff
@@ -1,7 +1,9 @@
 {
   "type": "Collection",
   "stac_version": "1.0.0-rc.2",
-  "stac_extensions": [],
+  "stac_extensions": [
+    "https://stac-extensions.github.io/scientific/v1.0.0/schema.json"
```
m-mohr, line 5:
```suggestion
    "https://stac-extensions.github.io/scientific/v1.0.1/schema.json"
```

m-mohr:
Looking at it again, I think we could simply not merge it and have these examples in the corresponding extension repos. It's nothing major anyway.

PR #1104, opened 2021-04-26
remove scientific extension

cholmes:
**Related Issue(s):** #1045 #1093 

**Proposed Changes:**

1. Remove scientific so we can merge this, as discussed in https://github.com/radiantearth/stac-spec/pull/1093/files#r620248117 so we don't have to wait for https://github.com/stac-extensions/scientific/pull/2 to merge.

**PR Checklist:**


examples/collection-only/collection.json
```diff
@@ -83,11 +82,6 @@
       "minimum": 6.78,
       "maximum": 89.9
     },
-    "sci:citation": {
```
m-mohr, line 86: Bummer, that was an interesting use case where you actually define a pattern for values. We should try to come up with such a use case for another extension field.

PR #1111, opened 2021-04-29
new overview section

cholmes:
**Related Issue(s):** #

**Proposed Changes:**

1. Added a 'foundations' section to the overview, to talk about the various standards we rely on.
2. Clarified that 'null' is not the same as leaving a field out.

**PR Checklist:**

### DIFF
--- a/overview.md
+++ b/overview.md
@@ -21,8 +21,8 @@ JSON Schemas provide the core testing definitions, and are used in an array of v
 on [RFC 8288 (Web Linking)](https://tools.ietf.org/rfc/rfc8288.txt) to express relationships between resources,
 and IANA [Media Types](https://en.wikipedia.org/wiki/Media_type) to describe file formats and format contents.
 The [OGC API - Features](https://ogcapi.ogc.org/features/) standard is a final core building block. The STAC
-Collection extends their [Collection](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_collection_)
-JSON (and the full API is the foundation for the STAC API specification).
+Collection extends the [Collection](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_collection_)
+JSON defined in OGC API - Features (and the full API definition is the foundation for the STAC API specification).
 
 The STAC specifications are written to be understandable without needing a full background in these. But if you 
 want to get deep into STAC tool implementation and are not familiar with any of the standards mentioned above it is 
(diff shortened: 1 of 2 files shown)


lossyrob (approved):

overview.md
```diff
@@ -3,14 +3,37 @@
...
+
+STAC is built on top of many great standards and practices. Every part of STAC is 
+[JSON](https://www.json.org/json-en.html), and [GeoJSON](https://geojson.org/) provides the core geometry fields 
+and [features](https://en.wikipedia.org/wiki/Simple_Features) definition. All fields are described in the 
+specifications, and the acceptable values are defined with [JSON Schema](https://json-schema.org/). The released
+JSON Schemas provide the core testing definitions, and are used in an array of validation tools. We also rely
+on [RFC 8288 (Web Linking)](https://tools.ietf.org/rfc/rfc8288.txt) to express relationships between resources,
+and IANA [Media Types](https://en.wikipedia.org/wiki/Media_type) to describe file formats and format contents.
+The [OGC API - Features](https://ogcapi.ogc.org/features/) standard is a final core building block. The STAC
+Collection extends their [Feature Collection](http://docs.opengeospatial.org/is/17-069r3/17-069r3.html#_collection_)
```
m-mohr, line 24: That's a bit confusing as people may think about GeoJSON FeatureCollections?!

overview.md
```diff
@@ -3,14 +3,37 @@
...
+
+The STAC specifications are written to be understandable without needing a full background in these. But if you 
+want to get deep into STAC tool implementation and are not familiar with any of the standards mentioned above it is 
+recommended to read up on them. STAC development is guided by set of core philosophical tenets, like 
+building small reusable parts that are loosely coupled, focusing on developers, and more - see our the 
+[principles](principles.md) document to learn more.
+
+*Note: Setting a field in JSON to `null` is not equivalent to a field not appearing. STAC defines `null` explicitly 
+for some fields, where it has a particular meaning. It should not be used otherwise - instead the field should be left 
+out entirely.*
```
m-mohr, line 35: > It should not be used otherwise - instead the field should be left 
out entirely.

I had to read and think about it like 2 minutes to figure out what was meant.

What is "it"? It seems to refer to "null", but then the "otherwise" is still confusing. Would try to rephrase...

overview.md
cholmes: Ah, good point. I'll just call it collection.

overview.md
cholmes: Cool, will try again. I was struggling on how to explain it well.

overview.md
philvarner: I'd also clarify what "their" is referring to.

cholmes: I just almost did that, and the reason I didn't both times is because I hate saying 'OGC API - Features'. But it's a good point, I'll figure out how to say it.

jisantuc (approved):

matthewhanson (approved):

m-mohr (approved):

Issue #1115, opened 2021-04-30
Empty catalogs are not allowed


jjrom:
The catalog requirement to have « at least one item or child link » does not allow to have empty catalog.

It should be possible to create a catalog before being populated. This is particularly true for dynamic catalogs.
For instance suppose a dynamic catalog of "Images of fires in France the last week" accessible at the url http://example.com/catalogs/firesinfrancelasteweek/catalog.json.

This url should always return a stac catalog wether if there were fires last week or not. Not allowing empty catalogs means that depending on the situation, the url should returns either an HTTP 200 (fires) or an HTTP 404 (no fires).

Morover, it should not always be possible to remove the empty catalogs at the parent level. For instance, if a dynamic implementation provides thousands of catalogs at the parent level, it needs to check the emptyness of each child catalog at runtime before rendering for each request (since everything is dynamic). This is an unwanted performance issue

PR #1116, opened 2021-04-30
Allow empty catalogs

jjrom:
**Related Issue(s):** #1115

**Proposed Changes:**

1. Remove the requirement that disallows empty catalogs. 

*Note: Said this is not a breaking change, as this is just relaxing a requirement, but could mark it as breaking if that is desired. It is a fairly major change, does require a new RC to be cut*

**PR Checklist:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+
+- An empty Catalog is now allowed, removed the requirement that it must have a child or item link. ([#1115](https://github.com/radiantearth/stac-spec/issues/1115))
+
 ## [v1.0.0-rc.3] - 2021-04-29
 
 ### Added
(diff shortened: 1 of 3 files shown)


cholmes:
@jjrom - this would be good to get in asap. The schema needs a bit more work, as does the wording. Would it be ok to collaborate directly on your branch on this?

jjrom:
@cholmes Yes sure !

cholmes:
Ok, I'm going to commit straight to your dev branch, as I think that is enabled.

cholmes (approved):
I updated this for the schema to reflect the intent (thanks @m-mohr for instructions on how) and also made it so it's recommended to have item or child, and touched on use cases when you may not.

cholmes:
@jjrom - check the PR to make sure it still matches what you want.
@m-mohr - didn't end up doing best practice section, as it seems like it'd be a very short section, and wasn't sure where else it would fit in. And didn't seem bad to just fit it into the main spec. But I can try again if you don't think it's right there.

jjrom:
@cholmes Yep it perfectly matches what i want :)

m-mohr (approved):

Issue #1120, opened 2021-05-04
Clarify Catalog and Collection extension use cases


lossyrob:
As described in the Gitter conversation [starting here](https://gitter.im/SpatioTemporal-Asset-Catalog/Lobby?at=609098d0f7e4221825b69a23), some clarification is needed around when a Collection should list an extension in it's `stac_extensions`. This will require updating the language of the core spec, probably in the `extensions/README.md`, as well as changes to the [stac-extensions](https://github.com/stac-extensions) repos.

Extensions should be listed in a Collection if:
- Any of the extension properties are listed in the 'summaries' list of that Catalog or Collection
- There is an extension to the Collection that brings in information specific to that extension; for example, if a Collection implements the [item-assets](https://github.com/stac-extensions/item-assets) extension and the Asset definitions have extension properties.
- Any of the Collection's `assets` entries have properties of that extension.

With #1122 Catalogs dropped summaries (again), so we won't need to mention Catalogs as part of these specific scenarios. However, we should add language that if a future extension extends Catalogs, and if part of that extension is bringing in other extension properties from child Collections or Items, that the Catalog should then list that second extension in the stac_extensions.

m-mohr:
Once #1122 is merged, you can ignore Catalogs mostly. ;-)

lossyrob:
Updated description to account for #1122

PR #1123, opened 2021-05-05
Add text around when to include an extension in stac_extensions

lossyrob:
**Related Issue(s):** #1120

**Proposed Changes:**

Adds clarification on when an extension ID (the JSON Schema URL) should be included in `stac_extensions`

**PR Checklist:**

- [ ] I have added my changes to the [CHANGELOG](https://github.com/radiantearth/stac-spec/blob/dev/CHANGELOG.md)
- [ ] ~~This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec),
      and I have opened issue/PR #XXX to track the change.~~


cholmes (approved):
Added a little bit of reorg and intro text to help explain to people how to 'use' an extension. Moved the 'general conventions' down to the 'new extensions' section, as it seemed to mostly be for those who are making extensions, and less relevant as the first thing to see.

lossyrob:
@cholmes changes LGTM

extensions/README.md
```diff
@@ -33,15 +35,37 @@ can be aware of it.
...
+identifiers for specific version of the extension that the STAC object implements. The extension ID can be
+found listed as the 'identifier' in the second line of the README of any extension made with the [extension 
+template](https://github.com/stac-extensions/template), and new ones get published automatically with any release made 
+with the template.
+
+### Extension IDs in `stac_extensions`
+
+The logic for when an object should list an extension ID in its `stac_extension` array is as follows:
+
+- If the object directly implements the extension, the  `stac_extensions` of that object should contain the extension ID.
```
m-mohr, line 59: From what I understand from recent conversations on GitHub, we should clarify what "implements" actually means. Someone may come to the conclusion that just providing the schema url without adding any fields is also an implementation...

extensions/README.md
```diff
@@ -56,7 +56,7 @@ with the template.
 
 The logic for when an object should list an extension ID in its `stac_extension` array is as follows:
 
-- If the object directly implements the extension, the  `stac_extensions` of that object should contain the extension ID.
+- If the object directly implements the extension (by including at least one of the fields of the extension, plus any additional specified requirements), the  `stac_extensions` of that object should contain the extension ID.
```
m-mohr: As there may be some extensions that don't add fields, maybe say something like:

> by including at least one of the fields of the extension (if there are any) or implementing proposed behaviors, plus any additional specified requirements.

Please improve the wording that seems like a very "German" way of writing it...

extensions/README.md
```diff
@@ -33,15 +35,38 @@ can be aware of it.
...
+extensions](https://stac-extensions.github.io/) and re-use fields there instead of creating your own version. This
+increases interoperability, as users know that the meaning of your fields is the same as in other STAC 
+implementations. Many clients will also understand more mature extensions for better display and querying. 
+
+To incorporate an extension in STAC the 'extension ID' of the extension must be added to the `stac_extensions`
+array of the STAC [Catalog](../catalog-spec/catalog-spec.md#stac_extensions), 
+[Collection](../collection-spec/collection-spec.md#stac_extensions) or [Item](../item-spec/item-spec.md#stac_extensions)
+object. This identifier is a link to the JSON Schema URL that validates the fields in the extension, so STAC validators
+can fetch the Schema to validate that the STAC object properly follows the extension. These JSON Schema URLs also act as 
+identifiers for specific version of the extension that the STAC object implements. The extension ID can be
```
m-mohr, line 50:
```suggestion
identifiers for specific version of the extension that the STAC object implements. The extension ID can usually be
```

Not all (3rd party) extensions may follow our template.

extensions/README.md
```diff
@@ -33,15 +35,38 @@ can be aware of it.
...
+The logic for when an object should list an extension ID in its `stac_extension` array is as follows:
+
+- If the object directly implements the extension (by including at least one of the fields of the extension, plus any 
+  additional specified requirements), the  `stac_extensions` of that object should contain the extension ID.
+- If an Asset implements fields of the extension, then `stac_extensions` of the Item or Collection which holds that
+  Asset should contain the extension ID.
+- If a Collection [summary](../collection-spec/collection-spec.md#summaries) contains Item fields that implement an extension, then
+  the `stac_extensions` array of that Collection should list the extension ID. For example, if a Collection `summaries` field
+  contains a summary of `eo:bands`, then that Collection should have the EO extension JSON Schema URL in the `stac_extensions` array.
+- If an object implements an extension that results in fields from a separate extension to be referenced, then the latter extension
```
m-mohr, line 66: Maybe add that this is usually the case, but extensions may specify otherwise (i.e. they say explicitly that stac_extensions should not be populated?)

extensions/README.md
```diff
@@ -79,6 +104,20 @@ with existing extensions as well as possible and may even re-use fields and thei
 into a new extension that combines commonly used fields across multiple extensions.
 Best practices for extension proposals are still emerging in this section.
 
+### General Conventions
+
+Creating a new extension involves defining a set of logically grouped fields, and specifying what the allowed values
```
m-mohr, line 109: Not all extensions may define fields, some may just define behavior (e.g. formerly the Commons extension) or define e.g. just rel types or asset roles.

extensions/README.md
```diff
@@ -79,6 +104,20 @@ with existing extensions as well as possible and may even re-use fields and thei
...
+can theoretically add fields anywhere in JSON there are some conventions as to where to add them in STAC objects.
+
+1. Additional attributes relating to an [Item](../item-spec/item-spec.md) should be added into the Item Properties object,
+   rather than directly in the Item object.
+2. In general, additional attributes that apply to an Item Asset should also be allowed in Item Properties and vice-versa.
+   For example, the `eo:bands` attribute may be used in Item Properties to describe the aggregation of all bands available in
+   the Item Asset objects contained in the Item, but may also be used in an individual Item Asset to describe only the bands available in that asset.
+3. Additional attributes relating to a [Catalog](../catalog-spec/catalog-spec.md) or
+   [Collection](../collection-spec/collection-spec.md) should be added to the root of the object.
+
```
m-mohr, line 120: Maybe as 4. point, extensions may also extend other extensions, e.g. to add new fields to eo:bands/raster:bands or so?

extensions/README.md
cholmes: Will there be extensions that don't add fields? What would they do?

If there are ones that don't add fields then I think I might just say '(by including the fields specified by the extension)'

extensions/README.md
cholmes: Good point

jbants (approved):

Issue #1125, opened 2021-05-09
Collection: Open intervals for temporal extents


m-mohr:
We recently took over a new clarification for the extents in collections from OGC APIs, see #1064. 
While implementing this for stac-migrate, I realized that we may have to change some things here.

The spec says for temporal intervals:
> Open date ranges are supported by setting either the start or the end time to null.

My understanding is that this allows only one of them to be null, so `[null, null]` is not allowed.
With the chanegs in #1064 the case may come up.

For example, if you had the following in a Collectionbefore:
`[[null, "1990-01-01T00:00:00Z"], ["2000-01-01T00:00:00Z", null]]`

The result of generating the "union" to fulfill the requirement in #1064 would be:
`[null, null]`

So what should we do? I'm not sure how to proceed with the implementation in stac-migrate.
For now I'm generating invalid extents as follows, but that feels pretty weird:
`[[null, null], [null, "1990-01-01T00:00:00Z"], ["2000-01-01T00:00:00Z", null]]`

cc @cportele

cportele:
Just to confirm that in OGC API Features `[[null, null]]` is valid. It is basically the temporal equivalent to a spatial extent `[[-180.0, -90.0, 180.0, 90.0]]` (which I think is valid in STAC).

Yes, a more specific extent will be useful for clients and for most datasets it should be possible to at least identify a start of the temporal extent. However, maybe there are cases where this is expensive to determine / keep up-to-date.

There is also a difference between a `temporal: [[null, null]]` or no `temporal` member for the API. The presence of a temporal extent indicates to clients that the items are temporal and can be queried in a meaningful way using the `datetime` query parameter. If there is no `temporal` member clients will assume that a `datetime` parameter has no effect.

m-mohr:
Thanks, @cportele. So we require temporal and spatial extents in STAC and have the wording above, but we may just slightly change to allow open intervals in both directions.

Issue #1095, opened 2021-04-20
Clarification on timezones - UTC versus local times and offsets


scottyhq:
https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#properties-object states that datetime requires UTC, but not very emphatically:

> REQUIRED. The searchable date and time of the assets, in UTC. It is formatted according to RFC 3339, section 5.6. null is allowed, but requires start_datetime and end_datetime from common metadata to be set.

I'm used to seeing UTC times in metadata, but it's also convenient to have the timezone information somewhere to easily convert to local time without relying on location-based lookups. "time-numoffset" according to (https://tools.ietf.org/html/rfc3339#section-5.6)

This came up in a discussion of how to structure libraries that load in data from STAC records.  Essentially, is it safe to assume times will always be UTC?
https://github.com/gjoseph92/stackstac/issues/2#issuecomment-822651315

scottyhq:
Also copying over the comment from @m-mohr in the issue linked above: 

> I'm not sure myself. I'd suggest opening an issue in stac-spec with a use case description so that we can discuss there. I could imagine that this is a legacy thing that we still want to change, although start_datetime and end_datetime also require UTC. On the other hand, created and updated don't require UTC.

I was mostly interested in `properties:datetime` but i could see an advantage in enforcing consistency and stating in bold that **all times in STAC metadata should be in UTC**. Which would also eliminate the possibility of storing valid ISO8601 strings like '2018-10-28T01:30:00+02:00'. And having code that ingests STAC do conversions such as (`pandas.to_datetime('2018-10-28T01:30:00+02:00').tz_convert('UTC')`)

m-mohr:
ToDos:

- If we decide on all UTC, we should add the requirement to the timestamp extension + created + updated. We should also add something to the JSON Schema that checks it, e.g. a pattern like `/Z$/` or so.
- If we decide against UTC, we need to remove it from datetime + start/end_datetime.

cholmes:
I think enforcing UTC is more in line with the general STAC philosophy - have the 'main' thing be clear and easy. We use GeoJSON, and don't get into projections for the main geometry stuff. 

But agree we should emphasize it more.

I think for easy conversion to local time I could see an extension that helps with that. Lets you add more datetimes in local time zones or something... But I prefer making the default path simple, and not allowing more options that force more complexity on every client.

m-mohr:
It's all UTC now.

PR #1127, opened 2021-05-10
Collection: Open intervals for temporal extents

m-mohr:
**Related Issue(s):** #1125

**Proposed Changes:**

1. An open date range to both sides is now allowed in the Collection's temporal extents.

**PR Checklist:**

### DIFF
--- a/collection-spec/collection-spec.md
+++ b/collection-spec/collection-spec.md
@@ -209,7 +209,7 @@ Timestamps consist of a date and time in UTC and MUST be formatted according to
 The temporal reference system is the Gregorian calendar.
 
 Open date ranges are supported by setting the start and/or the end time to `null`.
-Example for data from the beginning of 2019 until now: `[["2009-01-01T00:00:00Z", null]]`.
+Example for data from the beginning of 2019 until now: `[["2019-01-01T00:00:00Z", null]]`.
 It is recommended to provide at least a rough guideline on the temporal extent and thus
 it's not recommended to set both start and end time to `null`. Nevertheless, this is possible
 if there's a strong use case for an open date range to both sides.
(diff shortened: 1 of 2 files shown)


collection-spec/collection-spec.md
```diff
@@ -202,12 +202,17 @@ Clients only interested in the overall extent will only need to access the first
 It is recommended to only use multiple temporal extents if a union of them would then include a large
 uncovered time span (e.g. only having data for the years 2000, 2010 and 2020).
 
-Each inner array consists of exactly two dates and times, each in UTC.
-Each date and time MUST be formatted according to [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6).
+Each inner array consists of exactly two elements, either a timestamp or `null`.
```
philvarner, line 205: I think we use "datetime" most places in the text instead of "timestamp" -- I always think of unix epoch milliseconds when I see timestamp.

m-mohr: With the explanation below it should be clear?! It's a bit straneg to write "Datetimes consist of a date and time" below...

cholmes (approved):

collection-spec/collection-spec.md
philvarner, line 205:
```suggestion
Each inner array consists of exactly two elements, each of which are either a datetime string formatted according to RFC 3339 or `null`.
```

collection-spec/collection-spec.md
```diff
@@ -202,12 +202,17 @@ Clients only interested in the overall extent will only need to access the first
...
+
+Timestamps consist of a date and time in UTC and MUST be formatted according to
```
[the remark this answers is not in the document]
cholmes: Yeah, we decided on that a bit ago - https://github.com/radiantearth/stac-spec/issues/1095

collection-spec/collection-spec.md
cholmes: Discussing on the call to release - we only really use 'datetime' as our definition for the field in Item, not as a general way of talking about it. So I think we can just leave this as timestamp.

Issue #1130, opened 2021-05-11
Add more validation to Circle CI jobs


jbants:
Add another circle CI job to validate the stac examples

PR #1131, opened 2021-05-11
CI Validation 

jbants:
**Related Issue(s):** #1130 

**Proposed Changes:**

1. Added python stac-validator to circle ci

**PR Checklist:**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2
 jobs:
-  test_examples:
+  test_examples_node:
     working_directory: ~/stac
     docker:
       - image: circleci/node:12
@@ -12,7 +12,7 @@ jobs:
       - run:
           name: validate
           command: npm run check-examples
-  test_examples_2:
+  test_examples_python:
     working_directory: ~/stac
     docker:
       - image: circleci/python:3.8
@@ -55,8 +55,8 @@ workflows:
   version: 2
   ci:
     jobs:
-      - test_examples
-      - test_examples_2      
+      - test_examples_node
+      - test_examples_python   
       - test_docs
       - publish_schemas:
           filters:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -12,6 +12,18 @@ jobs:
       - run:
           name: validate
           command: npm run check-examples
+  test_examples_2:
```
m-mohr, line 15: Maybe rename to test_examples_python and test_examples_node (above) so that it is clearer from the PR directly which tool failed?

m-mohr:
Why does it try to run publish_schemas here?

jbants:
I'm not sure.
This should capture only tagged versions, correct?
``` yaml
      - publish_schemas:
          filters:
            tags:
              only: /^v.*/
            branches:
              ignore: /^((?!dev).)*$/
```              
https://github.com/radiantearth/stac-spec/blob/dev/.circleci/config.yml#L48-L53

m-mohr:
Tagged versions and dev. 
Anyway, it's just dev, so we may not even want to care?

m-mohr (approved):

cholmes (approved):

cholmes:
Why is this failing validation? I don't remember 'publish_schemas'...

jbants:
I'm guessing here, but since I made the PR from my repo instead of the stac-spec repo, my creds (jbants) are not authorized to publish. The error below is what i'm basing this off of. 

```
ProcessError: ERROR: The key you are authenticating with has been marked as read only.
fatal: Could not read from remote repository.
```

But, like Matthias said, this is for dev.

lossyrob (approved):
LGTM, as long as the failing CI, which doesn't seem to have to do with these changes, are cleared.

Seems like this could be an issue only if the PR is made off of fork with branch name `dev`?

PR #1185, opened 2022-08-24
update code of conduct to v2.1

philvarner:
**Related Issue(s):** 

n/a

**Proposed Changes:**

1. update code of conduct to version 2.1 

**PR Checklist:**

### DIFF
--- a/.remarkignore
+++ b/.remarkignore
@@ -1,0 +1,1 @@
+/CODE_OF_CONDUCT.md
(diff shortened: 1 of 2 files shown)


matthewhanson (approved):

m-mohr (approved):

.remarkignore
```diff
@@ -0,0 +1 @@
+/CODE_OF_CONDUCT.md
```
m-mohr, line 1: Can't we make it that the file just validates?

philvarner: I cut and pasted it directly from the original, and didn't want to modify it so it does every time.

PR #1188, opened 2022-09-20
Create CITATION.cff

m-mohr:
**Related Issue(s):** None

**Proposed Changes:**

Add a citation file. The main issue is that it only allows software or dataset, not techreport or specification:
https://github.com/citation-file-format/citation-file-format/issues/403

**PR Checklist:**


CITATION.cff
```diff
@@ -0,0 +1,19 @@
+cff-version: 1.2.0
+message: "If you are referring to the STAC specification in your publications, please cite it as below."
+type: software
+title: "SpatioTemporal Asset Catalog (STAC) specification"
+authors:
+  - name: "STAC Project Steering Committee"
```
pieschker, line 6: Should this be SpatioTemporal Asset Catalog PSC instead? Is the steering committee more of a governing body than an author? Is there a difference?

CITATION.cff
```diff
@@ -0,0 +1,19 @@
+cff-version: 1.2.0
+message: "If you are referring to the STAC specification in your publications, please cite it as below."
+type: software
```
pieschker, line 3: Is it a software or a standard?

pieschker (commented):
I'd like to know what others think on my two questions

CITATION.cff
m-mohr: This is a requirement due to citation-file-format/citation-file-format#403

CITATION.cff
m-mohr: I think the way it is makes sense and gives the most reasonable "citation line" with not so much "repetition".

jbants: GDAL says authors are contributors. https://github.com/OSGeo/gdal/blob/master/CITATION

Maybe something like 
```type: software
title: "SpatioTemporal Asset Catalog (STAC) specification"
authors:
  - name: "STAC Contributors"
  ```

Comment (m-mohr) [reply CITATION.cff:6]: Yes, that would also work. :-) I have no preference.

**Review by jbants — APPROVED** [review]
Looks good.

File: CITATION.cff [hunk 3]
```diff
@@ -0,0 +1,19 @@
...
+  title: "SpatioTemporal Asset Catalog (STAC) specification"
+  abstract: "Making geospatial assets openly searchable and crawlable."
+  version: 1.0.0
+  year: 2021
+  date-released: 2021-05-21
+  license: Apache-2.0
+  url: https://stacspec.org
+  repository: https://github.com/radiantearth/stac-spec
+  authors:
+    - name: "STAC Project Steering Committee"
```
lossyrob, line 18:
```suggestion
    - name: "STAC Contributors"
```

CITATION.cff
lossyrob, line 6:
```suggestion
  - name: "STAC Contributors"
```

lossyrob (approved):
LGTM after suggested changes

CITATION.cff
lossyrob: +1 on STAC Contributors, suggested that as changes.

cholmes (approved):

PR #1200, opened 2022-11-30
Fix a missing closing bracket

kurtmckee:
**Related Issue(s):** N/A

**Proposed Changes:**

1. Fixed a missing closing bracket

**PR Checklist:**


m-mohr:
Thank you for the PR, but the change is not correct.

The existing wording is:
> Map<string, \[\*]\|[Range Object](#range-object)\|[JSON Schema Object](#json-schema-object)>

This is meant to be an object (Map) with strings as keys and a value that may either be: 
- An array of anything (`\[\*]`),
- A [Range Object](#range-object), OR
- A [JSON Schema Object](#json-schema-object)

I agree that this is not very easy to understand, but it is correct as it is.

kurtmckee:
Ah, I see that now. Thank you for reviewing this, and for the explanation!

PR #1201, opened 2022-11-30
Fix a missing pluralization

kurtmckee:
**Related Issue(s):** N/A

**Proposed Changes:**

1. Fixed a missing 's' on the word "providers".

   `providers` is an array so this sentence only makes sense if the ID is unique across all of the listed providers (plural).

**PR Checklist:**


m-mohr (approved):

m-mohr (changes requested):
Looking at this again, I'm not sure this is actually correct. You can only verify the ID is unique across what you as a provider have control over. You can't control this for all providers.

m-mohr:
Thank you for the PR. We discussed it in the STAC PSC meeting today and came to the conclusion that both "provider" and "providers" is ambiguous and thus we drafted a new wording that hopefully is more clear. See PR #1209 for details.

kurtmckee:
Thanks Matthias!

Issue #1203, opened 2022-12-06
Refactor hard-to-maintain lists


m-mohr:
In the best practices we have lists of relation types, media types etc. With the slow release cycle of the main spec, this is getting outdated quickly and it's relatively hard to maintain the list. I'd propose to clean-up the list and maintain it separately. For example, the list of relation types seems for some still relatively "unmature" (origin seems to be the card4l extension), but we list them regardless in the main spec, which is weird.

Similarly, I'd propose to remove the list of stable extensions from the extensions page and instead just link to stac-extensions.github.io.

cboettig:
:+1: Thanks @m-mohr !  

Would love to see a more extensive, probably crowd-sourced list of what people are using for media types for different assets.  https://github.com/radiantearth/stac-spec/blob/master/best-practices.md#common-media-types-in-stac is a nice start but obviously hardly scratches the surface of types.  

Meanwhile, are their any guidelines for winging this? I'm not hopeful that we'll ever get registered MIME types for most of these (how long have we been waiting for parquet to get a MIME type??)  I'm currently trying to document some vector objects that use flatgeobuf, and ESRI's shapefile and geodatabase formats; wondering if the GDAL driver names might provide a reasonable basis of reference: https://gdal.org/drivers/vector/index.html , e.g. `application/x+<DRIVER SHORT NAME>` maybe?

m-mohr:
I've moved your issue to #1295 @cboettig 
It seems somewhat different from the general refactoring, which is not really meant to change contents at this point.
But it seems valuable to discuss your points so moved them into a separate issue to better keep track.

PR #1205, opened 2022-12-06
Update extensions readme

m-mohr:
**Related Issue(s):** #1203

**Proposed Changes:**

1. Remove the list of STAC extensions, only refer to stac-extensions.github.io for easier maintenance and updates
2. A lot of changes and improvements in wording/text, cater for recent changes e.g. in how we handle and write STAC extensions (scope list)
3. Better explain that you can basically extend every object
4. Tried to remove a bit of duplication
5. Moved extension maturity a level up as it is important for most chapters, not only for extending STAC

**PR Checklist:**

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -2,14 +2,13 @@
 
 - [Overview](#overview)
 - [Using Extensions](#using-extensions)
-  - [Extension IDs in `stac_extensions`](#extension-ids-in-stac_extensions)
-- [Stable STAC Extensions](#stable-stac-extensions)
+  - [Extension identifiers in `stac_extensions`](#extension-identifiers-in-stac_extensions)
 - [Community Extensions](#community-extensions)
   - [Proposed extensions](#proposed-extensions)
+- [Extension Maturity](#extension-maturity)
 - [Extending STAC](#extending-stac)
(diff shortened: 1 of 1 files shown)


cholmes (approved):
looks good, thanks!

extensions/README.md
```diff
@@ -18,74 +17,74 @@
...
+[extension template](https://github.com/stac-extensions/template).
+
+### Extension identifiers in `stac_extensions`
+
+Generally, if an extension is implemented in a STAC file in a place where the extension scope applies to, 
+the extension identifier should be added to the `stac_extension` array. The scope of an extension is usually
+explained in the README of an extension. Implementing an extension by following the specified requirements usually means including 
+fields, but occasionally also means implementing alternate behaviors.
+
+There is no direct inheritance between childs and parents though, so if for example an Item implements an extension,
```
philvarner, line 60:
```suggestion
There is no direct inheritance between children and parents objects though, so if for example an Item implements an extension,
```

m-mohr:
```suggestion
There is no direct inheritance between children and parents though, so if for example an Item implements an extension,
```

PR #1206, opened 2022-12-29
update description for bbox property in item-spec.md

nkleinbaer:
**Related Issue(s):** #
None

**Proposed Changes:**

1.  Update the description for the `bbox` property in `item-spec.md` to indicate that `bbox` is forbidden when `geometry` is `null`

If this was a new requirement it would be a breaking change, but IMO this is simply an editorial change to make the markdown more accurately describe a requirement that is already present in the schema:
https://github.com/radiantearth/stac-spec/blob/fd8fb4e3cf0b6c4a0fe74822accbc1725a92f92a/item-spec/json-schema/item.json#L67-L80

**PR Checklist:**


matthewhanson (approved):

m-mohr:
Thank you.

PR #1219, opened 2023-04-03
fix table formatting

romeokienzler:
**Related Issue(s):** #

**Proposed Changes:**

1. remote space to force proper cell asignment


philvarner:
API spec is no longer in this repo. It is located at <https://github.com/radiantearth/stac-api-spec> now

Issue #1198, opened 2022-11-04
HTTP additions for links


m-mohr:
In STAC API spec we have additional properties for links:
- method
- headers
- body
- (merge)

At least the `method` should be added to stac-spec as it could be helpful to specify e.g. the method for links (e.g. POST vs GET search or in https://github.com/stac-extensions/order/issues/9). `headers` and `body` could also be added as a general framework for HTTP requests. I think `merge` only makes sense in the specific API context.

They are described here (weirdly hidden in the Pagination section):
https://github.com/radiantearth/stac-api-spec/blob/604ade6158de15b8ab068320ca41e25e2bf0e116/item-search/README.md#pagination (scroll down to the table)

m-mohr:
Wait for #1199 to be merged, then we can fix it in a single location.

m-mohr:
Might also be useful for assets?
The GEE data seems to be only available via POST: https://matrix.to/#/!keYKZtlQXulQARQNzG:gitter.im/$Fa4N2N4_lkf0U3g-GKVsQwX_vO4R4yDxvaKH5e202M4?via=gitter.im&via=matrix.org&via=chard.com

ycespb:
We have an additional use case (for method="POST"), where we are using a STAC asset to provide an (OGC kind-of) offering allowing to "order" (make the product available "online") via an ordering URL.  The "order" is done via a link and method="POST" (with an empty/optional payload) and we would expose this I/F as a STAC asset as shown below.

----
    "offering_1": {
      "method": "POST",
      "roles": [
        "order"
      ],
      "href": "https://..../.../v1/Products(42d16f44-f4f1-64f6-9fad-a14e68bcef9e)/Order",
      "title": "Order",
      "type": "application/json"
    },
----

ycespb:
It would be nice if the link properties could be extended to allow for templated links as proposed by https://docs.ogc.org/DRAFTS/20-004.html#sc_templated_links_with_variables.   An collection asset representing an OGC WCS offering (GetCoverage) could then be provided with the allowed parameters and a client application (STACBrowser) can then replace them with actual values based on user choices.

m-mohr:
@ycespb Templated links are blocked by https://github.com/opengeospatial/ogcapi-records/issues/275#issue-1731925116 and related issues in OGC.

m-mohr:
Yet another use case is the example-links extension: https://github.com/stac-extensions/example-links#fields

gadomski:
[...]

PR #1253, opened 2023-09-27
New field in links (method, headers, body)

emmanuelmathot:
**Related Issue(s):** #1198

**Proposed Changes:**

1. Added `method`, `headers`, `body` fields in `links` objects

**PR Checklist:**

- [X] This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec),
      and I have opened issue https://github.com/radiantearth/stac-api-spec/issues/435 to track the change.


m-mohr (commented):
This looks generally good, but a couple of thoughts (for here and partially also the API):

- headers: Shouldn't this be strictly just be a `Map<string, string>`?
- body: Shouldn't this allow anything? Why is it only an object? Why can't it be a string or an array for example?
- Should we have a central place where we define links so that it's not so repetitive?

emmanuelmathot:
> This looks generally good, but a couple of thoughts (for here and partially also the API):
> 
> * headers: Shouldn't this be strictly just be a `Map<string, string>`?
> * body: Shouldn't this allow anything? Why is it only an object? Why can't it be a string or an array for example?

I used the same spec as per STAC API to be aligned since the initial issue was about being in line with STAC API.

> * Should we have a central place where we define links so that it's not so repetitive?

I moved link object definitions from item, catalog and collection to common metadata page

m-mohr:
> I used the same spec as per STAC API to be aligned since the initial issue was about being in line with STAC API.

Maybe that's something we should actually bring up in the API and change there, too...

item-spec/common-metadata.md
```diff
@@ -207,3 +208,23 @@ optical and short-wave IR bands are all 30 meters, but the panchromatic band is
...
+| rel        | string | **REQUIRED.** Relationship between the current document and the linked document. See chapter "Relation types" for more information.                                            |
+| type       | string | [Media type](../catalog-spec/catalog-spec.md#media-types) of the referenced entity.                                                                                            |
+| title      | string | A human readable title to be used in rendered displays of the link.                                                                                                            |
+| method     | string | The HTTP method required to be supported by the target resource. `GET` by default                                                                                              |
+| headers    | object | The HTTP headers required to be sent by the target resource.                                                                                                                   |
```
m-mohr, line 226: We don't use the object type in the spec, instead we use `Map<string, *>` (applies to headers and body for now)

m-mohr: `Map<string, string|array<string>>`

`{a: "b"}`
translates to:
`a: b`

`{a: ["b", "c"]}`
translates to:
```
a: b
a: c
```

This could hopefully also be clarified in the STAC API spec.

item-spec/common-metadata.md
```diff
@@ -221,15 +221,15 @@ with the links section, to describe things like the Catalog an Item is in, relat
...
+| rel        | string         | **REQUIRED.** Relationship between the current document and the linked document. See chapter "Relation types" for more information.                                            |
+| type       | string         | [Media type](../catalog-spec/catalog-spec.md#media-types) of the referenced entity.                                                                                            |
+| title      | string         | A human readable title to be used in rendered displays of the link.                                                                                                            |
+| method     | string         | The HTTP method required to be supported by the target resource. `GET` by default                                                                                              |
+| headers    | Map<string, *> | The HTTP headers required to be sent by the target resource.                                                                                                                   |
```
m-mohr, line 231: I think we still need to explain what we discussed in the last call. Arrays will be expanded to multiple headers.

Ultimately, we could also say:
`Map<string, string | [string]>`

item-spec/common-metadata.md
```diff
@@ -221,15 +221,15 @@ with the links section, to describe things like the Catalog an Item is in, relat
...
+| body       | object         | The HTTP body required to be sent by the target resource.                                                                                                                      |
```
m-mohr, line 232:
```suggestion
| body       | Map<string, *> | The HTTP body required to be sent by the target resource.                                                                                                                      |
```

emmanuelmathot: why the body as Map? body can be anything like the [stac api spec](https://github.com/radiantearth/stac-api-spec/blob/604ade6158de15b8ab068320ca41e25e2bf0e116/core/commons.yaml#L421), no?

m-mohr: Right, you are correct. But then it's also not an object. So it should be:

```suggestion
| body       | any            | The HTTP body required to be sent by the target resource.                                                                                                                      |
```

item-spec/common-metadata.md
emmanuelmathot: explaination added

item-spec/common-metadata.md
```diff
@@ -212,3 +213,23 @@ optical and short-wave IR bands are all 30 meters, but the panchromatic band is
...
+It is allowed to add additional fields such as a `title` and `type`.
+
+| Field Name | Type           | Description                                                                                                                                                                    |
+| ---------- | -------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| href       | string         | **REQUIRED.** The actual link in the format of an URL. Relative and absolute links are both allowed. [Trailing slashes are significant.](../best-practices.md#consistent-uris) |
+| rel        | string         | **REQUIRED.** Relationship between the current document and the linked document. See chapter "Relation types" for more information.                                            |
+| type       | string         | [Media type](../catalog-spec/catalog-spec.md#media-types) of the referenced entity.                                                                                            |
```
m-mohr, line 228: I guess this shouldn't link to the catalog-spec. 

```suggestion
| type       | string         | Media type) of the referenced entity.                                                                                            |
```

item-spec/common-metadata.md
m-mohr: Can't find the explanation. Where can it be found?

item-spec/json-schema/item.json
```diff
@@ -203,6 +203,18 @@
...
               "title": "Link title",
               "type": "string"
+            },
+            "method": {
+              "title": "Link method",
+              "type": "string"
+            },
+            "headers": {
+              "title": "Link headers",
+              "type": "object"
```
m-mohr, line 213: string or array of strings?!

item-spec/json-schema/item.json
```diff
@@ -203,6 +203,18 @@
...
+            },
+            "body": {
+              "title": "Link body",
+              "type": "object"
```
m-mohr, line 217:
```suggestion
              "$comment": "Any type is allowed."
```

item-spec/json-schema/item.json
```diff
@@ -203,6 +203,18 @@
             "title": {
               "title": "Link title",
               "type": "string"
+            },
+            "method": {
+              "title": "Link method",
+              "type": "string"
```
m-mohr, line 209: Should this be an enum?

item-spec/common-metadata.md
```diff
@@ -212,3 +213,29 @@ optical and short-wave IR bands are all 30 meters, but the panchromatic band is
...
+
+| Field Name | Type                             | Description                                                                                                                                                                    |
+| ---------- | -------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| href       | string                           | **REQUIRED.** The actual link in the format of an URL. Relative and absolute links are both allowed. [Trailing slashes are significant.](../best-practices.md#consistent-uris) |
+| rel        | string                           | **REQUIRED.** Relationship between the current document and the linked document. See chapter "Relation types" for more information.                                            |
+| type       | string                           | [Media type](../catalog-spec/catalog-spec.md#media-types) of the referenced entity.                                                                                            |
```
m-mohr, line 228:
```suggestion
| type       | string                           | Media type of the referenced entity.                                                                                            |
```

item-spec/common-metadata.md
```diff
@@ -212,3 +213,29 @@ optical and short-wave IR bands are all 30 meters, but the panchromatic band is
...
+| title      | string                           | A human readable title to be used in rendered displays of the link.                                                                                                            |
+| method     | string                           | The HTTP method required to be supported by the target resource. `GET` by default                                                                                              |
```
m-mohr, line 230:
```suggestion
| method     | string                           | The HTTP method that shall be used for the request to the target resource. `GET` by default                                                                                              |
```

item-spec/common-metadata.md
```diff
@@ -212,3 +213,29 @@ optical and short-wave IR bands are all 30 meters, but the panchromatic band is
...
+| headers    | Map<string, string \| \[string]> | The HTTP headers required to be sent by the target resource.                                                                                                                   |
```
m-mohr, line 231:
```suggestion
| headers    | Map<string, string \| \[string]> | The HTTP headers to be sent for the request to the target resource.                                                                                                                    |
```

item-spec/common-metadata.md
```diff
@@ -212,3 +213,29 @@ optical and short-wave IR bands are all 30 meters, but the panchromatic band is
...
+| body       | any                              | The HTTP body required to be sent by the target resource.                                                                                                                      |
```
m-mohr, line 232:
```suggestion
| body       | any                              | The HTTP body to be sent to the target resource.                                                                                                                      |
```

item-spec/common-metadata.md
```diff
@@ -212,3 +213,29 @@ optical and short-wave IR bands are all 30 meters, but the panchromatic band is
...
+
+#### HTTP headers
+
+The field `headers` allows to describe a dictionary of HTTP headers that are required to be sent by the client.
+The keys of the dictionary are the header names, and the values are either a single string or an array of strings.
+In case of an array, the header is expected to be sent multiple times with the different values.
+
+For a full discussion of the situations where relative and absolute links are recommended see the
+['Use of links'](../best-practices.md#use-of-links) section of the STAC best practices.
```
m-mohr, line 241: This seems unrelated to HTTP headers. Should this be in the other section below the table?

item-spec/common-metadata.md
```diff
@@ -212,3 +213,29 @@ optical and short-wave IR bands are all 30 meters, but the panchromatic band is
 PlanetScope Ortho Tile Product has an `gsd` of 3.7 (or 4 if rounding), even though the pixel size of the images is 3.125.
 For example, one might choose for WorldView-2 the Multispectral 20° off-nadir value of 2.07
 and for WorldView-3 the Multispectral 20° off-nadir value of 1.38.
+
+### Link Object
+
+This object describes a relationship with another entity. Data providers are advised to be liberal
+with the links section, to describe things like the Catalog an Item is in, related Items, parent or
+child Items (modeled in different ways, like an 'acquisition' or derived data).
+It is allowed to add additional fields such as a `title` and `type`.
```
gadomski, line 222: Is this sentence correct? `title` and `type` are both fields described in the table below.

item-spec/json-schema/item.json
m-mohr: @emmanuelmathot I think we should provide an enum here so that it's (1) clear which ones are allowed and (2) whether they are uppercase or lowercase (I think it's all uppercase).

item-spec/common-metadata.md
```diff
@@ -212,3 +213,28 @@ optical and short-wave IR bands are all 30 meters, but the panchromatic band is
...
+| href       | string                           | **REQUIRED.** The actual link in the format of an URL. Relative and absolute links are both allowed. [Trailing slashes are significant.](../best-practices.md#consistent-uris) |
+| rel        | string                           | **REQUIRED.** Relationship between the current document and the linked document. See chapter "Relation types" for more information.                                            |
+| type       | string                           | Media type of the referenced entity.                                                                                                                                           |
+| title      | string                           | A human readable title to be used in rendered displays of the link.                                                                                                            |
+| method     | string                           | The HTTP method that shall be used for the request to the target resource. `GET` by default                                                                                    |
```
m-mohr, line 229:
```suggestion
| method     | string                           | The HTTP method that shall be used for the request to the target resource, in uppercase. `GET` by default                                                                      |
```

m-mohr:
Looks solid, I just proposed to additional clarifications around the method. Happy to approve afterwards.

m-mohr (approved):

gadomski (approved):

Issue #1232, opened 2023-05-30
OGC API - Records alignment (license/title)


m-mohr:
I'm currently working on a [crosswalk between OGC API - Records and STAC](https://github.com/stac-utils/stac-crosswalks/blob/master/ogcapi-records/README.md) and how their "data structures" (i.e. the JSON representations of Items, Catalogs, Collections) can be better aligned.

While I've also opened a couple of issues in Records for this purpose, I think there are also a couple of links STAC could work on:

### title
`title` is required in Records for Collections and Items. Having titles in the resources and also in the links makes the browsing experience much nicer (see STAC Browser). Titles in Catalogs/Collections and their corresponding links should be aligned anyway. So I'm wondering whether we should require the title field in Catalogs and/or Collections and/or Items. Collections should always have a title, I think. It would also be good for Catalogs to explicitly choose a title and people use the id as title pretty often aynway. For Items real titles are often not available (e.g. for satellite captures), but anyway a default title provided by the provider would be good for UX.

### license
- STAC: SPDX identifier or `proprietary` or `various`
- OAR1: SPDX identifier, SPDX expression, `other` or anything(?)

Due to the divergence and the cirticism we heard a couple of times about "proprietary" for "open licenses that are non-SPDX", I'd propose to add "other" to the list of allowed values in STAC. At the same time we could also deprecate "proprietary" and/or "various"

m-mohr:
Crosswalk/Status: https://github.com/stac-utils/stac-crosswalks/blob/master/ogcapi-records/README.md

matthewhanson:
Decisions from STAC Sprint

## title
- Having title for Collections is nice, but requiring would be a breaking change
- Having title for Items we think should not be required and want to lobby for OAR1 to remove the requirement

## license
- Adopt `other` to replace `proprietary`
- deprecate `proprietary` and `various`
- Allow SPDX expressions (currently SPDX licenses aren't actually validated in STAC)

`datetime`
- No change in STAC

`type`
- No change in STAC

One option to promote/validate compatibility is to create an "OGC API Records Compatibility" extension which would, for example, make `title` and `type` required under properties.

matthewhanson:
The license issue has been merged, the only unresolved piece of this ticket is that we want to lobby for OAR1 to remove the requirement of a title on Collections.
Leaving this open until @m-mohr weighs in on if an issue has been opened for OAR on this.

m-mohr:
[...]

PR #1257, opened 2023-09-29
 Update license to follow OGC API - Records

m-mohr:
**Related Issue(s):** #1232 

**Proposed Changes:**

- The `license` field additionally supports SPDX expressions and the value `other`.
- The values `proprietary` and `various` are deprecated in favor of SPDX expressions and `other`.

**PR Checklist:**

### DIFF
--- a/examples/extensions-collection/collection.json
+++ b/examples/extensions-collection/collection.json
@@ -64,5 +64,5 @@
       ]
     }
   },
-  "license": "PDDL-1.0"
+  "license": "other"
 }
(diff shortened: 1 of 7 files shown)


matthewhanson (approved):

examples/extensions-collection/collection.json
```diff
@@ -64,5 +64,5 @@
       ]
     }
   },
-  "license": "PDDL-1.0"
```
philvarner, line 67: should we keep this as an example of using an SPDX license identifier?

m-mohr: I think there are other examples covering this case (if not, let me know). 
This was changed as there was a link in the document that didn't point to PDDL and thus was inconsistent.

philvarner (approved):

Issue #1126, opened 2021-05-10
Backport attribution from OGC APIs to a STAC extension


m-mohr:
OGC API has a field "attribution" in the landing page and in collections. While it may not be very useful for us in the landing page, it feels useful to have it in Collections somehow. As I don't want to merge it into core, I think we could make this an extension.

The issue in OGC API - Common: https://github.com/opengeospatial/ogcapi-common/issues/174

duckontheweb:
This seems like it has the potential to overlap with the scope of the Scientific Extension, at least at the Collection level...

> This extension adds the ability to indicate from which publication data originates and **how the data itself should be cited or referenced.**

If this becomes its own extensions, should there be either a restriction or a best practice that `attribution` and `sci:citation` must match for Collections?

m-mohr:
Scientific citations (usually in scientific publications) and attributions (usually in maps) are pretty different.

For example, the OSM attribution in Leaflet is simply 
> (c) OpenStreetMap contributors

On the other hand, the scientific citation would be more like 
> OpenStreetMap contributors. (2017). Planet dump retrieved from https://planet.osm.org . https://www.openstreetmap.org .

PowerChell:
To discuss and potentially close at the Sprint.

PowerChell:
Discussed and agreed to be put in the core.

m-mohr:
We discussed it again due to discussions in #1278. The mixture of CommonMark and HTML is an issue. We should test and evolve this in an extension.

m-mohr:
Done (funded by EOX): https://github.com/stac-extensions/attribution

PR #1278, opened 2024-05-07
Add attribution field to STAC entities

emmanuelmathot:
**Related Issue(s):** #1126

**Proposed Changes:**

1. Add attribution field to common basic metadata

**PR Checklist:**

### DIFF
--- a/item-spec/common-metadata.md
+++ b/item-spec/common-metadata.md
@@ -40,7 +40,7 @@ Descriptive fields to give a basic overview of a STAC entity (e.g. Catalog, Coll
 | description | string    | Detailed multi-line description to fully explain the STAC entity. [CommonMark 0.29](https://commonmark.org/) syntax MAY be used for rich text representation.                                                                                                                   |
 | keywords    | \[string] | List of keywords describing the STAC entity.                                                                                                                                                                                                                                    |
 | roles       | \[string] | The semantic roles of the entity, e.g. for assets, links, providers, bands, etc.                                                                                                                                                                                                |
-| attribution | string    | The attribution should be short and intended for presentation to a user, for example, in a corner of a map. Parts of the text can be links to other resources if additional information is needed. [CommonMark 0.29](https://commonmark.org/) syntax MAY be used for rich text. |
+| attribution | string    | The attribution should be short and intended for presentation to a user, for example, in a corner of a map. Parts of the text can be links to other resources if additional information is needed. HTML syntax MAY be used for rich text. |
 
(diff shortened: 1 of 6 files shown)


item-spec/json-schema/basics.json
```diff
@@ -29,6 +29,11 @@
       "items": {
         "type": "string"
       }
+    },
+    "attribution": {
+      "type": "string",
+      "title": "attribution for the entity",
+      "description": "Entity presentation to a user."
```
m-mohr, line 36:
```suggestion
      "title": "Attribution",
      "description": "The attribution should be short and intended for presentation to a user."
```

item-spec/common-metadata.md
```diff
@@ -34,12 +34,13 @@ Descriptive fields to give a basic overview of a STAC entity (e.g. Catalog, Coll
...
+| keywords    | \[string] | List of keywords describing the STAC entity.                                                                                                                                                                                                                                    |
+| roles       | \[string] | The semantic roles of the entity, e.g. for assets, links, providers, bands, etc.                                                                                                                                                                                                |
+| attribution | string    | The attribution should be short and intended for presentation to a user, for example, in a corner of a map. Parts of the text can be links to other resources if additional information is needed. [CommonMark 0.29](https://commonmark.org/) syntax MAY be used for rich text. |
```
m-mohr, line 43: In OGC API - Common it looks more like HTML is the markup language instead of CommonMark, although their language is a bit ambiguous.

emmanuelmathot: Since we allow markdown everywhere else when rich text is possible, allowing html seems inconsistent. md to html converter are pretty frequent

m-mohr: Yeah, on the other hand it would better align with OGC and Markdown renderers can usually just leverage HTML, too.

emmanuelmathot: I hesitated but if you say so, I put HTML

m-mohr: https://github.com/opengeospatial/ogcapi-common/issues/303#issuecomment-2098313243

m-mohr: Funny how quick my "what should go wrong" yesterday was made obsolete. The HTML/Markup issue makes me think we might want to go for an extension...

emmanuelmathot: Add to discussion for next STAC meetup

emmanuelmathot:
Discussion in STAC meetup: move it to extension

Issue #1241, opened 2023-07-19
Verify the self link is absolute


m-mohr:
Add a rule to the JSON schema that verifies that the self links are absolute.

m-mohr:
If you can't provide absolute links, don't provide a self link at all. I don't know what a relative self link really helps with?

m-mohr:
Indeed, they are optional, but if you provide them they must be absolute.

PR #1281, opened 2024-05-14
Add validation for absolute self link in item schema

emmanuelmathot:
**Related Issue(s):** #1241

**Proposed Changes:**

1. Add validation for absolute self link in item schema

**PR Checklist:**


item-spec/json-schema/item.json
```diff
@@ -204,6 +204,21 @@
...
+            "properties": {
+              "rel": {
+                "const": "self"
+              }
+            }
+          },
+          "then": {
+            "properties": {
+              "href": {
+                "format": "iri"
```
gadomski, line 219: What about local (filesystem) catalogs? E.g. with a self link of `/path/to/my/catalog.json`?

m-mohr: Shouldn't that be using the file protocol? file://path/to/my/...?

gadomski: Most tooling that I've seen _won't_ do that, e.g. **pystac**.

m-mohr: I'd say then pystac is generating invalid STAC. The STAC Spec in 1.0 says explicitly "Absolute URL" and /path/to/my/catalog.json is not an absolute URL, it's just an absolute path.

m-mohr: Then we need to fix the spec, whick pretty much always talks about URLs. That might actually require a 2.0 though, because some tooling may now expect URLs (such as STAC Browser).

m-mohr (approved):

emmanuelmathot:
@gadomski ready for review

gadomski (approved):

Issue #1160, opened 2021-10-25
GeometryCollection not allowed according to Item schema


aznan2:
The section for the geometry field in [item-spec.md](https://github.com/radiantearth/stac-spec/blob/v1.0.0/item-spec/item-spec.md) refers to [RFC 7946, section 3.1](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1), which includes all seven geometry types, so I assume that the intent is that items can have a GeometryCollection.

In the JSON schema, the geometry field must be null or conform to https://geojson.org/schema/Geometry.json. However, this file only defines the six base geometry types and does not allow for the object to be a GeometryCollection.

Not sure if this is actually an issue with the GeoJSON schema but as things stand, items with a GeometryCollection as its geometry will not validate against the schema.

m-mohr:
We don't allow GeometryCollection and remove it from the Item Spec Markdown.

PowerChell:
Just needs a PR and then can close @matthewhanson

PR #1294, opened 2024-06-27
Disallow GeometryCollection, reformatting

m-mohr:
**Related Issue(s):** #1160

**Proposed Changes:**

1. Disallow GeometryCollection in Item Spec (as defined in the schema) #1160
2. Reformatted the Item Field table to be more lightweight by moving geometry, bbox, and collection descriptions to separate sections. Also, clarified that RFC section 3.1 and 3.2 apply, for Geometry Object and null respectively.

**PR Checklist:**

### DIFF
--- a/item-spec/item-spec.md
+++ b/item-spec/item-spec.md
@@ -112,8 +112,6 @@ Items that are linked to, but the best practices around this are still emerging.
 
 Defines the full footprint of the asset represented by this item, formatted according to RFC 7946.
 
-The footprint should be the default GeoJSON geometry, though additional geometries can be included.
-
 If **a geometry** is provided, the value must be a Geometry Object according to
 [RFC 7946, section 3.1](https://tools.ietf.org/html/rfc7946#section-3.1)
 with the exception that the type `GeometryCollection` is not allowed in STAC.
--- a/item-spec/item-spec.md
+++ b/item-spec/item-spec.md
@@ -139,7 +139,7 @@ STAC compliant APIs are required to compute intersection operations with the Ite
 The `id` of the STAC Collection this Item references to with the [`collection` relation type](#relation-types) in the `links`  array.
 
 This field provides an easy way for a user to search for any Items that belong in a specified Collection.
-Must be a non-empty string.
+If present, must be a non-empty string.
 
 ### Properties Object
 
--- a/item-spec/item-spec.md
+++ b/item-spec/item-spec.md
@@ -7,7 +7,9 @@
     - [stac\_extensions](#stac_extensions)
     - [id](#id)
     - [assets](#assets)
+    - [geometry](#geometry)
     - [bbox](#bbox)
+    - [collection](#collection)
   - [Properties Object](#properties-object)
     - [datetime](#datetime)
     - [Additional Fields](#additional-fields)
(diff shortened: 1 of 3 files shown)


jbants (approved):

emmanuelmathot (changes requested):
This change in the schema would enforce it at validation
```
diff --git a/item-spec/json-schema/item.json b/item-spec/json-schema/item.json
index ede4d62a..61475e1b 100644
--- a/item-spec/json-schema/item.json
+++ b/item-spec/json-schema/item.json
@@ -15,6 +15,30 @@
         {
           "$ref": "https://geojson.org/schema/Feature.json"
         },
+        {
+          "$comment": "Disallow GeometryCollection",
+          "not": {
+            "type": "object",
+            "required": [
+              "geometry"
+            ],
+            "properties": {
+              "geometry": {
+                "type": "object",
+                "required": [
+                  "type"
+                ],
+                "properties": {
+                  "type": {
+                    "enum": [
+                      "GeometryCollection"
+                    ]
+                  }
+                }
+              }
+            }
+          }
+        },
         {
           "oneOf": [
             {
@@ -260,4 +284,4 @@
       ]
     }
   }
-}
+}
\ No newline at end of file
```

m-mohr:
I thought we were already enforcing it in validation, I'll check, thanks.

emmanuelmathot:
yes right, with the coordinates fields

m-mohr:
This line enforces it, because it doesn't contain GeometryCollection:
https://github.com/radiantearth/stac-spec/blob/master/item-spec/json-schema/item.json#L47

If that's fine for you, you could approve it @emmanuelmathot

item-spec/item-spec.md
```diff
@@ -106,6 +108,20 @@ by multiple files - all should be linked to. It is generally recommended that di
 levels or formats are not exhaustively listed in an Item, but instead are represented by related
 Items that are linked to, but the best practices around this are still emerging.
 
+#### geometry
+
+Defines the full footprint of the asset represented by this item, formatted according to RFC 7946.
+
+The footprint should be the default GeoJSON geometry, though additional geometries can be included.
```
gadomski, line 115: I don't quite understand what this sentence is communicating. What is a "default GeoJSON geometry"? Are there non-default GeoJSON geometries?

item-spec/item-spec.md
```diff
@@ -118,6 +134,13 @@ and the elevation of the northeasterly most extent is the maximum.
 This field enables more naive clients to easily index and search geospatially.
 STAC compliant APIs are required to compute intersection operations with the Item's geometry field, not its bbox.
 
+#### collection
+
+The `id` of the STAC Collection this Item references to with the [`collection` relation type](#relation-types) in the `links`  array.
+
+This field provides an easy way for a user to search for any Items that belong in a specified Collection.
+Must be a non-empty string.
```
gadomski, line 142:
```suggestion
If present, must be a non-empty string.
```

gadomski (changes requested):

item-spec/item-spec.md
m-mohr: I'm not sure either. It was there before this PR.

gadomski: If we don't understand it, we should remove it, ya?

m-mohr: It's there since the beginning, it seems. I guess the "default" is meant to be some kind of a counterpart to the "additional geometries". I'm actually not so sure how additional geometries could be added?! Maybe we should remove the whole paragraph?

m-mohr, line 115:
```suggestion
```

m-mohr:
@gadomski @emmanuelmathot Ready for another review.

gadomski (approved):

emmanuelmathot (approved):

Issue #1228, opened 2023-04-29
created / updated: Different meaning in assets and properties


m-mohr:
In general, we seem to move forward towards the "common metadata" model, which means that if a keyword is present in Item properties, it applies to all assets if not overridden in assets. This works for many properties, but for examples `created` and `updated` have different meanings. In properties it applies to the metadata and in assets it applies to the data. This also applies for the timestamp extension.

Is this fine and clients need to handle it based individually for fields or should this be changed in STAC 2.0? This would probably mean that the metadata for the STAC Item needs to move from the properties to the top-level?

m-mohr:
Related: https://github.com/radiantearth/stac-spec/blob/master/best-practices.md#common-use-cases-of-additional-fields-for-assets

When do we need properties in Item Properties, when in Assets? How's the hierarchy working? Is falling back from Assets to Item Properties valid it the asset is missing properties? We need much of this called out more specifically.

m-mohr:
Also applies to eo:bands, gsd, most timestamps, ...

m-mohr:
STAC sprint, day 2:

State in the spec that the assets inherit values from the "top-level" metadata (e.g. from Item Properties, in top-level collection properties). Mention that the top-level metadata might be nominal.

m-mohr:
I'm not 100% sure either, I think the term is not so good as it's unintuitive here.

I think what we want to say here that in some cases the inheritance doesn't necessarily make sense if other semantics have been defined for the fields (gsd [best resolution], updated, created [metadata vs data], eo:bands [union of bands in assets], ...)

PR #1299, opened 2024-07-04
Clarify the inheritance of the properties at different levels

emmanuelmathot:
**Related Issue(s):** #1228

**Proposed Changes:**

1. Clarify the inheritance of the asset properties 
2. 

**PR Checklist:**

### DIFF
--- a/item-spec/item-spec.md
+++ b/item-spec/item-spec.md
@@ -298,7 +298,7 @@ For more information on how to use roles see the [Asset Roles](../best-practices
 As detailed above, Items contain properties, which are the main source of metadata for searching across Items. Many content
 extensions can add further property fields as well. Any property that can be specified for an Item can also be specified for
 a specific asset. This can be used to override a property defined in the Item, or to specify fields for which there is no
-single value for all assets.
+single value for all assets. If a property is defined at the Item level, it is the default value for all assets.
 
 **It is important to note that the STAC API does not facilitate searching across Asset properties in this way, and this
 should be used sparingly.** It is primarily used to define properties at the Asset level that may be used during use of
--- a/item-spec/item-spec.md
+++ b/item-spec/item-spec.md
@@ -304,8 +304,9 @@ If a property is defined in a Collection on the top-level, it is the default val
 Assets may override the properties inherited for specific assets (see example below).
 Note that in some cases the inheritance may not lead to the expected results if other semantics have been defined for the property.
 See the examples for `gsd` and `eo:bands` below.
-It also applies to some other fields such as `created` or `updated`
-(inheritance may be undesirable in these cases, e.g. the `created` attribute of an item would describe when the metadata was generated, whereas on an asset the `created` attribute would describe when the data file itself was created).
+It also applies to some other fields such as `created` or `updated`.
+Inheritance may be undesirable in these cases, e.g. the `created` attribute of an item would describe when the metadata
+was generated, whereas on an asset the `created` attribute would describe when the data file itself was created.
 
 **It is important to note that the STAC API does not facilitate searching across Asset properties in this way, and this
 should be used sparingly.** It is primarily used to define properties at the Asset level that may be used during use of
--- a/item-spec/item-spec.md
+++ b/item-spec/item-spec.md
@@ -302,6 +302,7 @@ single value for all assets.
 If a property is defined in the Item Properties, it is the default value for all assets in the Item.
 If a property is defined in a Collection on the top-level, it is the default value for all assets in the Collection.
 Assets may override the properties inherited for specific assets (see example below).
+Inheritance does not apply across multiple files, i.e. is restricted to a single Item or Collection.
 Note that in some cases the inheritance may not lead to the expected results if other semantics have been defined for the property.
 See the examples for `gsd` and `eo:bands` below.
 It also applies to some other fields such as `created` or `updated`.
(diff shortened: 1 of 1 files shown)


item-spec/item-spec.md
```diff
@@ -298,7 +298,9 @@ For more information on how to use roles see the [Asset Roles](../best-practices
 As detailed above, Items contain properties, which are the main source of metadata for searching across Items. Many content
 extensions can add further property fields as well. Any property that can be specified for an Item can also be specified for
 a specific asset. This can be used to override a property defined in the Item, or to specify fields for which there is no
-single value for all assets.
+single value for all assets. If a property is defined at the STAC entity level (Item or Collection), it is the default value for all assets.
+The values in the top level properties are considered nominal for the the Item or the Collection and thus may be adjusted for
+specific assets (see example below).
```
m-mohr, line 303: Trying to make it more clear, also this is just the Item Spec so Collections should have this paragraph, too.

```suggestion
single value for all assets.
If a property is defined in the Item Properties, it is the default value for all assets.
Assets may override the properties in the Item Properties for specific assets (see example below).
Note that in some cases the inheritance may not lead to the expected results if other semantics have been defined for the property.
See the examples for `gsd` and `eo:bands` below, but it also applies to some other fields such as `created` or `updated`.
```

item-spec/item-spec.md
```diff
@@ -299,6 +299,13 @@ As detailed above, Items contain properties, which are the main source of metada
 extensions can add further property fields as well. Any property that can be specified for an Item can also be specified for
 a specific asset. This can be used to override a property defined in the Item, or to specify fields for which there is no
 single value for all assets.
+If a property is defined in the Item Properties, it is the default value for all assets in the Item.
+If a property is defined in a Collection on the top-level, it is the default value for all assets in the Collection.
+Assets may override the properties inherited for specific assets (see example below).
+Note that in some cases the inheritance may not lead to the expected results if other semantics have been defined for the property.
+See the examples for `gsd` and `eo:bands` below.
+It also applies to some other fields such as `created` or `updated`
+(metadata-related timestamps get inherited and data-related timestamps are defined in assets).
```
m-mohr, line 308:
```suggestion
(metadata-related timestamps would get inherited and mixed with data-related timestamps that are defined in assets).
```

m-mohr (approved):

item-spec/item-spec.md
```diff
@@ -299,6 +299,13 @@ As detailed above, Items contain properties, which are the main source of metada
 extensions can add further property fields as well. Any property that can be specified for an Item can also be specified for
 a specific asset. This can be used to override a property defined in the Item, or to specify fields for which there is no
 single value for all assets.
+If a property is defined in the Item Properties, it is the default value for all assets in the Item.
+If a property is defined in a Collection on the top-level, it is the default value for all assets in the Collection.
```
gadomski, line 303: > it is the default value for all assets in the Collection

Not items? E.g. if a `Collection` has `proj:epsg=32613`, does that projection information apply to the item (e.g. in the case of a `proj:geometry` on the item).

item-spec/item-spec.md
```diff
@@ -299,6 +299,13 @@ As detailed above, Items contain properties, which are the main source of metada
...
+Assets may override the properties inherited for specific assets (see example below).
+Note that in some cases the inheritance may not lead to the expected results if other semantics have been defined for the property.
+See the examples for `gsd` and `eo:bands` below.
+It also applies to some other fields such as `created` or `updated`
+(metadata-related timestamps would get inherited and mixed with data-related timestamps that are defined in assets).
```
gadomski, line 308: This parenthetical note is confusing to me - I imagine it will be for other people as well.

```suggestion
(inheritance may be undesirable in these cases, e.g. the `created` attribute of an item would describe when the metadata was generated, whereas on an asset the `created` attribute would describe when the data file itself was created).
```

item-spec/item-spec.md
m-mohr: No, we only inherit in the same file.

item-spec/item-spec.md
```diff
@@ -299,6 +299,13 @@ As detailed above, Items contain properties, which are the main source of metada
 extensions can add further property fields as well. Any property that can be specified for an Item can also be specified for
 a specific asset. This can be used to override a property defined in the Item, or to specify fields for which there is no
 single value for all assets.
+If a property is defined in the Item Properties, it is the default value for all assets in the Item.
+If a property is defined in a Collection on the top-level, it is the default value for all assets in the Collection.
+Assets may override the properties inherited for specific assets (see example below).
+Note that in some cases the inheritance may not lead to the expected results if other semantics have been defined for the property.
+See the examples for `gsd` and `eo:bands` below.
+It also applies to some other fields such as `created` or `updated`
+(inheritance may be undesirable in these cases, e.g. the `created` attribute of an item would describe when the metadata was generated, whereas on an asset the `created` attribute would describe when the data file itself was created).
```
m-mohr, line 308:
```suggestion
It also applies to some other fields such as `created` or `updated`.
Inheritance may be undesirable in these cases, e.g. the `created` attribute of an item would describe when the metadata
was generated, whereas on an asset the `created` attribute would describe when the data file itself was created.
```

item-spec/item-spec.md
```diff
@@ -299,6 +299,14 @@ As detailed above, Items contain properties, which are the main source of metada
 extensions can add further property fields as well. Any property that can be specified for an Item can also be specified for
 a specific asset. This can be used to override a property defined in the Item, or to specify fields for which there is no
 single value for all assets.
+If a property is defined in the Item Properties, it is the default value for all assets in the Item.
+If a property is defined in a Collection on the top-level, it is the default value for all assets in the Collection.
+Assets may override the properties inherited for specific assets (see example below).
```
m-mohr, line 304:
```suggestion
Assets may override the properties inherited for specific assets (see example below).
Inheritance does not apply across multiple files, i.e. is restricted to a single Item or Collection.
```

gadomski (approved):

PR #1309, opened 2024-08-09
Fix small grammer error in common metadata

gadomski:
**PR Checklist:**

### DIFF
--- a/commons/common-metadata.md
+++ b/commons/common-metadata.md
@@ -223,7 +223,7 @@ and for WorldView-3 the Multispectral 20° off-nadir value of 1.38.
 | bands      | \[[Band Object](#band-object)] | An array of available bands where each object is a [Band Object](#band-object). |
 
 The `bands` array is used to describe the available bands in a STAC entity or Asset.
-This fields describes the general construct of a band or layer, which doesn't necessarily need to be a spectral band.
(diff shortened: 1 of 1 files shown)


commons/common-metadata.md
```diff
@@ -223,7 +223,7 @@ and for WorldView-3 the Multispectral 20° off-nadir value of 1.38.
 | bands      | \[[Band Object](#band-object)] | An array of available bands where each object is a [Band Object](#band-object). |
 
 The `bands` array is used to describe the available bands in a STAC entity or Asset.
-This fields describes the general construct of a band or layer, which doesn't necessarily need to be a spectral band.
+These fields describe the general construct of a band or layer, which doesn't necessarily need to be a spectral band.
```
m-mohr, line 226:
```suggestion
This field describes the general construct of a band or layer, which doesn't necessarily need to be a spectral band.
```

Shouldn't it be singular? It just one bands field

gadomski: Yeah, I wasn't sure on the intent — I read it first as describing each of the component `Band` objects in the array. But if it's singular, then it should be as you suggest 👍🏼

m-mohr (approved):

emmanuelmathot (approved):

PR #1370, opened 2025-11-10
Remove best practices content and reference new best practices repo

jsignell:
I just realized I had never pushed this branch 🙈. New changes to best practices should go on the new repo: https://github.com/radiantearth/stac-best-practices

**PR Checklist:**

- [x] ~This PR affects the [STAC API spec](https://github.com/radiantearth/stac-api-spec),
      and I have opened issue/PR #XXX to track the change.~


gadomski (approved):

m-mohr:
Just wondering: are the file locations / names in stac-best-practices final?
I'm wondering, because the all have this lengthy name which is prefixed by "best-practices-", making them a bit clunky. Is this final or an oversight?

jsignell:
> Just wondering: are the file locations / names in stac-best-practices final? I'm wondering, because the all have this lengthy name which is prefixed by "best-practices-", making them a bit clunky. Is this final or an oversight?

It was intentional, but I am not wedded to it. There is also the `implementation-guide.md` which is intended to grow and is not prefixed.

m-mohr:
I feel like it would be a little nicer, especially when we start to render it, it's part of the URL, probably. And we probably want the URLs to be permanent. But it's also not super important, indeed.

m-mohr (approved):

jsignell:
I'll go ahead and merge this for now and I will update these URLs if we change the file naming pattern in stac-best-practices. I agree we should do that soon if we want to do it.

PR #1392, opened 2026-09-10
Clarify Collection `item_assets` 

### LINKED ISSUE #1389: Clarification of Collection `item_assets`
I'm trying to align the openEO Python client and geotrellis openEO backend implementation with the STAC spec, and I'm struggling with understanding some details related to "item_assets" in a STAC Collection.

---

About keys in "item_assets" (https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md#item_assets):
 
> item_assets
> This serves two purposes:
> 1 Provide a human-readable definition of assets available in any Items belonging to this Collection so that the user can determine the key(s) of assets they are interested in.
> 2 Provide a way to programmatically determine what assets are available in any member Item. 

As far as I know, and the description above reinforces that I think: the **asset key** is the only "handle" to connect entries listed in a Collection's "item_assets" and assets listed in a STAC Item. 

But that feels like to conflict with the description at https://github.com/radiantearth/stac-spec/blob/master/commons/assets.md, which states that the keys should be considered to have no meaning:

> In general, the keys don't have any meaning and are considered to be non-descriptive unique identifiers. Providers may assign any meaning to the keys for their respective use cases, but must not expect that clients understand them.

There is no mentioning or hint that assets of the same kind (e.g. asset covering the same band or variable) should use the same key across STAC Items to signal that relation. The description "non-descriptive unique identifiers" can even be interpreted "each asset should get some globally unique id" which is the exact opposite of what "item_assets" wants to accomplish.

--- 

About "Item Asset Definition Object" at https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md#item-asset-definition-object

> Assets included at the Collection level do not imply that all assets are available from all Items.

I think there is something grammatically wrong with this sentence. 
It is also confusing what "Assets included at the Collection level" means: assets under Collection "assets", or under Collection "item_assets"?

I suspect the intended meaning is:

> It is not required that each Asset listed in the Item Asset definition object is present in each member Item of the Collection.

is that correct?

The next sentence can also be finetuned I think:

>  However, it is recommended that the Asset Definition is a complete set of all assets that may be available from any member Items.
 
to

>  However, it is recommended that the Item Asset Definition is a complete set of all assets that may be available from any member Items.

---

a bit further, pretty confusing set of sentences (numbering added):

> (1) Any property that exists for a Collection-level asset object must also exist in the corresponding assets object in each Item. 
> (2) If a collection's asset object contains properties that are not explicitly stated in the Item's asset object then that property does not apply to the item's asset.
> (3) Item asset objects at the Collection-level can describe any of the properties of an asset, but those assets properties and values must also reside in the item's asset object. 

Aren't (1) and (3) saying the same, and isn't (2) stating the opposite?

Additionally, it's again unclear if this is about assets under "assets" (which is what "Collection-level asset object" sounds like) or "item_assets" (which is probably intended?).

Last sentence of that paragraph is also pretty weird: why promoting stoaing invalid STAC as an implementation detail?
> To consolidate item-level asset object properties in an API setting, consider storing the STAC Item objects without the larger properties internally as 'invalid' STAC items, and merge in the desired properties at serving time from the Collection-level.

Side-note: the whole paragraph started by these sentences is also sloppy w.r.t capitalization of Asset/Item/Collection.

soxofaan:
**Related Issue(s):** #1389 

**Proposed Changes:**

initial stab at incorporating the clarification of https://github.com/radiantearth/stac-spec/issues/1389#issuecomment-5370599268

**PR Checklist:**

### DIFF
--- a/collection-spec/collection-spec.md
+++ b/collection-spec/collection-spec.md
@@ -305,20 +305,25 @@ This serves two purposes:
 1. Provide a human-readable definition of assets available in **any** Items
    belonging to this Collection so that the user can determine the key(s)
    of assets they are interested in.
-2. Provide a way to programmatically determine what assets are available
+2. Provide a way to programmatically determine what kind of assets are available
    in **any** member Item. Otherwise a random Item needs to be examined to
    determine assets available, but a random Item may not be representative of the set.
+
+Note that the correspondence between Collection-level Item Asset objects and Item-level Asset objects is established by the Asset key, both for human-readable purposes and for programmatic access.
+As such, it is recommended to adopt a consistent asset key naming scheme at minimum, and ideally a descriptive one.
 
 An Item Asset Object defined at the Collection level is nearly the same as the
 [Asset Object in Items](../commons/assets.md#asset-object), except for two differences.
 The `href` field is not required, because Item Asset Definitions don't point to any data by themselves, but at least two other fields must be present.
 
-#### Item Asset Definition Object
+The Item Asset objects provide an overview of the Assets that can be expected in the member Items of the Collection.
+- The assets in this summary should be a *union* view:
+  while member Items do not need to include each and every kind of asset in the Collection,
+  it is recommended that the Item Asset Definition is a complete set of **all** kinds of assets that may be available from **any** member Items.
+- Conversely, the Item Asset objects should only list the *intersection* of properties
+  that are common to all corresponding assets in the member Items.
 
-An item asset is an object that contains details about the datafiles that will be included in member Items.
-Assets included at the Collection level do not imply that all assets are available from all Items.
-However, it is recommended that the Asset Definition is a complete set of **all** assets that may be available from **any** member Items.
-So this should be the union of the available assets, not just the intersection of the available assets.
+#### Item Asset Definition Object
 
 | Field Name  | Type      | Description                                                                                                                                                                                  |
 | ----------- | --------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
@@ -329,13 +334,7 @@ So this should be the union of the available assets, not just the intersection o
 
 Other custom fields, or fields from other extensions may also be included in the Asset object.
 
-Any property that exists for a Collection-level asset object must also exist in the corresponding assets object in
-each Item. If a collection's asset object contains properties that are not explicitly stated in the Item's asset
-object then that property does not apply to the item's asset. Item asset objects at the Collection-level can
-describe any of the properties of an asset, but those assets properties and values must also reside in the item's
-asset object. To consolidate item-level asset object properties in an API setting, consider storing the STAC Item
-objects without the larger properties internally as 'invalid' STAC items, and merge in the desired properties at
-serving time from the Collection-level.
+Any property that exists for a Collection-level asset object must also exist in the corresponding assets object in each Item.
 
 At least two fields (e.g. `title` and `type`) are required to be provided, in order for it to adequately describe Item assets.
 The two fields must not necessarily be taken from the list above and may include any custom field.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Recommend that thumbnails and similar assets use HTTP(S) as access mechanism
 - Remove best practices content and reference new best practices repo ([#1370](https://github.com/radiantearth/stac-spec/pull/1370))
+- Clarify Collection `item_assets` ([#1389](https://github.com/radiantearth/stac-spec/issues/1389))
 
 ## [v1.1.0] - 2024-09-10
 


m-mohr (approved):
Thanks, I think this is much better.

collection-spec/collection-spec.md
```diff
@@ -305,20 +305,25 @@ This serves two purposes:
...
 [Asset Object in Items](../commons/assets.md#asset-object), except for two differences.
 The `href` field is not required, because Item Asset Definitions don't point to any data by themselves, but at least two other fields must be present.
 
-#### Item Asset Definition Object
+The Item Asset objects provide an overview of the Assets that can be expected in the member Items of the Collection.
+- The assets in this summary should be a *union* view:
+  while member Items do not need to include each and every kind of asset in the Collection,
+  it is recommended that the Item Asset Definition is a complete set of **all** kinds of assets that may be available from **any** member Items.
+- Conversely, the Item Asset objects should only list the *intersection* of properties
+  that are common to all corresponding assets in the member Items.
```
emmanuelmathot, line 324: A simple set of examples would be useful I think.

emmanuelmathot (approved):

collection-spec/collection-spec.md
soxofaan: you mean inline examples in this document, or as separate JSON files in an examples folder?

emmanuelmathot: I believe stac collection and items excerpt inline could work

m-mohr: I feel like inline JSON examples could get a bit too extensive.

collection-spec/collection-spec.md
```diff
@@ -305,20 +305,25 @@ This serves two purposes:
 1. Provide a human-readable definition of assets available in **any** Items
    belonging to this Collection so that the user can determine the key(s)
    of assets they are interested in.
-2. Provide a way to programmatically determine what assets are available
+2. Provide a way to programmatically determine what kind of assets are available
    in **any** member Item. Otherwise a random Item needs to be examined to
    determine assets available, but a random Item may not be representative of the set.
 
+Note that the correspondence between Collection-level Item Asset objects and Item-level Asset objects is established by the Asset key, both for human-readable purposes and for programmatic access.
+As such, it is recommended to adopt a consistent asset key naming scheme at minimum, and ideally a descriptive one.
```
gadomski, line 313: Nit: this guidance really belongs on either [Assets](https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#assets) in the item spec, or in the [best practices](https://github.com/radiantearth/stac-best-practices/blob/main/best-practices-asset-and-link.md) — while `item_assets` are one reason to have a consistent asset key naming scheme, there's others. It's not bad to have it here, so this shouldn't block this PR.

gadomski (approved):

collection-spec/collection-spec.md
```diff
@@ -329,13 +334,7 @@ So this should be the union of the available assets, not just the intersection o
...
 Other custom fields, or fields from other extensions may also be included in the Asset object.
 
-Any property that exists for a Collection-level asset object must also exist in the corresponding assets object in
-each Item. If a collection's asset object contains properties that are not explicitly stated in the Item's asset
-object then that property does not apply to the item's asset. Item asset objects at the Collection-level can
-describe any of the properties of an asset, but those assets properties and values must also reside in the item's
-asset object. To consolidate item-level asset object properties in an API setting, consider storing the STAC Item
-objects without the larger properties internally as 'invalid' STAC items, and merge in the desired properties at
-serving time from the Collection-level.
+Any property that exists for a Collection-level asset object must also exist in the corresponding assets object in each Item.
```
jsignell, line 337: Is this true? It feels like this should be true of collection-level item asset objects, but not just the collection-level assets themselves

m-mohr: Yes, you are correct, @jsignell

PS: Welcome back :-)

collection-spec/collection-spec.md
soxofaan: I fully agree that there is room for clarification about asset keys in the context of cross-referencing  things, but that is out of scope of this PR I think

collection-spec/collection-spec.md
soxofaan: yes, I'm afraid that illustrating the description with JSON snippets will take quite a bit of space, because there are multiple documents in play here. I'd prefer improving the textual wording from this PR, so suggestions are welcome

collection-spec/collection-spec.md
soxofaan: not sure I understand the action point here. Do you mean:
```suggestion
Any property that exists for a Collection-level item asset object must also exist in the corresponding assets object in each Item.
```
(because this section is about `#### Item Asset Definition Object`)

Or should this describe both collection-level item asset objects _and_ collection-level asset objects?

m-mohr:
```suggestion
Any property that exists for a Collection-level Item Asset Definition Object must also exist in the corresponding Assets Object in each Item.
```

I guess this is the most precise way to say it.

PR #1393, opened 2026-09-17
Update licensing.json - allowing whitespace

Fred-Leclercq:
**Related Issue(s):** #

**Proposed Changes:**

1. Allowing whitespaces in license field - this allows values such as:

CC-BY 4.0
CC BY-SA 4.0
Creative Commons Attribution 4.0
CC-BY+something 

**PR Checklist:**


gadomski:
I don't think this correct — per [the collection spec](https://github.com/radiantearth/stac-spec/blob/master/collection-spec/collection-spec.md#license):

> The license(s) can be provided as:
> - [SPDX License identifier](https://spdx.org/licenses/)
> - [SPDX License expression](https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/)
> - String with the value other if the license is not on the SPDX license list. The strings various and proprietary are deprecated.

The SPDX identifier for "CC-BY 4.0" is `CC-BY-4.0`, per https://spdx.org/licenses/CC-BY-4.0.html.

Fred-Leclercq:
You're right about the CC-BY-4.0 license, but how to deal with SPDX License compound expressions where we need white spaces?

gadomski:
Yup, good point. From https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/:

> There MUST be white space on either side of the operator "WITH". There MUST be white space and/or parentheses on either side of the operators AND and OR.

So 👍🏼

item-spec/json-schema/licensing.json
```diff
@@ -5,8 +5,8 @@
   "type": "object",
   "properties": {
     "license": {
-      "type": "string",
-      "pattern": "^[\\w\\-\\.\\+]+$"
+      "type": "string",
```
gadomski, line 8: Trailing whitespace

item-spec/json-schema/licensing.json
```diff
@@ -5,8 +5,8 @@
   "type": "object",
   "properties": {
     "license": {
-      "type": "string",
-      "pattern": "^[\\w\\-\\.\\+]+$"
+      "type": "string",      
+      "pattern": "^[\\w\\-\\.\\+\\s]+$"
     }
   }
-}
```
gadomski, line 12: No reason to change the line ending in this PR

gadomski (changes requested):
If we're fixing this json-schema for expressions, we'll need to handle brackets and slashes too? E.g. `(MIT OR Apache-2.0) AND BSD-2-Clause`, `MIT/Apache-2.0`.

olegkho:
@gadomski  agree, but don't some [SPDX License expressions](https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/) require you to have whitespaces (or parentheses) like OR, AND, WITH? 

For example, "CC-BY-4.0 OR CC-BY-3.0" is a valid SPDX expression but would fail the current regex, while something like "ThisIsNotAValidLicense" would pass.

How do we proceed with tracking actual valid expressions? 

EDIT: My bad, didn't refresh page, assumed no reply.

Fred-Leclercq:
@gadomski - Should we consider removing the regex pattern?
Otherwise, this: "pattern": "^[\\w\\-\\.\\+\\s\\\\\\(\\)\\[\\]]+$" ?

